PR #247, opened 2014-08-04
Update install.rst

iboulard:
'bootstrapform' was missing in the list of installed apps

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -42,6 +42,7 @@ Adding To Your Django Project
         'south',  # Highly recommended to make database migrations simpler.
         'markdown_deux',  # Required for Knowledgebase item formatting
         'helpdesk',  # This is new!
+        'bootstrapform', 
     )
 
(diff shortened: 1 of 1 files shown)


docs/install.rst
```diff
@@ -42,6 +42,7 @@ Adding To Your Django Project
         'south',  # Highly recommended to make database migrations simpler.
         'markdown_deux',  # Required for Knowledgebase item formatting
         'helpdesk',  # This is new!
+        'bootstrapform',
```
flinz: I guess we should add 'bootstrapform' before 'helpdesk', since it's a dependency.

alexbarcelo:
The `install.rst` has been fixed in some other commit, it seems:
https://github.com/django-helpdesk/django-helpdesk/blob/master/docs/install.rst

This PR can be closed.

reduxionist:
Thanks for the effort nonetheless everyone!

PR #268, opened 2014-09-29
Added server templates

tk421:
Added a server template in the documentation for Ubuntu 14.04 / CentOS 7 and Amazon Linux. This allows users to:
- Have working instructions to install the application from scratch
- Create a temporarily server with a working version of django-helpdesk to play around
- Deploy a production server with djando-helpdesk 

I think this is a great add-on for the documentation. Novice users will be able to run the application with no problem and advanced users has templates that they just need to customise.

Feedback is appreciated!

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -3,6 +3,13 @@ Installation
 
 django-helpdesk installation isn't difficult, but it requires you have a bit of existing know-how about Django.
 
+Server templates
+----------------
+Server templates allows to create a temporary server to test djando helpdesk. You can also deploy production servers and fork configurations for customization.
+
+- Ubuntu 14.04: https://manageacloud.com/cookbook/django_helpdesk
+- CentOS 7: https://manageacloud.com/cookbook/django_helpdesk_centos_7
+- Amazon Linux: https://manageacloud.com/cookbook/django_helpdesk_amazon_2014032
 
 Getting The Code
 ----------------


reduxionist:
Personally, I would prefer to take on responsibility for supporting a widely used OSS deployment tool such as Docker, Vagrant, Ansible, etc., rather than for a SAAS's free-tier. 

Can we get a vote from @flinz or @gwasser ?

As for the analogy to GitHub, I see the following differences:
- GitHub launched when there was still no dominant resource for effective social coding, and has since established itself as the gold standard in code collaboration services.
- GitHub's continuing respect of, and contributions to, the OSS community has earned it a far greater level of trust from that community than many of its members would normally grant to a proprietary SAAS.
- The network effect is so highly relevant to this particular form of collaboration that, speaking practically, not using GitHub for an active OSS project is a disservice to that project's future potential.

None of these exceptions to the inherent tension between OSS and proprietary SAAS appears to hold true of manageacloud, however beneficial their product may actually be.

gwasser:
I'm a big free software proponent and tend to agree with @reduxionist on preferring OSS solutions rather than proprietary. I would definitely frown on integrating custom code for non-OSS SAAS interoperability.

Since this is only documentation though, I would be OK with integrating it if were made more clear that it is not an official django-helpdesk thing. Maybe add a separate "Unsupported methods" section in install.rst, or even an install-unsupported.rst file?

reduxionist:
Putting the manageacloud references into an "Unofficial" or "Third-Party" methods section at the bottom of install.rst would be an acceptable compromise for me.

The actual templates themselves, however, say they're for Django 1.6.1 which we no longer support, so they should get an update before we publish any links to them.

gwasser:
Yep, @tk421 if you would like to please update the manageacloud templates for Django 1.8+ and add it to an "Unsupported 3rd Party Methods" section. Thanks for your work!

reduxionist:
It's been one week since the criteria for merger were established. If in the next 30 days there's still no update as to whether those changes will be forthcoming or not then we will close the PR until such a time as someone wishes to reopen with new activity...

reduxionist:
As the thirty-days notice has passed, I'm closing this PR for now. Anyone interested is encouraged to re-open it with up-to-date templates. Thanks in any case!

PR #281, opened 2014-11-25
Django 1.7 compatibility for UserSettings

RafaelOrtiz:
Moving CreateUser settings to signals.py, changing name of the signal for Django 1.7 compatibility.


djcoin:
Hi there!
Just to say that I'm looking forward for this two patches to be merged!
Thanks for your amazing work
Cheers

RafaelOrtiz:
Thanks Djcoin!. as i said in an earlier comment, just some commits are made to be upstreamed. ;)

rossp:
Thanks for following up guys. As it stands this pull request is unusable for me without manually pulling it to pieces. To re-iterate my question from November: Is it as simple as changing `post_syncdb` to `post_migrate`?

As best as I can tell the changes to RafaelOrtiz@17df014 are unrelated: one moves the signal to signals.py (which is fine, but not directly addressing this issue) and it changes `blank=True` to `blank=False` on `IgnoreEmail`. I am hesitant to just pull this in as part of another fix.

Thank you.

djcoin:
Just trying to figure this out (because the two patch doesn't seem to work for me).
What did change between 1.6 and 1.7 that may trigger this bug?

Progentator:
@rossp I got it working without the change on IgnoreEmail.  Just needed to move the signal to signals.py and Django1.7 works.  As a side note, I could not get this application working at all on Django 1.5+ until this fix.  That may be due to the fact I am using LdapAuth.

Anyways.  The fix referenced here allowed me to use Django 1.7 with LDAP Authentication.  Thanks.

RafaelOrtiz:
@Progentator feel free to re-send this patch, and then this PR can be closed.

cheers.

djcoin:
I made a pull request using a data migration, this should fix it : see #287 .
The build just failed but I'm not sure how it is related to my change.
Cheers!

rossp:
Thanks guys, I've pulled in #287 from @djcoin which should negate the need for this PR if I'm understanding correctly?

PR #296, opened 2015-02-11
In Django < 1.7, if settings.AUTH_USER_MODEL is defined, the sender is e...

flinz:
...xpected to be the actual instance, not a string. This Closes rossp/django-helpdesk#295 by checking for django version < 1.7.


flinz:
ok, I will look into why the build failed.

PR #298, opened 2015-02-11
Bugfixes for Django 1.6 backwards compatibility.

flinz:
Added south migration to populate the initially existing users with usersettings - Closes rossp/django-helpdesk#297. 
In Django < 1.7, if settings.AUTH_USER_MODEL is defined, the sender is expected to be the actual instance, not a string - Closes rossp/django-helpdesk#295 by checking for django version < 1.7.


flinz:
Ok, fixed a small bug (missing import) that I overlooked and opened a combined request that fixes two issues.

Issue #293, opened 2015-02-11
Creating Custom Form Fields based on Queue rather than Ticket Model


JJediny:
Testing on this app in another django project and it's integrated well; however there is one last project requirement not supported. Considering supporting this myself but ignorant to whether its practical in the context of the apps current logic/scheme. I need to be able to customize the form based on the Queue rather than how it's currently setup; appended to the Ticket model using a Mixin. Before diving deeper I'm wondering if there are any fatal flaws to changing/relating the CustomMixin to Queue model vs the Ticket model? The basic idea would be to make the user select the Queue field first and tie a separate view/form to render based on the selection {{ Ticketfields }} + {{ QueueCustomMixin }}

rossp:
Like most changes like this I think it's feasible, IF the normal use-case of system-wide custom fields stays the same. The biggest impact is that it obviously impacts the flow for creating a new ticket for both the public and staff views.

To keep it simple for end users, can we load the remainder of the form dynamically with JavaScript when the queue is changed? Need to ensure that a no-JS version also works (ie redirect to a second page with the new form).

JJediny:
@rossp was able to do some work towards this https://github.com/JJediny/django-helpdesk/compare/rossp:master...master. Code is still fragmented as I'm still prototyping/debugging but here is the basic approach I'm taking (so far - any feedback would be appreciated)... 

## Approach for Feedback
### Adding a New Model TicketType 

`Queue` seems more akin to Office/Department/Mailbox than the "type" of ticket/customfields, so for this reason and so I wouldn't disturb the existing framework I created a new abstract model class called `tickettype`. I stole heavily from [django.blog.zinnia](https://github.com/Fantomas42/django-blog-zinnia/blob/develop/zinnia/models/category.py) 'category' which uses mptt model class (already a dependance in django-helpdesk) to create an abstract/flexible/hierarchical model. 

### Relationships to Existing Models

[...]

PR #308, opened 2015-02-23
Added ManytoMany relation from Queues to CustomFields

JJediny:
Rethought/reworked the approach; think you both @rossp @flinz we're right about #293 . Thanks for reviewing - still have my django training wheels on so tried to emulate something I found worked else where... But realized my work last week was WAY too complicated... So instead reevaluated the approach and kept it simple... the django project I'm using as a base for the app is (GeoNode)[https://github.com/GeoNode/geonode]  which has dropped using South for migrations as the models are still dynamic *not my call. I tried adding but it gave me some major errors so sorry I couldn't follow the contributing guidelines to include the migration files w the commit. Have some work else where on the project but should be coming back to it sometime this week with the javascript... but thinking I'll have the split the forms.py file (inline) so it can be controlled by JS via div/template. This PR is a placeholder for the relationship and admin view to be exploded

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 *.pyc
+build/*
 dist
 django_helpdesk.egg-info
 docs/html/*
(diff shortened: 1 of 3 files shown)


timthelion:
While this isn't a terrible change it's not actually complete, it doesn't actually filter the custom fields by queue. It is also a bit old.

PR #314, opened 2015-03-27
Preliminary support for python 3 & django 1.7

atmaniak:


rossp:
Thank you! Looks like this should be >2.7 compatible, have you tested on 2.x? (This is where I wish I'd started with proper tests years ago :( )

atmaniak:
nope, i don't have any python 2.x installation :( But the code must be compatible with python 2.7.

atmaniak:
Any idea/timeframe to integrate theses changes ?

rossp:
I'm not willing to merge this until we can do some basic testing on Python 2.7. "the code must be compatible" is not enough, sorry.

I appreciate that we don't have a formal test suite, but at the bare minimum I'd expect somebody to have installed it on Py2.x with Django 1.6 and confirmed that it runs alright.

atmaniak:
Hi,
i did some testing, everything seems to be running correctly on Ubuntu 15.04, python 2.7 & Django 1.6.
(py27):~/src/clean$ django-admin.py --version
1.6.11
(py27):~/src/clean$ python -V
Python 2.7.9
(py27):~/src/clean$ python manage.py validate
0 errors found

I didn't test the mail stuff, since i don't have changes in this part !

rossp:
Thank you for that - what about `runserver` then create, update, close tickets? `validate` won't capture any issues within views, only model and configuration issues.

atmaniak:
I was saying previously, everything seems to be working correctly : ticket creation, update, tacking, changing status, ...

The only thing i didn't test is the mail part (POP/IMAP fetching, ...)

atmaniak:
I've made somes mistake in templatetags commenting everything. I'll fix it in a few minutes...

(Sorry)

PR #320, opened 2015-04-16
Remove deprecated load url from future

tony-zhu:
Django 1.8 deprecated the load url from future.

According to Django 1.5 release notes:
https://docs.djangoproject.com/en/1.8/releases/1.5/

This template tag line could be removed safely from that release.

Since the requirements.txt says "Django>1.4". It should be safe to merge this.


rossp:
Thank you! Now that 1.8 is out I think we should require Django >=1.6 - ie support the current release and two most recent... obviously we have to get 1.8 compatibility right first though.

PR #239, opened 2014-07-28
Custom staff filter

brente:
Allow customisation of the function used to validate the user for access to "staff-only" areas

Issue #200, opened 2014-01-29
Review settings.py contents


rossp:
There are lots of items in settings.py that either appear inconsequential, or that control template output (ie menu items or footer config). 

My preference is that we remove these and people override them with custom templates as required.

uhurusurfa:
It is not clear why user customisation that requires developer skills is preferred over configurable behaviour.
The key to a successful app is being able to adapt to different users needs without requiring technical knowledge. 
Once the configiuration is in place it becomes reusable whereas customisation by end users is only bavailable to those who have the skillset to do it.

Closing this ticket as it requires significant dicscussion on where Heldesk will go and perhaps better done as a [Discussion](https://github.com/django-helpdesk/django-helpdesk/discussions)

PR #323, opened 2015-04-28
Users public tickets

atmaniak:
Add a new setting HELPDESK_USERS_TICKETS_PUBLIC allowing logged in user to view previous tickets.

### DIFF
--- a/helpdesk/templates/helpdesk/public_homepage.html
+++ b/helpdesk/templates/helpdesk/public_homepage.html
@@ -21,6 +21,40 @@
 </table>
 {% endif %}
 
+{% if helpdesk_settings.HELPDESK_USERS_TICKETS_PUBLIC %}
+{% if user.is_authenticated %}
+<div class="col-xs-12">
+<div class="panel panel-default">
+
+<div class="panel-body">
+    <h2 name='submit'>{% trans "My tickets" %}</h2>
+<p>{% trans "You can view your previous tickets here." %}</p>
+<div><table class="table table-bordered">
+<tbody><tr>
+                      <th>{% trans "Title" %}</th>
+                      <th>{% trans "Status" %}</th>
+                      <th>{% trans "Action" %}</th>
+                    </tr>
+
+{% for ticket in tickets %}
+       <tr>
+            <td>{{ ticket.title }}</td>
+            <td>{{ ticket.status }}</td>
+            <td><a href={% url 'helpdesk_public_view' %}?ticket={{ ticket.queue.title}}-{{ticket.id }}&email={{ request.user.email }}>{% trans "Show ticket" %}</td>
+        </tr>
+{% empty %}
+<tr><td>{% trans "You don't have any tickets" %}</td><td></td><td></td></tr>
+{% endfor %}
+    </tbody>
+    </table>
+</div>
+</div>
+</div>
+</div>
+{% endif %}
+{% endif %}
+
+
 {% if helpdesk_settings.HELPDESK_SUBMIT_A_TICKET_PUBLIC %}
 <div class="col-xs-6">
 <div class="panel panel-default">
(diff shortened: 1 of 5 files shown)


rossp:
Thank you for your work on this, although based on #200 I'm trying to reduce the number of settings-controlled functions within `django-helpdesk`.

Would this be better addressed by working on #239 and getting to a point where a custom staff filter can be configured that allows all users to be seen as staff?

JayTurnr:
This needs to be accepted, settings aren't a bad thing @rossp and this feature is huge

gwasser:
Per @rossp's comments, #239 was merged for the 0.3 release so it might only take a little more work to develop a filter that treats all users as staff. That may be a better way to implement this than introducing a new setting.

JayTurnr:
Maybe I'm misinterpreting what this PR does. Like I said, I glimpsed. I'm after an option of a non-staff user to quickly see all tickets they've opened.

blackboxoperations:
The PR allows for registered users to submit a ticket, then be shown only their tickets on the main page. They're able to view the ticket's details and subsequent follow-ups posted by admin/staff users, as well as select a button that approves and closes the ticket once the ticket has been marked as resolved, but...

A handful of gotchas you'll have to work out (helpdesk:public_view and not helpdesk_public_view in the changes to the public_homepage.html template, for example), and there's an infinite loop on 'submit a ticket', but if you know your way around a codebase, you can take this PR and bend the current version of Helpdesk well enough to give your end-users the ability to submit and follow along with a ticket's movement up until it's resolved and they double-verify then close it for you. 

This dynamic is non-obvious, so be careful. I'm with the original developers/authors on this one, in that you don't want to just up and make ticketing systems publicly accessible/available without some thought and well understood configuration. It's not so much about the code, it's much more about the business requirements and hazardous use-case scenarios re: management, data security, etc. <-- just my two cents, ignore accordingly if you hate my face, it's cool.

helpdesk/templates/helpdesk/public_homepage.html
```diff
@@ -21,6 +21,40 @@ <h2>{% trans "Knowledgebase Articles" %}</h2>
...
+                      <th>{% trans "Title" %}</th>
+                      <th>{% trans "Status" %}</th>
+                      <th>{% trans "Action" %}</th>
+                    </tr>
+
+{% for ticket in tickets %}
+       <tr>
+            <td>{{ ticket.title }}</td>
+            <td>{{ ticket.status }}</td>
+            <td><a href={% url 'helpdesk_public_view' %}?ticket={{ ticket.queue.title}}-{{ticket.id }}&email={{ request.user.email }}>{% trans "Show ticket" %}</td>
```
blackboxoperations: change to helpdesk:public_view

gwasser:
Thanks for the feedback! I agree that the use cases for non-staff users have not been totally thought out yet, as evidenced by our lack of real permissions checks beyond "are you staff?". So to fully merge this PR will take some thought on requirements.

Grunze1s:
Why isn't this feature been looked up ?

Mte90:
I think that to be moved on requires some updates to the project documentation.

gwasser:
There were some questions about user permissions that were unresolved, but I admit I'd have to dig back into this for details because I don't remember exactly.

Mte90:
Probably is better to close it and moving the discussion on a ticket with all the various feedback and a reference to do a new implementation.
I think that this PR after all those years have a lot of conflicts and maybe doesn't work anymore.

gwasser:
You do make a good point @Mte90 that this PR would probably take significant work to merge into what helpdesk is today, it might be better to start over with a better sense of how to implement. For now I'll keep this ticket open until we figure out what we want to do next though.

timthelion:
This is badly out of date and will be implemented in a new PR. Thank you for your effort.

PR #328, opened 2015-06-03
Remove migration 0002_socks_proxy.py...

reduxionist:
...which has been squashed into 0001, rename migrations 0003 and 0004, and update their dependencies accordingly.

Problem that prompted this PR:
A fresh install of django-helpdesk/master fails at the manage.py migrate stage, when running migration 0002.

Diagnosis:
The contents of migration helpdesk/migrations/0002_socks_proxy.py were added to helpdesk/migrations/0001_initial.py by commit 70c57f90, presumably as the result of squashing migrations. Since 0002 was not removed at that time, when running through all migrations the fields are now created in 0001_initial and thus are unable to be added by 0002.

Solution:
As per commit message: remove 0002, rename 0003 and 0004 accordingly, update dependencies.

This issue affects Django 1.7+'s built-in migrations, I do not know if it is present for south-based setups which use a different set of migrations.

If you prefer some other resolution, e.g. squashing 0003 and 0004 into 0001 rather than renaming & updating dependencies, please let me know and I will adjust this PR accordingly.

Thanks for sharing your code!


rossp:
I'm comfortable with this approach, thank you - I have a feeling the entire migratino set needs to be redone at some point as we've had issues for a while, although hopefully this helps address things.

PR #344, opened 2015-10-19
Added french translations

Ardrake:
I have added missing french translation.

Bonne journée !

André Cooke


rossp:
Thanks for your submission! I have copied these new strings onto the [French translation at Transifex](https://www.transifex.com/rossp/django-helpdesk/) and will pull them into source soon. We can't accept translations via pull request sorry as the work gets lots when we update from Transifex.

Ardrake:
Excellent!

PR #349, opened 2015-11-12
From gettatr to get, because dict doesn't contatin attributes.

ErhoSen:
ticket.assigned_to.usersettings.settings is a dict, not an object.


ErhoSen:
The same as here https://github.com/rossp/django-helpdesk/blob/master/helpdesk/forms.py#L293

PR #351, opened 2015-11-13
Django 1.8 compatibility including updated testrunners

flinz:
Towards preparing for 1.8 compatibility, I had to adapt our test suite runner ``quicktest.py`` and rename the tests.

https://docs.djangoproject.com/en/1.8/internals/deprecation/#deprecation-removed-in-1-8
> The module django.test.simple and the class django.test.simple.DjangoTestSuiteRunner will be removed. Instead use django.test.runner.DiscoverRunner.

This required the following changes:
* **quicktest.py**: Added switch for Django version, now importing the django.test.runner.DiscoverRunner if deprecated import raises an Exception.
* **test/**: Renamed the tests themselves to comply to the DiscoverRunner default ``test*.py`` naming scheme, thus making them discoverable by the DiscoverRunner

This change was tested in django 1.6, 1.7 and 1.8.


flinz:
This does not yet seem to work. Testing locally, this runs all 10 tests. Travis runs 0.
I will have to look into this, sorry.

flinz:
I updated the tests discovery, which should now also run correctly in django 1.5.

* I propose to add Django 1.8 as a build target on travis once this is merged
* In the near future, we should discontinue support for at least Django <= 1.6, as discussed previously here https://github.com/rossp/django-helpdesk/issues/335#issuecomment-132828237

flinz:
Just as a note: in Django == 1.7 one of the added migrations is reversed on ``./manage.py makemigration`` (1.8 changed the max-length of the email fields).

flinz:
@reduxionist @ssadler could you guys check this PR?

ssadler:
Looks great.

rossp:
Thank you! 

I will merge this in now but not do a Pypi release until we have a bit more testing done. Given the impending release of Django 1.9 I have no issue cutting off support for 1.6 and earlier since that'll simplify this even further.

Issue #345, opened 2015-11-02
Stats not working when user name or queue names have utf-8 characters


Ardrake:
Django 1.7.8
Phthon 2.7.8
psycopg2==2.6.1 -> Postgres database 

Everything is working great, with the exception of the stats screens, they remain at 0, for all stats.

have tried these settings, to no avail

HELPDESK_NAVIGATION_STATS_ENABLED = True
HELPDESK_ENABLE_PER_QUEUE_MEMBERSHIP = True

Ardrake:
When switching to english language I get partial stats, all stats not assigned to a user are showing up.

reduxionist:
Andre,

Unless you intend to restrict users to only be able to see those queues you have explicitly granted them access to, you do not want:

Ardrake:
Thanks for the quick response, ok, switched to 

HELPDESK_ENABLE_PER_QUEUE_MEMBERSHIP = False

Still not wroking as intended as en example

LANGUAGE_CODE = 'fr-CA'
![screen02](https://cloud.githubusercontent.com/assets/4698209/10908445/03b9b290-8200-11e5-9856-e10ad19d89af.jpg)

LANGUAGE_CODE = 'en-US'
![screen01](https://cloud.githubusercontent.com/assets/4698209/10908457/1246fac0-8200-11e5-9dea-d8bac9cda72e.jpg)

I'm thinking perhaps an encoding issue, then again this is a guess, I have not looked deeper in the issue.

reduxionist:
The next step then would be to install the [django-debug-toolbar](https://github.com/django-debug-toolbar/django-debug-toolbar) and compare the debug output (esp. SQL queries) generated by the English and French versions. Bon chance! :)

Ardrake:
Changed my name in the admin file, removed the utf-8 characters, André switch to Andy
and voila, I have some stats to my name, the Queries under django-debug-toolbar are identical.

LANGUAGE_CODE = 'en-US'
![screen03](https://cloud.githubusercontent.com/assets/4698209/10908975/97bdf17e-8203-11e5-9c00-82d56f744e28.jpg)

LANGUAGE_CODE = 'fr-CA'
![screen04](https://cloud.githubusercontent.com/assets/4698209/10909049/1687f95a-8204-11e5-9d1d-56745ad16ef4.jpg)

rossp:
Thanks for narrowing this down. I'll try to investigate in the next 24 hours.

flinz:
[...]

PR #356, opened 2015-11-18
Python3 compatibility

flinz:
* tested on python 2.7, 3.3, 3.4 and 3.5 with django 1.8.6
* changed model's ``__unicode__`` methods to ``__str__`` in backwards compatible fashion
* fixed problems arising from distinctions between ``byte`` and ``str`` which broke some tests - had to remove explicit calls to ``.encode('utf-8')`` particularly in the recent [per-user queue code](https://github.com/flinz/django-helpdesk/blob/8d55dcb9a0bf0841147d53f7755d70b93e7c8c61/helpdesk/views/staff.py#L1083). @reduxionist could you have a look at this? tests pass and everything seems functional.

I would prefer that maybe people give this a little test drive (@rossp, @reduxionist ?) before merging.

### DIFF
--- a/helpdesk/lib.py
+++ b/helpdesk/lib.py
@@ -21,7 +21,6 @@ import logging
 logger = logging.getLogger('helpdesk')
 
 from django.utils.encoding import smart_str
-from builtins import str as text
 
 def send_templated_mail(template_name, email_context, recipients, sender=None, bcc=None, fail_silently=False, files=None):
     """
@@ -119,7 +118,7 @@ def send_templated_mail(template_name, email_context, recipients, sender=None, b
             "subject": t.subject,
         }).render(context)
 
-    if isinstance(recipients, text):
+    if isinstance(recipients, str):
         if recipients.find(','):
             recipients = recipients.split(',')
     elif type(recipients) != list:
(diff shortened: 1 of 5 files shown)


flinz:
See also the discussion about remaining to-do's for a python3 compatible release @ https://github.com/rossp/django-helpdesk/issues/290.

reduxionist:
Although they're adjacent to it, the explicit calls to .encode('utf-8') were not actually introduced by the per-user queue code. git log -S".encode('utf-8')" suggests they were most recently introduced by commit ac95f9d ~4.5 years ago.

While I don't have any special insight as to their purpose, I still took a look nonetheless and the only impact that occurs to me is how it might effect queues with non-ASCII titles. I suspect, however, that any problems caused by this could be addressed as part of the existing issue encountered in #345, i.e. thorough i18n/UTF support.

flinz:
@reduxionist I had a similar problem as in https://github.com/rossp/django-helpdesk/issues/345, since we were comparing the utf-8 encoded byte strings to unicode strings, which broke things in python 3. now everything remains in unicode.. I will have a look at the issue and see if this might even have fixed the problems there.

PR #361, opened 2015-12-15
Bugfix: Integer given as string in models.py

flinz:
Bugfix, related issue is https://github.com/rossp/django-helpdesk/issues/360.
This reportedly caused problems under postgresql.

Thanks to @alexbarcelo for spotting.

### Changes
- changed max_length given by string to integer;


flinz:
Will merge this into the Django 1.9 compatibility branch, since it only occurs under Django >= 1.9 https://github.com/rossp/django-helpdesk/issues/358

PR #363, opened 2015-12-15
Use [auth.models.]Permission approach instead of QueueMembership

alexbarcelo:
This proposal simplifies some code by relying on the `django.contrib.auth.models.Permission` approach on managing access permissions. The previous approach which used QueueMembership is deprecated[1].

When a Queue is created, a new Permission is prepared and the administrator may add that permission to user _and groups_.

In addition to that, some extra isolation is done regarding tickets: now tickets from foreign queues are banished from the user's ticket list, and it should be safe from malicious behaviour.

The existing test on permission has also been adapted to this new approach and it works as expected.

There is a feature lacking: recreation of permission --e.g. an administrator decides that permissions are required, there is no mechanism yet for recreating Permission objects for existing Queues. However, as django-helpdesk already implements some specific `./manage.py` commands, this should be straightforward. If this PR is seen as beneficial I will add this feature.

[1] I was thinking in maintaining dual behaviour, but couldn't think of a clean way of doing so. And it would be more extra settings, which maybe is overkill. As the Permission approach superseeds the features given by the QueueMembership approach, it should not be very problematic. I am not sure if it is possible to prepare a migration to cope automatically with those scenarios, I am not very familiar with fancy migrations.


alexbarcelo:
I am preparing a new PR, with more commits and features. However, it depends on #366 because migrations are required and it can be messy to ignore it. For the moment, I'm closing this and working on alexbarcelo/django-helpdesk. I will soon proceed to send the PR

PR #364, opened 2015-12-16
Change supported django+python versions

flinz:
- support only python 3.4+ (3.3 support is dropped for Django 1.8+) (issue: https://github.com/rossp/django-helpdesk/issues/362)
- drop support for django 1.7 + python 3.5 (not supported)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: python
 python:
     - "2.7"
-    - "3.3"
     - "3.4"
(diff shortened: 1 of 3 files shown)


flinz:
The one remaining error is a known error with Django 1.7 and Python 3.5.. https://code.djangoproject.com/ticket/23763 or http://www.thefourtheye.in/2015/02/python-35-and-django-17s-htmlparseerror.html

Issue #360, opened 2015-12-15
Bug: Integer given as string in models.py


flinz:
Thanks to @alexbarcelo for [spotting this](https://github.com/rossp/django-helpdesk/issues/355#issuecomment-164671890).

https://github.com/flinz/django-helpdesk/blob/django-1.9-fixes/helpdesk/models.py#L1216 has ``max_length`` given as a string. This seems to work for MYSQL, however might cause problems in PostgreSQL.

Since it's the only instance of max_length given by a string in models.py, I suggest to just change it to int.

alexbarcelo:
This seems to prevent any database operation (`syncdb` or migrate-related stuff) in my production environment, which is Django 1.9 and PostgreSQL.

Changing `Django==1.8` in my requirements seems to make it work (maintaining the Postgres database). So I am not sure if it is some combination of factors or what.

However, as you said, it is the only instance of `max_length` given by a string, so your bugfix seems the sensible approach. Thanks a lot!

flinz:
Ah, in this case I will move the bugfix into the django 1.9 PR.

alexbarcelo:
wait, wrong test, give me a couple of minutes, maybe I'm mistaken

alexbarcelo:
Sorry about that, I had just realized that the `requirements.txt` was not in a "consistent state", and wanted to make sure that what I said it was true. Now I have checked with Django==1.8, master branch and Postgres and it seems to work. So the is indeed for Django 1.9 compatibility, related to Postgres usage it seems.

Thanks again and sorry for the fuss

flinz:
Ok, thanks for double checking.

PR #358, opened 2015-12-10
Django 1.9 compatibility

flinz:
Now that Django 1.9 has been released, its time to update our supported Django versions - didn't have to fix much. This branch has been tested in Django 1.9, 1.8 and 1.7 under python 2.7.10.

Python 3 compatibility is introduced in my other PR https://github.com/rossp/django-helpdesk/pull/356.

Related issue is https://github.com/rossp/django-helpdesk/issues/355.

PR #366, opened 2015-12-16
adding migration for bugfix 'changed max_length given by string to integer;'

alexbarcelo:
PR #358 included a bugfix (discussed in #360) that requires a migration.

This commit includes the trivial migration, which should be safe to apply as no real changes are made.


flinz:
Thanks! The failing builds will be gone once https://github.com/rossp/django-helpdesk/pull/364 is merged.

Issue #362, opened 2015-12-15
Towards the next release (0.1.17): open PR's and things to do


flinz:
Hey @rossp, I think we might be collecting a few compatibility PR's that would probably merit a new release, keeping up with Django versions as well as adding Python3 compatibility. 

Since it was a little confusing to me, I chose to start this issue - here's a little overview of open PR's and related things to do towards a release. Would you have any thoughts or extensions on what we should include? 

Maybe it would be worth opening a milestone for the next release and add these todo items as issues, to sort this out a little?

 - [x] Update docs & installers to only support Python 2.7/3.3-3.5 (Issue: https://github.com/rossp/django-helpdesk/issues/290, PR: https://github.com/rossp/django-helpdesk/pull/356)
 - [x] Update docs & installers to only support Django>=1.7 (Issue: https://github.com/rossp/django-helpdesk/issues/355, PR: https://github.com/rossp/django-helpdesk/pull/358)
 - [x] Update travis testing to include Python 2.7 as well as 3.3-3.5
 - [x] Update travis testing to include Django 1.7, 1.8 & 1.9
 - [x] Fix broken tests :) https://travis-ci.org/rossp/django-helpdesk/builds/97108228
 - [ ] Couple days quick testing
 - [ ] Update release approx ETA 21st December 2016

rossp:
I've merged the big changes for Django 1.9 and Python 3.

Let's give this a couple of days then update the Pypi release - would rather get some more thorough testing of both of these done prior to updating the installer.

Note Travis is currently failing on some combos of Django >=1.8 and Python 3. Needs quick review.

 - [x] Update docs & installers to only support Python 2.7/3.3-3.5 (Issue: https://github.com/rossp/django-helpdesk/issues/290, PR: https://github.com/rossp/django-helpdesk/pull/356)
 - [x] Update docs & installers to only support Django>=1.7 (Issue: https://github.com/rossp/django-helpdesk/issues/355, PR: https://github.com/rossp/django-helpdesk/pull/358)
 - [x] Update travis testing to include Python 2.7 as well as 3.3-3.5
 - [x] Update travis testing to include Django 1.7, 1.8 & 1.9
 - [ ] Fix broken tests :) https://travis-ci.org/rossp/django-helpdesk/builds/97108228
 - [ ] Couple days quick testing
 - [ ] Update release approx ETA 21st December 2016

flinz:
this is a little weird. most of the python 3 errors are due to errors involving this 
[...]

PR #367, opened 2015-12-16
Usage of [contrib.auth.models.]Permission instead of [helpdesk.models.]QueueMembership

alexbarcelo:
This series of commits change the previous behaviour of "QueueMembership", which could be used to give access to certain ticket Queues, into a more ACL-like behaviour based on the Django `contrib.auth` package `Permission`.

To achieve this, a Permission is generated for each Queue, and the user is checked that permission in order to access a ticket.

In addition to the code, this PR includes an automatic migration for existing settings and also a management command to generate the permission in bulk. Once the appropriate settings flag is set, the new Queues will trigger the creation of their Permission.

Permissions can be assigned to either user or groups by the built-in User view in the admin (no changes needed). The Permission can be found under "helpdesk | Queue | Permission for queue: <title of Queue>", and internally has a codename of "queue_access_<slug of Queue>".

Removal of Queue remove the permission (given that the removal is not a batch removal).


flinz:
The migration from perqueuemembership looks tight to me. I also think the usage of Django permissions is a little cleaner. I'll give this a little testrun, but so far

:thumbsup:

flinz:
@alexbarcelo i can't seem to get the migrations to run. specifically, i am trying to migrate a db with some previous queue memberships ([here is the file](https://github.com/rossp/django-helpdesk/files/65537/db.sqlite3.zip)).

During ``./manage.py migrate`` I get the following error
```python
File "django-helpdesk/helpdesk/migrations/0009_migrate_queuemembership.py", line 29, in create_and_assign_permissions
    basename = q.prepare_permission_name
AttributeError: 'Queue' object has no attribute 'prepare_permission_name'
```

Could you have a look and make the migrations work?

Just by checking it out, I believe that using the ``RunPython`` function will give you the historical class, i.e. the old ``Queue`` that does not have the ``prepare_permissions_name`` function. You probably have to import the new class in addition, i.e.

``from helpdesk.models import Queue as NewQueue``
and then use ``NewQueue.prepare_permissions_name``

alexbarcelo:
Aaaaah! Sh**. Ok, I did this last commit, and didn't realize it may break the previous migration (as it actually did). I had tested the migration, but without the prepare_permission_name.

I think that it will be cleaner to add that prepare_permissions_name and the extras in `adding required model extra info for permissions`. I will check if reordering the branch is cleaner, maintining the general approach. Otherwise I will add this `import NewQueue`.

flinz:
On a second note, I think you probably should provide a forward and a backward data migration.
And you should combine all of this into one single migration.

I.e.
```python
class Migration(migrations.Migration):

    dependencies = [
        ('helpdesk', '0007_max_length_by_integer'),
    ]

    operations = [
        migrations.AddField(
            model_name='queue',
            name='permission_name',
            field=models.CharField(blank=True, help_text='Name used in the django.contrib.auth permission system', max_length=50, null=True, verbose_name='Django auth permission name'),
        ),
        migrations.RunPython(forward_data, backward_data),
        migrations.RemoveField(
            model_name='queuemembership',
            name='queues',
        ),
        migrations.RemoveField(
            model_name='queuemembership',
            name='user',
        ),
        migrations.DeleteModel(
            name='QueueMembership',
        ),
    ]
```

alexbarcelo:
With this I was able to perform a migration and rollback and it seemed to work.

The migration backwards won't be able to fix group permission, because QueueMembership doesn't understand them, but typical scenarios will work ok, I expect.

rossp:
Wonderful, thanks!

I'm going to merge this in so we can play with it a bit before doing a release with Django 1.9 / py3 / etc on Monday (ref #362)

alexbarcelo:
@flinz Sorry, about joining the migrations, I haven't done it because I was fearing that it will break. I am not sure which order is done, and the migration requires some extra files to perform de `RunPython` and the deletion should not be done until the `RunPython` has finished. 

I don't know what is advised in this situations. If performing a `squashmigration` is safe and you think it is the right way, I will add it --no problem about that. Simply that I was doubtful about correctness and stylishness

alexbarcelo:
Thanks @rossp! I have played a bit more and I think that "it feels right", and migration should be straightforward. Maybe there are some bugs about ACL and malicious behaviour may break it. But the typical actions are ok, so I hope it only needs a bit testing to be on the safe side.

@flinz Sorry that I didn't elaborate on your comment about combining into a single migration. I was fearing that it would be strange because the `RunPython` requires the `permission_name` and should be executed before the QueueMembership objects aren't gone. I am not proficient enough on migrations, and this way seemed to be the "safe/conservative way" --although you already found me shooting myself in the foot :-/

PR #370, opened 2015-12-18
Making Queue slug unique and improving certain admin-related behaviour

alexbarcelo:
Most changes are improvements to the admin related to Queues (mainly the auto slug generation in the admin Queue creation).

However, the slug uniqueness is a restriction to avoid bugs between different Queues with the same slug (which, by default, I expect to be a non-desired scenario... it would be awkward, wouldn't it?)


flinz:
I think making the slug unique is indeed a desired behavior.
:+1:

PR #374, opened 2015-12-22
Set app_name as 'helpdesk' at the urls.py, and droping url prefixes

alexbarcelo:
Seems a correct way to manage the whole application, instead of relying on prefixes. https://docs.djangoproject.com/en/1.9/topics/http/urls/#namespaces-and-include

Moreover, I realized that the helpdesk installation was interfering with my own `login` view, because they were clashing the namespace. Using the `app_name` seems to behave nicely.

I realize that this will break some existing installations, but it seems a more consolidated approach --for instance, it feels more familiar with the way the `admin` namespace works, or the way the `api` namespace behaves in the Django Rest Framework.

There is some behaviour divergence between Django 1.8 and Django 1.9. I tried to update the documentation based on my own rudimentary tests, but I would love to hear some other experiences and opinions.


alexbarcelo:
I have horribly messed up something, and I don't understand why even Django 1.9 is failing... Looking into it.

flinz:
@alexbarcelo could you look into solving the open conflicts (i.e. merging master into your branch)?

alexbarcelo:
I haven't been able to test it thoroughly, but Travis seems to be ok with the update. Now it should be easier to merge, I simply solved the conflicts and fixed some extra `reverse` staff.

I hope that I did not mess up the pull request... this invalidated the review, sorry if I caused trouble.

reduxionist:
@alexbarcelo No trouble caused at all, we greatly appreciate your continued efforts on this PR!

@gwasser Putting breaking changes into the 0.2 branch definitely sounds sensible to me!

alexbarcelo:
The last commit are built on top of #438. If I'm not mistaken, all the files in which this PR affects are already PEP8, so that is out of the way. The #438 should be addressed first, though.

I am also waiting for changes (expecting minor ones) due to #436

reduxionist:
Using namespaced URLs gives us the flexible and compatible routing one can expect of modern Django apps, but which just wasn't available when a long-lived project like this first wired up its views. Thanks very much @alexbarcelo !!

alexbarcelo:
Thanks @reduxionist for your merging marathon! It is nice to have all those shiny new features in django-helpdesk :3

PR #379, opened 2016-01-26
Fix user_admin_url template tag for Django >= 1.8

Tyris:
Untested (I don't have an environment I can test in handy at the moment).

If someone could confirm it works before merging.


powderflask:
Alternate, tested pull request here: https://github.com/rossp/django-helpdesk/pull/380

Tyris:
Use tested #380

PR #381, opened 2016-02-16
Extended Queue model with 'default_owner' field

imbashamba:
Hello!

It's a common usecase that public ticket needs to be automatically assigned to a person, cause he's responsible for all the tickets (e.g. head of support department). 

I've added a nullable default_owner field to Queue model, code to assign a ticket from public form to that person, email notification of that assigned person.

Hope you'll find my changes usefull.

### DIFF
--- a/helpdesk/templatetags/user_admin_url.py
+++ b/helpdesk/templatetags/user_admin_url.py
@@ -14,8 +14,13 @@ from django.contrib.auth import get_user_model
 
 def user_admin_url(action):
     user = get_user_model()
+    try:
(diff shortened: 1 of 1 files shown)


rossp:
Hi @imbashamba, thanks for the submission!

To keep things nice and clean, could you please submit the pull request without the translations? We can then use Transifex to build up new translations and put those in a singular PR. 

Any manual changes to .po and .mo files tend to get lost when we `tx pull`, so I'd rather not have them mixed in with code changes. Thank you :)

Issue #375, opened 2015-12-23
Adding more capabilities to mail threads and replies


alexbarcelo:
It would be nice if the tickets could be more integrated with the mail responses. Use case:
- Mail received from alpha@example.net, sends to Queue (queue_name@example.com) and also puts as CC bravo@example.net
- Bravo sends a response to both alpha@example.net and the Queue
- Charlie is staff and replies directly to the mail sent by helpdesk
- Delta is staff and puts a public comment through the web interface.

Desired behaviour:
- First Alpha mail opens a ticket. (No mail outbound). By default, no CC is added (although maybe it would be a feature? Detect bravo@example.net as a CC and add it to CC?)
- Bravo reply is stored as a comment to the ticket. (No mail outbound)
- Charlie reply is resent to Alpha. Helpdesk sends the update mails as a **mail reply** to those.
- Delta comment is also sent as a mail reply.

Internal mechanism:
- When the ticket is opened by Alpha, the Mail-ID is stored somewhere. This would be the main ID for all outgoing mails on this ticket. (The In-Reply-To mail header).
- When Bravo sends a reply, the Mail-ID is stored, and it will be added in the References mail header.
- Charlie receives a mail from helpdesk with Reply-To something like queue_name+ticket_identifier@example.com --this ensures that the Charlie reply is associated to the ticket. To avoid more guessing and more robust behaviour. The outgoing mail has In-Reply-To and References correctly populated.
- Delta comment is sent as a mail reply, in a similar fashion to the last one.

Final goal:
- The mail integration would be nicer, the end-user will see the tickets just as a regular mail thread (whatever client or webmail are they using).
- The staff are able to reply directly from mail. Don't know if there should be a mechanism to add options --like a [close] tag in the subject to force a ticket close? But, at least, it would be more easy to address simple tickets.

There will be issues, and maybe there are some problems with this approach. Do you feel that I am overlooking something?

rossp:
Generally speaking I like the idea of having more powerful email integration.

What I *don't* want to do is have email addresses that are controlled by `django-helpdesk` - ie your example of *queue_name+ticket_identifier@example.com*. This requires integrating with incoming email in ways that I have done in the past and am not in a rush to support again - every set up is completely different. I'd prefer a single email address per queue (per current functionality) and use of the subject, which seems to work in most cases already.

To address your sample scenario:
1. I think adding Bravo as a CC on ticket create is a good idea
[...]

PR #385, opened 2016-02-20
Add user threading feature - issue #375

brunotikami:
Hello,

I'd like you to have  a look at my approach to enable user threading to your tool. Please feel free to request changes and question this PR.

Keep up the good work,

Tkm


rossp:
Can you please give us a few particular pointers here - there is a huge amount of change in this PR and manually reviewing files to try to get the gist of what it's meant to do.

brunotikami:
Sure, no problem!

The changes:

- the TicketCCForm to overwrite the <clean> method so we can check if either a valid user or an email is provided.

- the <send_templated_mail> function from the libs module to accept extra headers so "In-Reply-To" can be specified

- I've added the "message_id" field to the FollowUp model to enable mail threading / tracking and the migration 0012 to support this

- the <subscribe_staff_member_to_ticket> function from the views/staff module now uses a < subscribe_to_ticket_updates>. The new method allow subscribing without a User instance if you have a valid email

- tests for a big part of this operations

- the big change: management.commands.get_email may create either a new Ticket or a FollowUp from one mail message:

    -  If the "In-Reply-To" header is found on the database or if we can find the ticket ID applying the regexp on the email's subject, we will add this to the Ticket's thread.
    -  else, it's a new Ticket.

    After creating the right object, notifications are sent using the same templates that the Queue uses. Another feature is that any email on the Cc list from the mail message gets a TicketCC object linked to the Ticket.

alexbarcelo:
You have some `print` statement that does not work on Python 3. The quickest fix should be performing the `from __future__ import print_function` and using the `print()` form.

I have not checked if there are extra python 3 problems --I have all my deployment in 2.7, so :-/

brunotikami:
Hi @alexbarcelo, I didn't change these statements in order to avoid a PR that would touch various aspects of the code. Anyway, as I'll add an functionality to this PR, I'll change that too. 

Just so you and @rossp know, I'll add an extra field / flag to the Queue model to decide when to notify users when an email is processed.

alexbarcelo:
Sorry, I saw the error in the tests and just assumed... I should have
checked.

Maybe it is better to prepare a different PR to fix those print statements
and other Python 3 stuff. To isolate different issues in different PR.

Ross, what do you think about that?

brunotikami:
It turns out, @alexbarcelo, that the print statements are not the only things I'd need to change to get the project 3.x compatible. In a quick look I've found an backwards incompatible change of the email library (email.Utils is now called email.utils) and, of course, the unicode X bytes part. I'll open the separate PR for the print statements anyway :-)

alexbarcelo:
Hey @brunotikami, are you still interested in this feature? I was very interested, but we had to drop the idea for a while because of some time constraints and overlapping project goals.

Now we wanted to resume the idea of mail integration and user threading, and just now I was looking the status on this. You started an awesome PR :+1: but I am not sure on the impact of the 0.2 branch on this.

gwasser:
I merged this in manually to the `develop` branch for the 0.3 release. Thanks very much for the effort!

Issue #162, opened 2012-12-13
fix for attachment filename problems


kotowicz:
see https://github.com/kotowicz/django-helpdesk/commit/0d84fc39908c28f33c2f1727191d37ff46f0b5f1

we need to ignore non-ascii characters (I had problems with the ° symbol). 
alternatively, we can build a file system safe string by using base64.urlsafe_b64encode but we'll lose the file extension.

PR #393, opened 2016-03-08
panel attachment mime-type error: TypeError

pawelmarkowski:
can't use a string pattern on a bytes-like object

Title changed from TEST to . Null title on attachment upload.

above 2 problems solved

### DIFF
--- a/helpdesk/views/staff.py
+++ b/helpdesk/views/staff.py
@@ -457,7 +457,7 @@ def update_ticket(request, ticket_id, public=False):
     if request.FILES:
         import mimetypes, os
         for file in request.FILES.getlist('attachment'):
-            filename = file.name.encode('ascii', 'ignore')
+            filename = file.name
             a = Attachment(
                 followup=f,
                 filename=filename,
@@ -473,7 +473,7 @@ def update_ticket(request, ticket_id, public=False):
                 files.append([a.filename, a.file])
 
 
-    if title != ticket.title:
+    if title != ticket.title and title != "":
         c = TicketChange(
             followup=f,
             field=_('Title'),


adrianboyko:
I'm running Django 1.8.5  with Python 3 and ran into this problem on every attempt to attach a second file to a ticket (interestingly, the _first_ attachment _always_ works). I applied Pawel's commit to a branch off the 0.1.17 tag and it fixed the problem.

flinz:
Nice catch - and it seems this is not covered by our tests. I will add this to the Testing project.

helpdesk/views/staff.py
```diff
@@ -457,7 +457,7 @@ def update_ticket(request, ticket_id, public=False):
     if request.FILES:
         import mimetypes, os
         for file in request.FILES.getlist('attachment'):
-            filename = file.name.encode('ascii', 'ignore')
+            filename = file.name
```
flinz: Are you sure this change does not break things in python 2.7 when using non-ascii characters in filenames?
The ``'.encode('ascii', ignore')`` effectively removes non-ascii characters from unicode.

flinz:
Upon a little research, it seems that In fact this ecode was introduced as a "fix" a while ago.
See https://github.com/django-helpdesk/django-helpdesk/issues/162

Simply removing it will not solve this issue. I strongly suggest we add a test that covers attachments, also with non-ascii filenames, then propose a solution that passes these tests.

helpdesk/views/staff.py
gwasser: django.utils.six will be required here. Py2/3 handle strings very differently.

pawelmarkowski: It is always a good idea to check. Can not we solve this problem without django.utils.six :)? I think we can :) just check this code on py2.7 it should work even there.

reduxionist:
Closing this PR since, effectively, its contents were merged through #463 and #464.

PR #399, opened 2016-06-21
Changed render_to_response() to render()

mrkiwi-nz:
This should work for django==1.9.4 plus django==1.10

Without this change, in django1.10 the 'user' object is missing from the view rendering context which results in a very hard to track down bug manifesting as ;
"NoReverseMatch: Reverse for 'helpdesk_rss_user' with arguments '('',)' and keyword arguments '{}' not found."
At first glance users may think this is the dots-in-usernames bug, but the username is totally missing.

There is a good chance i'm missing something about why 'user' is missing in the context, after reading the [release notes for django 1.10](https://github.com/django/django/blob/master/docs/releases/1.10.txt) I couldn't see anything obvious.
helpdesk seems to be running fine under django==1.10a1 with these edits.
Regards
Daryl.

### DIFF
--- a/helpdesk/views/api.py
+++ b/helpdesk/views/api.py
@@ -19,7 +19,7 @@ try:
 except ImportError:
     from django.contrib.auth.models import User
 from django.http import HttpResponse
-from django.shortcuts import render_to_response
+from django.shortcuts import render
 from django.template import loader, Context
 import simplejson
 from django.views.decorators.csrf import csrf_exempt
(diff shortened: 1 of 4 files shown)


pawelmarkowski:
Reverse for 'helpdesk_rss_user' with arguments '('admin',)' and keyword arguments '{}' not found. 0 pattern(s) tried: [] -- error from current helpdesk version dj 1.10.4

Issue #400, opened 2016-06-21
django.contrib.sites is required


mrkiwi-nz:
The documentation makes it clear that the following are required

{INSTALLED_APPS = (
    'django.contrib.auth',
    'django.contrib.contenttypes',
    'django.contrib.sessions',
    'django.contrib.sites',
    'django.contrib.admin',  # Required for helpdesk admin/maintenance
    'django.contrib.humanize',  # Required for elapsed time formatting
    'markdown_deux',  # Required for Knowledgebase item formatting
    'bootstrapform', # Required for nicer formatting of forms with the default templates
    'helpdesk',  # This is us!
)}

However ... it's not clear (until you get an error) that django.contrib.sites is a requirement, not an option.

Maybe just a 1-line addition to the docs?

The docs [here] (https://django-helpdesk.readthedocs.io/en/latest/configuration.html) say 

>  '6.  Log in to your Django admin screen, and go to the ‘Sites’ module. If the site example.com is listed, click it and update the details so they are relevant for your website.

Maybe this could be

>  '6. Log in to your Django admin screen, and go to the 'Sites' module (add 'django.contrib.sites' to your installed apps if you haven't allready). Create a new Site, or update example.com so that the details are relevant for your website.

D.

janhh:
Noticed the same. Added a command to my project to update Site object via terminal/bash to make it easier to do scripted deployment. When "Site.domain" wasn't probably configured, the URL in email pointed to wrong domain. I also put "SITE_ID = 1" into settings.py. Would be nice if this was pointed out in the documentation.

mrkiwi-nz:
@janhh : I've created PR #401 - would this solve the issue for you?

PR #401, opened 2016-06-23
Update README.rst

mrkiwi-nz:
Added a line item to alert users that 'django.contrib.sites' is required, and having at least 1 site is required.


mrkiwi-nz:
If merged, this solves #400

PR #407, opened 2016-06-28
Fixed remote code execution through unpickling untrusted code.

mhannig:
Hi!

I noticed, that you used pickle for (de-)serializing a saved query.
However, I could alter the payload in the save form and was able to insert base64 encoded pickle code, and thus be able to execute arbitrary code on the remote machine.

This patch fixes this problem.

Kind regards,

Matthias Hannig


rossp:
Many thanks for identifying and submitting this.

PR #409, opened 2016-06-29
Add an enabled field to queue - hide a queue without deleting it

mrkiwi-nz:
If you decide a queue is not required or not suitable you have two choices;
- delete the queue, deleting all tickets on that queue
- edit every ticket, moving them to another queue

This change adds a feature where you can disable a queue. A disabled
queue wont show in the logged in or public create ticket form, but
can still be displayed when viewing a ticket from that disabled queue.


mrkiwi-nz:
Ignore this PR - new PR to come soon.

flinz:
As mentioned [here](https://github.com/django-helpdesk/django-helpdesk/pull/409#issuecomment-229264387) I will close this PR since the author will open a new one.

PR #410, opened 2016-06-29
Fix savequery

mrkiwi-nz:
My browser sends 'on' when a checkbox is ticked, django ORM only
recognises '1', 'true' or 't' as valid 'True' responses.
This throws an error
ValidationError at /helpdesk/save_query/
[u"'on' value must be either True or False."]

This could be fixed with " value='1'" in the template, but testing that is harder
My fix is to add a check in the view. 2 more lines, but easier to unittest.
Core devs need to make a call as to which solution is best.

D

### DIFF
--- a/helpdesk/models.py
+++ b/helpdesk/models.py
@@ -562,6 +562,14 @@ class Ticket(models.Model):
 
         super(Ticket, self).save(*args, **kwargs)
 
+    @classmethod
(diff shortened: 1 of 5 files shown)


mrkiwi-nz:
After a couple of false starts, this is ready to merge :)
Not sure how this hasn't surfaced before, but full test coverage now.

PR #411, opened 2016-06-30
fix #408 : Bug in test_per_queue_staff_permission.py

### LINKED ISSUE #408: test_per_queue_staff_permissions uses an int as a password
In test_per_queue_staff_permissions.py ...
Line 13 : IDENTIFIERS = (1, 2)
Line 34 : for identifier in self.IDENTIFIERS:
Line 44 : user.set_password(identifier)

Line 44 sets the test user's password to the int 1 (or 2). This is fine for the django auth backend, however the ldap backend of Peter Sagerson's excellent django-auth-ldap (https://pythonhosted.org/django-auth-ldap/) throws an error when trying to len(password):
TypeError: object of type 'int' has no len()

A password should always be a string, so maybe we should to this:
Line 44 : user.set_password(str(identifier))

mrkiwi-nz:
Using the django auth backend allows integers to be passed as a password
Using Peter Sagerson's ldap auth backend there is an error thrown because
some code tries to do a len() on the password.
You could argue that the ldap auth backend should str(password), but
you could also argue that passing an int as a password is bad practice

This PR ensures that a string is sent to the auth module.

### DIFF
--- a/helpdesk/tests/test_per_queue_staff_permission.py
+++ b/helpdesk/tests/test_per_queue_staff_permission.py
@@ -41,7 +41,7 @@ class PerQueueStaffMembershipTestCase(TestCase):
                 username='User_%d' % identifier,
                 is_staff=True,
             )
-            user.set_password(identifier)
+            user.set_password(str(identifier))
             user.save()
 
             # The prefix 'helpdesk.' must be trimmed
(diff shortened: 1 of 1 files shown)


mrkiwi-nz:
Hi Ross, This one is ready to merge.

Issue #376, opened 2016-01-11
Get_email broken under Python 3.5


BenHagan:
Let me start by saying than you for keeping this going.  Django is a great platform and needs a good helpdesk app.

Now, when trying to run get_email with Python 3.5.1 and Django 1.9.1 I get errors:

`/helpdesk/management/commands/get_email.py", line 86
    print "Processing: %s" % q
                         ^
SyntaxError: Missing parentheses in call to 'print'`

This makes me think that this code was overlooked during a Python 3.5 port.  I fixed a few print statements in get_email.py but then hit this wall:

`/helpdesk/management/commands/get_email.py", line 102, in process_queue
    socket.socket = socket._socketobject
AttributeError: module 'socket' has no attribute '_socketobject'
`

Just to be sure I went and did another installation under 2.7 and everything works fine.

I'd like to take a stab at fixing this (might need some guidance).  Thoughts?

BenHagan:
I keep working my way through these, but it looks like the whole file is going to need refactoring for python 3.  If that becomes the case, does it become beneficial to create a separate 3.5 branch?

rossp:
@BenHagan thanks for spending some time on this. It should be quite possible to get this working on both 2.7 and 3.5 without having a separate branch - see recent work on #290 and #356 for some examples of this. 

It looks like you're coming across an issue where we've used undocumented and/or internal objects within `socket.socket` which will probably require a bit of troubleshooting to a) identify why it was done like that, and b) make it work on multiple systems.

It's probably easier to backport 3.5-OK code to 2.7 than the alternative - so see if you can get this working and we can go ahead and make it work on older versions too.

flinz:
This functionality should be generally covered by tests, which it obviously is not right now (python 3.5 passes all tests). Adding this to the Testing project.

gwasser:
I have a new get_email test in #420. I'm not sure about the socket issue, not clear to me what it's doing.

reduxionist:
After looking at the socket._socketobject assignment's potential impact in the commit where it first appeared and in a current stable release, I'm led to believe that it has never actually served any purpose at all.

It could be an artifact left over from a refactoring, or perhaps it was part of the pattern used to add the SOCKS support which motivated the commit it originates from, and that part just happens to be unnecessary in our case (it might be helpful if switching back between SOCKS and vanilla sockets, but we don't).

[...]

PR #420, opened 2016-09-15
get_email local mail directory, logging support, Python3 via six

gwasser:
Can now set a local directory to poll for email, not just remote POP3/IMAP (more options!). My IT department sets up mail to get dumped in a folder on the server, so just have to parse it. Maybe this is useful to others. Ticketing systems like JIRA check local directories too, so seems common use case. The directory can be set by the user in the Queue admin page.

Additionally, I replaced the print() statements with logging statements. Logging will be printed to console unless --quiet option is enabled. By default, logging to file is turned off but user can set the log file directory in the Queue admin page now.

Used python-six to add python3 compatibility along side existing python2. This may go toward addressing #376 but I haven't tested the POP3/IMAP support yet. I can say that it *does* successfully process mail from a local directory on Python 3.5 though, so the logic is good other than the potential socket issue.


helpdesk/management/commands/get_email.py
```diff
@@ -5,18 +5,21 @@
...
 """
-from __future__ import print_function
 
 import email
 import imaplib
 import mimetypes
 import poplib
 import re
 import socket
+import six
```
flinz: Can we use [django.utils.six](https://docs.djangoproject.com/en/1.7/topics/python3/#module-django.utils.six) instead? This would prevent the additional dependency.

helpdesk/management/commands/get_email.py
```diff
@@ -39,20 +43,34 @@
...
                 '--quiet', '-q',
                 default=False,
                 action='store_true',
                 help='Hide details about each queue/message as they are processed'),
             )
 
-    help = 'Process Jutda Helpdesk queues and process e-mails via POP3/IMAP as required, feeding them into the helpdesk.'
+    help = 'Process django-helpdesk queues and process e-mails via POP3/IMAP or from a local mailbox directory as required, feeding them into the helpdesk.'
+
+    def add_arguments(self, parser):
```
flinz: So, I assume correctly that this would have failed in Django > 1.7, since we did not provide the ``add_argument``?
Maybe we could add some tests that cover this?

helpdesk/management/commands/get_email.py
```diff
@@ -100,7 +137,11 @@ def process_queue(q, quiet=False):
         socks.set_default_proxy(proxy_type=proxy_type, addr=q.socks_proxy_host, port=q.socks_proxy_port)
         socket.socket = socks.socksocket
     else:
-        socket.socket = socket._socketobject
+        if six.PY2:
```
flinz: Is setting the socket to _socketobject necessary at all? I can't find documentation about this anywhere, and this might be some legacy code..
Are you aware of how the socket.socket object is used in the remainder?

flinz: Also, see if we can use the django bundled six version to do this fork.

helpdesk/management/commands/get_email.py
```diff
@@ -287,16 +381,23 @@ def ticket_from_message(message, queue, quiet):
...
-        print((" [%s-%s] %s" % (t.queue.slug, t.id, t.title,)).encode('ascii', 'replace'))
+    if six.PY2:
+        logger.info(("[%s-%s] %s" % (t.queue.slug, t.id, t.title,)).encode('ascii', 'replace'))
+    elif six.PY3:
+        logger.info("[%s-%s] %s" % (t.queue.slug, t.id, t.title,))
 
     for file in files:
         if file['content']:
-            filename = file['filename'].encode('ascii', 'replace').replace(' ', '_')
+            if six.PY2:
```
flinz: Same here: Can we use django.utils.six instead?

helpdesk/management/commands/get_email.py
```diff
@@ -287,16 +381,23 @@ def ticket_from_message(message, queue, quiet):
...
         f.new_status = Ticket.REOPENED_STATUS
         f.title = _('Ticket Re-Opened by E-Mail Received from %(sender_email)s' % {'sender_email': sender_email})
-    
+
     f.save()
+    logger.debug("Created new FollowUp for Ticket")
 
-    if not quiet:
-        print((" [%s-%s] %s" % (t.queue.slug, t.id, t.title,)).encode('ascii', 'replace'))
+    if six.PY2:
```
flinz: Same here: Can we use django.utils.six instead?

gwasser:
@flinz do you have any thoughts on this? I made a new test for get_email which only fails on Python 3.4 with Django 1.7. I can't think of what's wrong or how to test it for it.

gwasser:
To be honest, since 1.7 is no longer supported and 1.8 is LTS, I'm tempted to just drop Django 1.7 from support. If we do that, perhaps that can be part of the 0.2 branch since it may break some installations (though anyone using Django 1.7 still should really upgrade!).

reduxionist:
I'm also in favor of dropping support for 1.7 in our 0.2 branch

flinz:
@gwasser I will have a look at this, also regarding the mocking of the tests referred to in #376. 
Regarding the dropping of django 1.7 support, I opened a general ticket here, #433 but would still aim to make this PR also compatible with 1.7. Will have a look at it, too.

flinz:
@reduxionist Even if this is now merged, I would strongly recommend we push to increase the amount of tests done on get_email - many parts, like the infamous [_socketobject lines](https://github.com/django-helpdesk/django-helpdesk/pull/420/files#diff-38279aea05eea2eaf213d3c3ac29f6e7R145) seem a little dubious. I'll open an issue.

gwasser:
Completely agree, for anyone looking the discussion I think will continue in #444

PR #428, opened 2016-10-18
Fix form translations

jflesch:
Hello,

I'm working with Python 3.4, so I use the Git version of django-helpdesk. And I noticed an annoying problem with the french translations (see attached screenshot). Here is the fix (and a small cleanup commit ;).

helpdesk-fr-bug


reduxionist:
Merci beaucoup pour le fix @jflesch !

Issue #434, opened 2016-10-21
Coding style and code policies


alexbarcelo:
I have been bugged by some "stylistically strange" parts on the code. Personally, I like to follow, for the most part, pep8. The line length is a different topic (see below), but without considering that, a call to `pep8 --ignore=E501 ./django-helpdesk` yields 872 lines, of a total of 36 different complains.

Regarding the long lines, a quick look at them gives:
- 1974 lines longer than 79 characters
- 982 lines longer than 99 characters
- 414 lines longer than 119 characters
- 220 lines longer than 139 characters

Long lines are counted with multiplicity.

There are a lot of long lines in `migrations` and also `south_migrations`. Maybe in the branch 0.2 the migrations can be squashed and cleaned up... ok, that is a different topic. But note that without both migration folders, there are 831 non-length issues and 449 long lines (>79).

My question would be, it is a good idea to bring a major clean-up? Do you feel that it is worth to follow pep8 (from now on)? Or you feel that it is better to go beyond that and use `pylint`?

It is a tedious task, but it is quite straightforward. I don't mind to prepare a PR for that, but I am not sure if you agree on that topic, or maybe prefer to follow a subset of the `pep8` may be followed, or go all out with `pylint`.

reduxionist:
I am definitely in favor of normalizing the code style, and I'll be happy with either PEP8 or pylint as long as line-length isn't limited to 79 chars.

gwasser:
PEP8 sounds fine to me. I would agree that >79 chars isn't a big deal; can do it when possible.

flinz:
I opt for pep8. 
What is the opinion of adding a pep8 call to travis? I.e. PRs will fail the build if they are not pep8 compliant. I used this in other projects already and think its a good idea, which will keep coding guidelines enforced.

alexbarcelo:
I think that the Travis build would be a good idea, I didn't know that option.

Right now Travis would fail (there are still `pep8` issues) but as an automatic checker it would be quite nice (IMHO).

alexbarcelo:
#438 has been merged and now the project is PEP8-compliant and travis checks it in `before_script`.

Nice! Closing this. Thanks all for the feedback!

PR #435, opened 2016-10-21
General clean-up and opinionated coding style

alexbarcelo:
As I recently commented in #434, here the first commits of the general clean up on django-helpdesk code base.

Along the path I have fixed some dead code and some ancient Django stuff. So it is *not* a pure coding style PR, as there is something more. However, no real changes should have been made (unless I messed up).

There is still work to do, and pylint complains about a lot of stuff. But as a starting point...

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,0 +1,378 @@
+[MASTER]
+
+# Specify a configuration file.
+#rcfile=
+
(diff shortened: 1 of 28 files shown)


reduxionist:
Thanks for giving us some style @alexbarcelo !! 🎩

PR #438, opened 2016-10-23
More Code Style --PEP8-centric

alexbarcelo:
Those changes have been made with the `autopep8` util, plus some fine tuning. Long line lengths have been ignored. Some paths have been deliberately ignored at the moment. But right now, almost all the files are pep8-compliant.

### DIFF
--- a/helpdesk/akismet.py
+++ b/helpdesk/akismet.py
@@ -24,14 +24,14 @@ You should pass in the keyword argument 'agent' to the name of your program,
 when you create an Akismet instance. This sets the ``user-agent`` to a useful
 value.
 
-The default is : ::
+The default is::
 
     Python Interface by Fuzzyman | akismet.py/0.2.0
 
 Whatever you pass in, will replace the *Python Interface by Fuzzyman* part.
 **0.2.0** will change with the version of this interface.
 
-Usage example:
+Usage example::
 
     from akismet import Akismet
 
--- a/helpdesk/akismet.py
+++ b/helpdesk/akismet.py
@@ -15,7 +15,7 @@
 
 """
 A python interface to the `Akismet <http://akismet.com>`_ API.
-This is a web service for blocking SPAM comments to blogs - or other online 
+This is a web service for blocking SPAM comments to blogs - or other online
 services.
 
 You will need a Wordpress API key, from `wordpress.com <http://wordpress.com>`_.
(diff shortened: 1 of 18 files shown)


flinz:
Could we aim at making a single PR that lets a ``pep8 --ignore=E501 helpdesk`` pass cleanly?
Then you could include in this PR the pep8 call in the .travis.yml in either the ``script`` or the ``before_script`` block (see here for an example http://cramer.io/2012/05/03/using-travis-ci)

alexbarcelo:
Added that. Also included `--exclude=migrations,south_migrations` because maybe it is not worth cleanint them up.

flinz:
Looks good to me!

There is one issue we should discuss: Adding the pep8 to travis might break some passing PR's that are ready to be merged.. 
Perhaps we could aim to add pep8 formatting with CI checks to 0.2.0 as well?
We will still have to reformat some PR's designated for that branch.

gwasser:
I tend to think we should do minimal bug fixing only and branch the 0.1 release ASAP, then add breaking changes (such as maybe this pep8 change) to the 0.2 branch. 0.2 PRs might need to get resubmitted, but that's ok with me. Maybe set a date for the branching and post it somewhere (README?)? Anyone that wants a PR pre-pep8 on the 0.1 branch needs to do it by XXX date, otherwise you may need to resubmit. Just my offhand thoughts.

alexbarcelo:
I mostly agree with @gwasser, however I am not sure if you are thinking on applying the pep8 in both branches or only in the 0.2 branch.

Both options seems fine to me.

reduxionist:
I agree too, and since practically all of the merge-able PRs are from people in this thread, I think it'll be safe to issue the 0.1 release after a simple 24-hour grace period, as of this notice, for any objections or breaking issues from project watchers.

helpdesk/akismet.py
```diff
@@ -31,10 +31,10 @@
 Whatever you pass in, will replace the *Python Interface by Fuzzyman* part.
 **0.2.0** will change with the version of this interface.
 
-Usage example::
```
reduxionist: The second colon is needed for rst literal-block syntax and should be restored.

While we're at it, line 27 above should also end with just the double colons, the single colon and surrounding spaces that precede them should be removed so the line ends with just "is::".

reduxionist:
Thanks again for the contribution @alexbarcelo !

Issue #198, opened 2014-01-29
Replace API with django-rest-framework or similar


rossp:

rossp:
On further thought - API should be removed entirely. Not our job to provide IMO - people can address this with multiple external means (tastyppie, django-rest-framework, SQL, custom Python code). I'm going to add deprecation warnings then remove it entirely next month.

TauPan:
I can recommend https://github.com/izimobil/django-rest-framework-datatables if you want to migrate your datables views to django-rest-framework as well. I'm in the process of converting legacy datatables views to rest framework and the resulting code  is usually much simpler.

gwasser:
Wasn't aware of that django app, thanks for the recommend! Sounds like something to be considered in the future when get to tackling the API.

Mte90:
Any updates on this?
If can be helpful a rest documentation from another OSS helpdesk https://www.opensupports.com/apidoc/

gwasser:
@Mte90 I personally have not had time to investigate or implement this, I welcome volunteers to work on it and submit a PR though!

Mte90:
So it is moving on :-D

To me is enough to add the user rest endpoints and I am fine :-D

gwasser:
Yes an API based on django-rest-framework should be available in the next release `0.3.4` which should come soon once we've cleaned up some testing and what not.

gwasser:
Will close this ticket as part of getting ready for the next release. Thanks!

PR #441, opened 2016-10-29
Remove all the deprecated API related files and references

alexbarcelo:
As the docs already stated:

> The django-helpdesk API is deprecated, and no longer maintained. See https://github.com/django-helpdesk/django-helpdesk/issues/198 for more details.
> The API will be removed in January 2016

Now that we have a new-and-shiny 0.2, I think that we can safely drop all usages.

(I feel bad for removing so much code in a single commit)


reduxionist:
Awesome! You should feel awesome for removing so much code in a single commit! 😈

alexbarcelo:
Well, that's a way to look at it, :laughing: 

Less than 20 minutes from the PR until LGTM+Merge. Ok, that was insane. Even for a trivial PR. Loving it :wink:

PR #443, opened 2016-10-30
Clean and avoid "repeating yourself" on forms.py

alexbarcelo:
The main commit (the second one) is a-hell-of-a-diff because I took all the repeated code in `TicketForm` and `PublicTicketForm` and put it in an `AbstractTicketForm`.

I found some inconsistencies while translating django-helpdesk, and then understood that they were due to some repeated code and repeated fields between those two forms.

~~However, I have yet to address the failing build on Python 3.4. I don't recall doing any "dangerous stuff" there, but Travis doesn't lie.~~ edit: Nevermind, I must have my fork master somewhat dirty, and it failed. But it is unrelated to this PR because here it seems to be OK regarding tests.

### DIFF
--- a/helpdesk/forms.py
+++ b/helpdesk/forms.py
@@ -133,7 +133,11 @@ class EditFollowUpForm(forms.ModelForm):
         self.fields["ticket"].queryset = Ticket.objects.filter(status__in=(Ticket.OPEN_STATUS, Ticket.REOPENED_STATUS))
 
 
-class TicketForm(CustomFieldMixin, forms.Form):
+class AbstractTicketForm(CustomFieldMixin, forms.Form):
+    """
+    Contain all the common code and fields between "TicketForm" and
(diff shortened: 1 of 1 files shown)


gwasser:
This is great! Thanks for the work @alexbarcelo.

alexbarcelo:
@reduxionist thx for the comments. Regarding variable naming, I almost changed them, but decided to leave the original ones in some places. In this case, it was a better idea to avoid single-letter variables everywhere. 

The last commit uses a dict to simplify a little bit the code on the case pattern. There is room for improvement, but at the time it seemed a straightforward reasonable way to improve readability.

reduxionist:
The module is much cleaner now, and a refactoring that increases test coverage without even having to add new tests is a big win! 😃  Thanks very much @alexbarcelo!

Issue #444, opened 2016-10-31
Proper testing for get_email.py functionality


flinz:
At the moment large chunks of [get_email.py](https://github.com/django-helpdesk/django-helpdesk/blob/master/helpdesk/management/commands/get_email.py) remain largely untested.

Problem areas are:
- [imap and pop3 functionality](https://codecov.io/gh/django-helpdesk/django-helpdesk/src/d54017f69f75d933c357a9f49b836de748972082/helpdesk/management/commands/get_email.py#L153)
- [SOCKS proxy support](https://codecov.io/gh/django-helpdesk/django-helpdesk/src/d54017f69f75d933c357a9f49b836de748972082/helpdesk/management/commands/get_email.py#L125)

Particularly, to which degree we should test or mock imap and pop3 with socks sockets escapes me for now. I will look at extending this incrementally, but if anybody wants to help out, feel welcome.

Edit: sorry for formatting errors

flinz:
See https://github.com/flinz/django-helpdesk/tree/tests/get-email for my active branch.

flinz:
@gwasser something I just noticed during writing a little on tests: 
in [this line and before](https://github.com/django-helpdesk/django-helpdesk/blob/master/helpdesk/management/commands/get_email.py#L101), even if you set logging_type = 'none', you still add a filehandler to the default directory ``/var/log/helpdesk/``.
Thus, I think as it is it is impossible to disable logging.

I think a possible fix would be this (at least it makes it runnable in tests for me)
```python
logger = logging.getLogger('django.helpdesk.queue.' + q.slug)

if not q.logging_type or q.logging_type == 'none':
    logging.disable(logging.CRITICAL)  # disable all messages
else:
    if q.logging_type == 'info':
        logger.setLevel(logging.INFO)
    elif q.logging_type == 'warn':
        logger.setLevel(logging.WARN)
    elif q.logging_type == 'error':
        logger.setLevel(logging.ERROR)
    elif q.logging_type == 'crit':
        logger.setLevel(logging.CRITICAL)
    elif q.logging_type == 'debug':
        logger.setLevel(logging.DEBUG)
    
    logdir = q.logging_dir or '/var/log/helpdesk/'
    handler = logging.FileHandler(logdir + q.slug + '_get_email.log')
    logger.addHandler(handler)

if quiet:
    logger.propagate = False  # do not propagate ...
```
If this is correct, mind opening a quick PR to fix it?

[...]

PR #446, opened 2016-11-01
patch: implicit closing of open files;

flinz:
``get_email`` opened files were never closed.

### DIFF
--- a/helpdesk/management/commands/get_email.py
+++ b/helpdesk/management/commands/get_email.py
@@ -234,17 +234,17 @@ def process_queue(q, logger):
         logger.info("Found %s messages in local mailbox directory" % str(len(mail)))
         for m in mail:
             logger.info("Processing message %s" % str(m))
-            f = open(m, 'r')
-            ticket = ticket_from_message(message=f.read(), queue=q, logger=logger)
-            if ticket:
-                logger.info("Successfully processed message %s, ticket/comment created." % str(m))
-                try:
-                    # unlink(m) #delete message file if ticket was successful
-                    logger.info("Successfully deleted message %s." % str(m))
-                except:
-                    logger.error("Unable to delete message %s." % str(m))
-            else:
-                logger.warn("Message %s was not successfully processed, and will be left in local directory" % str(m))
+            with open(m, 'r') as f:
+                ticket = ticket_from_message(message=f.read(), queue=q, logger=logger)
+                if ticket:
+                    logger.info("Successfully processed message %s, ticket/comment created." % str(m))
+                    try:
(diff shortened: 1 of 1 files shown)


helpdesk/management/commands/get_email.py
```diff
@@ -234,17 +234,17 @@ def process_queue(q, logger):
...
-                except:
-                    logger.error("Unable to delete message %s." % str(m))
-            else:
-                logger.warn("Message %s was not successfully processed, and will be left in local directory" % str(m))
+            with open(m, 'r') as f:
+                ticket = ticket_from_message(message=f.read(), queue=q, logger=logger)
+                if ticket:
+                    logger.info("Successfully processed message %s, ticket/comment created." % str(m))
+                    try:
+                        # unlink(m) #delete message file if ticket was successful
```
reduxionist: @gwasser So that this commented code doesn't get forgotten about, please could you tell us under what conditions it should be activated in future? That is, of course, IIRC that this line was introduced in a previously-merged PR of yours! :)

gwasser: To be honest, I think it was left over from manual testing. I didn't want to delete files in testing so I could re-run in it easily (this was before I learned about mock).

I believe the line should be uncommented so that mail files are removed as they are processed, and not processed again later. I can add this fix to a PR for #444 that I need to do anyway.

flinz: I personally would be very careful about deleting system files from django - on the other hand I also do not know to which degree this could be circumvented. Any opinions?

gwasser: I modeled this behavior after how JIRA takes in new tickets. New emails are dumped in a directory, helpdesk reads email to create a ticket, and then deletes it once confirmed successful to ensure that the same email doesn't generate duplicate tickets. I personally don't think there's a good way to avoid doing so. A compromise might be that emails could simply be renamed, rather than deleted, but that might not be desirable, it could fill up a drive with duplicate data. I would prefer simply deleting them; possibly documentation could be better improved to make it clear that files are permanently deleted and to be sure this is the behavior you want.

reduxionist: Yeah, that's how I see it. This behaviour is analgous to how we deal with POP3/IMAP mail, it's just that we don't have the extra protocol layer to pass deletions through since this mail is delivered directly as file-on-disk IIRC. There are past issues about duplicate ticket creation by emails that didn't get deleted, e.g. due to exceptions thrown during attachment processing so I think that loud and explicit documentation remains the best option. $0.02...

Issue #437, opened 2016-10-23
Things which may be dropped in new 0.2


alexbarcelo:
Here a quick list of stuff that I think might be considered for the 0.2. I am not advocating for removing (all of) them, but having a list may be a good idea. Tell me if I have left something out.
- [x] **Django 1.7** Ok, this seems almost done with #436, already commented in #433 
- [x] **API** According to itself, it is deprecated and would be removed on January. #198 --edit: merged in #441 
- [x] **south_migrations** and support for south in general. I am not sure on that, I have myself never used south_migrations, but given that Django 1.8 has native migrations... Is it still being used? Is it worth keeping? Sorry for my ignorance.
- [x] ~~**squash migrations** Not exactly "drop". There are not a lot of migrations, but there is some old stuff (and very bad style). A unique squashed migration may be worth doing? This would potentially mean that updates from 0.1 into 0.2 should be done in two steps, and this may be a deal breaker.~~ No. Leave the migrations for upgrading users.

rossp:
south_migrations shouldn't be necessary if Django 1.7 is being dropped. When Django 1.7 migrations were released they needed the 'migrations' folder; the final version of South for <1.7 included code to look in south_migrations first. Having both in django-helpdesk allowed for both Django 1.6 and 1.7 compatibility.

Unsure how practical it is to squash the migrations, but the worst of them was in the South days so should be much tidier once you drop those.

gwasser:
Do older releases (1.6 and below) need to use south_migrations to get up to 1.7 BEFORE they can use django migrations to get up to 1.10? I'm thinking of the full upgrade path here. If someone on say 1.6 needs it to upgrade up to a supported django, I would suggest that we deprecate the south_migrations in 0.2, stick a big warning in the README to upgrade/migrate ASAP, and remove in 0.3 release (as a tentative suggested date, maybe we can drop the south_migrations when django 1.11 LTS is out in April 2017).

rossp:
The answer to that probably depends on whether the `south_migrations` are in check with the `migrations` up to the point where a Django release is no longer supported - ie could somebody on 1.6 use `south_migrations` to get to a schema state that is identical to somebody who is on 1.9 using `migrations`?

Unfortunately I do not know the answer to that, and controlling Django version deprecation and enforcing concurrent `migrations` and `south_migrations` through the transition process wasn't properly done by me. Sorry :(

[...]

PR #447, opened 2016-11-02
Drop south_migrations

alexbarcelo:
This has been discussed in #437 

The PR proposes the deletion of `south_migrations` folder, plus some leftover references. I could not find any documentation regarding those migrations, so it seems quite a safe PR (given that support for 1.7 has been dropped).


reduxionist:
Thanks for the continued contributions!

PR #451, opened 2016-11-11
convert reverse-URL lookups to helpdesk namespace

reduxionist:
in views.staff, this fixes:
-run_report's no-tickets-found redirect to report_index
-del_attachment's redirect to parent ticket view after deletion

Although there's no tests for this pair of one character changes, I am applying the new testing requirements to myself on a larger PR which would otherwise be ready to address the open attachment issues. ;)


reduxionist:
The tests are for a different PR, a PR for the various attachment-filename issues. :smile:

Issue #450, opened 2016-11-11
ticket list view - order by creation time bug


reduxionist:
I haven't even begun to look at the JS behind our beautiful new UI but we seem to be using lexicographical, rather than chronological, sorting for this column.

gwasser:
Hmm, so it makes use of DataTables (https://datatables.net/reference/option/order) and as far as I can tell, there's no option to sort by non-lexicographical order (I'm guessing our numbered tickets is a bit of a corner case?).

Rather than mess with the library, I wonder if we could have some magic to insert leading zeros into the displayed numbers so they're compatible with lexicographical order? Or would that be too confusing?

reduxionist:
when sorting by ctime, an example would be that 20 mins was ranked < 30 seconds.

I didn't even think to check effect of our ticket naming scheme!

You may be pointing to the solution though - disable time-ago humanization if it's coming from server-side...

gwasser:
Oh I apologize, misread something and thought you were looking at a different column. My bad!

I think my comment still applies though. Should it by default list as "0 mins 30 seconds" so it gets sorted before 20 mins? A bit more wordy, but maybe ok.

reduxionist:
While normalizing with zeroes like that works in the examples used so far, applied to a wider range of creation times (seconds to months is 6 units) I think it gets so wordy that it's not worth using humanization any more. It might look cleaner to just output a standard javascript date type since DataTables can sort those correctly out-of-the-box.

Alternatively, we could use the absolute datetime value (which is stored in the title attribute of the span enclosing the relative value), as an orthogonal data source to sort by as per [this DataTables example](https://datatables.net/manual/data/orthogonal-data#Predefined-values). We'd probably need to adjust the absolute datetime format to be numeric-sort-friendly (e.g. epoch time), or use one of the [DataTables date sorting plugins](https://datatables.net/plug-ins/sorting/#Plug-ins), too.

As a user, sometimes I appreciate relative dates and sometimes I want to know exactly when something occurred and find the coarse-grained months/years -ago format frustrating, so any approach is OK with me really as long it sorts correctly in the end! 😈

gwasser:
Turns out the solution seemed much simpler than I was worried about, thanks for doing some preliminary research! See PR #455 .

reduxionist:
Fixed by #455 (sub-second precision not required IMO!)

Really loving the new front-end, well done @gwasser!!

PR #455, opened 2016-11-21
Use HTML5 table ordering based on Unix epoch

gwasser:
So turns out HTML5 allows you to add additional data to sort by independently from the displayed data: https://datatables.net/manual/data/orthogonal-data#HTML-5

Also Django has a filter for outputting the Unix time: https://docs.djangoproject.com/en/1.10/ref/templates/builtins/#std:templatefilter-date (see option -U)

So this seems to work great on my end. My only concern is it only prints the number of seconds since epoch, not microseconds, so it may not be totally exactly in chronological order if multiple tickets are submitted within one second (but at that level does it matter?)

This addresses #450 .


reduxionist:
Don't you love it when a fix turns out to be this clean? Thanks very much @gwasser! 🍰

PR #459, opened 2016-12-02
Expand testing of get_email

reduxionist:
Using @flinz's parameterized test case, this adds tests for mocked IMAP and POP3 retrieval.
In order not to break the build, the IMAP case exempts a known error from failing the test because the fix for it is part of a get_email refactoring in the PR following this one, and the special-casing gets removed at that point.
There's plenty more testing that could be done, but this is a start that I wanted to get out there as a basis for further discussion on how we should handle email processing.

### DIFF
--- a/helpdesk/tests/test_get_email.py
+++ b/helpdesk/tests/test_get_email.py
@@ -1,12 +1,10 @@
 from helpdesk.models import Queue, Ticket
-from helpdesk.management.commands.get_email import process_email
 from django.test import TestCase
-from django.core import mail
 from django.core.management import call_command
-from django.test.client import Client
 from django.utils import six
-from django.core.urlresolvers import reverse
 from django.shortcuts import get_object_or_404
+import itertools
+import sys
 
 try:  # python 3
     from urllib.parse import urlparse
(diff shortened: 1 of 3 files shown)


helpdesk/tests/test_get_email.py
```diff
@@ -20,35 +20,114 @@
...
-            mocked_listdir.return_value = ['filename1', 'filename2']
+        test_mail_len = len(test_email)
 
-            call_command('get_email')
+        if self.socks:
+            from socks import ProxyConnectionError
+            with self.assertRaisesRegexp(ProxyConnectionError, '%s:%s' % (unrouted_socks_server, unrouted_port)):
+                call_command('get_email')
+
+            # with mock.patch('socket.socket') as mocked_socket,\
```
flinz: Are the commented lines for development purposes? Could you remove them?

helpdesk/tests/test_get_email.py
```diff
@@ -20,35 +20,114 @@
     # Python < 3.3
     import mock
 
+unrouted_socks_server = "127.0.0.1"
```
flinz: I was unsure about this when I started developing this..
Can we be sure the test does not accidentally go awry if somebody has a socks server or email server running on localhost:12345?

However, on travis this will for sure be not the case, so I think as a first approach this is fine.

helpdesk/tests/test_get_email.py
flinz: Are the commented lines for development purposes? Could you remove them?

helpdesk/tests/test_get_email.py
reduxionist: For that sense of unrouted I think the safest address would be an invalid one, e.g. 0.1.1.1. You're guaranteed no traffic is going anywhere that way...

helpdesk/tests/test_get_email.py
reduxionist: Good feedback, I will definitely comment it more, submitted the PR too immediately after writing the code to think of it myself, so thanks! 😄

helpdesk/tests/test_get_email.py
flinz: Ok, ideally this would be the way to go. Be advised, I ran into some weird deviation from the expected testing behavior when using non-existing ports or nonsensical ips -- some errors that are raised differ from the ones I tried to catch

helpdesk/tests/test_get_email.py
flinz: *Gulp* :) See, i already didn't even remember writing this, haha.
Thanks again for picking up the branch!

reduxionist:
Although I waited a few days so that it wasn't too fresh in my head, please don't be shy to tell me of any areas that need more commenting because I know how it's always just too obvious to the author at the time but that I'll be glad for them in future! 😉

PR #462, opened 2016-12-08
PY3 support and msgNum error solved in process_queue method

pawelmarkowski:


pawelmarkowski:
In my case type of msg variable is <class 'bytes'>. Python 3.4 / django 1.10.4  (imap connection)

reduxionist:
Thanks for PR, so as not to duplicate any efforts, I would check against #463 which refactors this code path and establishes some test coverage over. I have also included your commit fixing the blank title bug on a rebased version of PR393 currently queued up on my personal fork. 😄

pawelmarkowski:
Great, so I am waiting for your upgrade.

In my opinion there is some kind of bug. If queue.slug does not match matchobj = re.match(r".*\[" + queue.slug + "-(?P<id>\d+)\]", subject)  get_email will create new ticket even if received message was an update for other queue.

reduxionist:
As this PR was also effectively merged by #463, I'm closing it too. Thanks for all the contributions you've made @pawelmarkowski! 🏅  Please keep 'em coming! 😀

Issue #456, opened 2016-11-28
ticket_from_message(message, queue, logger)


rando305:
ticket_from_message now requires three arguments, but the calling instances in get_email.py only pass two.  (They aren't passing the logger).

reduxionist:
Sorry you ran into this, thanks for reporting it. A patch is going through CI now and should be merged shortly.

rando305:
In line 217 of get_email.py, msgNum is referenced prior to assignment.  I expect it is supposed to be just 'num'?

Two lines below is a similar problem.

I obviously don't know the correct way to alert this in github - I'd be appreciative of guidance on how to best provide feedback.

reduxionist:
In this case, this way of alerting is fine; technically, each different bug should be filed under a new issue, but since these are all tiny ones in the same module (probably relating to the same commit too), I don't mind collecting them here.

In any case, I believe that this one is fixed in [a dev branch on my fork](https://github.com/reduxionist/django-helpdesk/tree/issue/multiple/attachments), and since the first one you reported above was also already fixed there,  I was hoping that perhaps you'd be willing to give that branch a try directly? If not I can keep pulling out patches and hot-fixing, I just prefer to keep the main tree clean when I have the option to merge a rebased version of a feature-branch later... Thanks for the reports in either case! 😄

reduxionist:
Uhhh, unless you're on Django 1.8 and Python 3. Looks like I missed a string encoding update, but Travis just caught it for us so don't try my branch if those versions are your poison... 😉

rando305:
No, I'm on 1.9 and Python 3.  

I've never pip ed from a branch before.

Is it just pip install git+https://github.com/reduxionist/django-helpdesk/  ?

reduxionist:
Just to let you know, both of these patches have been merged into master now so you should be safe to install from the official source next time it comes up. Thanks again for the bug reports!

reduxionist:
The results from Travis have just shown me I've broken my local test environment (pyenv + tox not going so well for me today), and it looks like issue is actually worse now because I got the difference in using the std. lib.'s email.message methods backwards and am actually throwing exceptions under Py3 despite what I believed my local testrunner was showing me. I'm going to have to fix my dev env before I can make any sensible approach, so now that at least you know where my code went bad please feel free to apply a fix if you're so inclined. Otherwise, I'll let you know once I'm back on it myself... 🙈

reduxionist:
Whoops, wrong issue!

Issue #439, opened 2016-10-24
Error on attachement files


Nicoloren:
It seems a user can't attach a file with a reply : 

`Internal Server Error: /helpdesk/tickets/2/update/

TypeError at /helpdesk/tickets/2/update/
cannot use a string pattern on a bytes-like object

Request Method: POST
Request URL: http://xxxxxxxxx.xx/helpdesk/tickets/2/update/
Django Version: 1.9.2
Python Executable: /usr/local/bin/python3.5
Python Version: 3.5.2
Python Path: ['/app', '/usr/local/bin', '/usr/local/lib/python35.zip', '/usr/local/lib/python3.5', '/usr/local/lib/python3.5/plat-linux', '/usr/local/lib/python3.5/lib-dynload', '/usr/local/lib/python3.5/site-packages', '/usr/local/lib/python3.5/site-packages/newrelic-2.70.0.51']
Server time: lun, 24 Oct 2016 15:51:13 +0200
Traceback:  

File "/usr/local/lib/python3.5/site-packages/django/core/handlers/base.py" in get_response
1.                     response = self.process_exception_by_middleware(e, request)

File "/usr/local/lib/python3.5/site-packages/django/core/handlers/base.py" in get_response
1.                     response = wrapped_callback(request, _callback_args, *_callback_kwargs)

File "/usr/local/lib/python3.5/site-packages/helpdesk/views/staff.py" in update_ticket
1.                 mime_type=mimetypes.guess_type(filename)[0] or 'application/octet-stream',

File "/usr/local/lib/python3.5/mimetypes.py" in guess_type
1.     return _db.guess_type(url, strict)

File "/usr/local/lib/python3.5/mimetypes.py" in guess_type
1.         scheme, url = urllib.parse.splittype(url)

File "/usr/local/lib/python3.5/urllib/parse.py" in splittype
1.     match = _typeprog.match(url)

Exception Type: TypeError at /helpdesk/tickets/2/update/
Exception Value: cannot use a string pattern on a bytes-like object
Request information:
GET: No GET data

POST:
due_date_year = '2016'
new_status = '2'
csrfmiddlewaretoken = '7w8xgnOiUeKJlYD2kCA3xGkF44sDoRXe'
owner = '0'
comment = 'Voici une photo de moi'
priority = '3'
title = "Demande d'étude de raccordement"
public = '1'
due_date_day = '20'
due_date_month = '11'

FILES:
[...]

Issue #413, opened 2016-07-04
ValueError: I/O operation on closed file 


mrkiwi-nz:
When attaching a file to a ticket the following error occurs
ValueError: I/O operation on closed file
`[04/Jul/2016 15:01:08] ERROR [django.request:255] Internal Server Error: /helpdesk/tickets/21/update/
Traceback (most recent call last):
  File "/home/e-smith/files/ibays/django/django_app/m10_repo/m10virtualenv/lib/python2.7/site-packages/django/middleware/exception.py", line 35, in __call__
    response = self.get_response(request)
  File "/home/e-smith/files/ibays/django/django_app/m10_repo/m10virtualenv/lib/python2.7/site-packages/django/core/handlers/base.py", line 277, in _legacy_get_response
    response = self._get_response(request)
  File "/home/e-smith/files/ibays/django/django_app/m10_repo/m10virtualenv/lib/python2.7/site-packages/django/core/handlers/base.py", line 194, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "/home/e-smith/files/ibays/django/django_app/m10_repo/m10virtualenv/lib/python2.7/site-packages/django/core/handlers/base.py", line 192, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/home/e-smith/files/ibays/django/django_app/m10_repo/m10virtualenv/lib/python2.7/site-packages/helpdesk/views/staff.py", line 553, in update_ticket
    files=files,
  File "/home/e-smith/files/ibays/django/django_app/m10_repo/m10virtualenv/lib/python2.7/site-packages/helpdesk/lib.py", line 143, in send_templated_mail
    msg.attach(filename=attachment[0], content=file_to_attach.read())
ValueError: I/O operation on closed file
`

rossp:
@mrkiwi-nz do you have any information about what triggered this? does the file exist (ie has it been saved to the filesystem), what file storage backend are you using, etc?

mrkiwi-nz:
The file is saved to the file system and the ticket status/attachment is all saved and ok, as proved by viewing the ticket in another window and being able to download the attachment. 
I think this bug triggers when the email system tries to pickup the file and attach to the email? I didn't have time to replicate on my localhost, attach the debugger and step through, but i will do later in the week - crazy workload right now. I am using an ldap backend for auth, but other than that it's a standard install (django 1.10a1, mysql). More debug later, D

[...]

Issue #343, opened 2015-10-16
Sending attachments with email to helpdesk results in: No plain-text email body available.


dominicfollett:
Hi there,

The resulting ticket ends up looking like this: 

![screenshot from 2015-10-16 10 43 24](https://cloud.githubusercontent.com/assets/9007502/10537539/c062bf7a-73f2-11e5-817f-59ed73244c32.png)

Clicking on either of the resources results in a 404. 

Is this an error, or part of some config that needs to be handled on my end? Ideally we'd like users to be able to add attachments to their email/tickets.

rossp:
Do other file attachments work for you? 

What's your media path set to? Do the above files appear in there?

Is it only files attached via email? ie if you edit an issue and attach a file... What shows up then?

reduxionist:
I'm not actually clear on whether this behaviour was triggered by our design for handling submissions with no plain-text part, or if it was caused by one the several bugs that were just fixed by PR #463. Because of that uncertainty, I'm not going to close this ticket yet so that if the cause was a design decision we might still get feedback as to that effect.
There has previously been at least one disagreement with our choice for the approach that could trigger this without there being a bug involved, so we're open to considering alternative options if indeed the problem does persist. Thanks for the report in either case! 😄

PR #463, opened 2016-12-08
refactor all handling of attached files

reduxionist:
Extract attachment processing from forms, views.staff, and management.command.get_email, and consolidate it into a unified lib module function.
Also refactor the affected components, most notably lib.send_templated_email, to make it easier (IMO) to reason about changes to them.
Add unit tests for attachments with UTF-8 filenames, and functional tests for submission of same, as well as ASCII versions, through the public ticket-form.
Remove unused Attachment method "get_upload_to".

### DIFF
--- a/helpdesk/forms.py
+++ b/helpdesk/forms.py
@@ -17,7 +17,7 @@ from django.utils.translation import ugettext_lazy as _
 from django.contrib.auth import get_user_model
 from django.utils import timezone
 
-from helpdesk.lib import send_templated_mail, safe_template_context
+from helpdesk.lib import send_templated_mail, safe_template_context, process_attachments
 from helpdesk.models import (Ticket, Queue, FollowUp, Attachment, IgnoreEmail, TicketCC,
                              CustomField, TicketCustomFieldValue, TicketDependency)
 from helpdesk import settings as helpdesk_settings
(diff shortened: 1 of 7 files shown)


reduxionist:
This is directed towards issues #343, #376, #413, #439, and #456, but because the new tests were written for the refactored code, they do not specially address each issue with its own individual test-case as they would normally. Nonetheless, they attempt to cover the consolidated code path that they all now share. Also, this continues our progress on #444.

pawelmarkowski:
@reduxionist To ensure data consistency my comment from #462  


Great, so I am waiting for your upgrade.

In my opinion there is some kind of bug. If queue.slug does not match matchobj = re.match(r".*[" + queue.slug + "-(?P\d+)]", subject) get_email will create new ticket even if received message was an update for other queue.

reduxionist:
That would definitely be bad, but that section of ticket_from_message is not one I touched in the refactoring, so that brittleness in the parsing has been that way since commit 306d8bf in Feb 2013... Regardless, from having implemented @brunotikami's PR #385 for a client installation, I remember that it used RFC 2822's message-id header to provide a far more robust threading mechanism. There are three more PRs, including one of of yours, waiting to be reviewed after this one. Once those are done, however, and after get testing of this reply-handling in place first, then I can try rebasing that code back into shape again for consideration to merge...
Thanks for your help and patience! 🙇

reduxionist:
Great, thanks @gwasser!

PR #464, opened 2016-12-10
solves Title changed to . (Null title) on failed attachment upload.

reduxionist:
Although the primary content of the original version of this patch, PR #393, was handled as a part of PR #463, this part fixes a separate issue, so I rebased it into a separate commit, preserving @pawelmarkowski's authorship. Thanks for the contribution Pawel! 🍰


reduxionist:
It's not actually my code so I should be allowed to review and approve, but since I can't, I'm using admin privs to merge regardless.

Issue #466, opened 2016-12-12
get_email attachement issue


pawelmarkowski:
Regarding last @reduxionist updates I have a problem with attachment from an email. Attachment has bigger size than original file and it can not be opened. Probably some kind of plain text is added to attached file. I will take a look in a future on that, but maybe author might identify issue faster.

Pawel.

reduxionist:
I'll look into it immediately, it shouldn't matter but just in case, which method of attaching are you using; oubliuc ticket creation, private ticket creation, addition of follow-up, or reply by email?

pawelmarkowski:
Reply by email

reduxionist:
Great, one more detail: roughly what mimetype was the content?

reduxionist:
I had a vague memory of your environment being Python 3 and IMAP, so that's what I tried first and I've already run into two bugs! I apologize for letting them slip through. the test cases I added weren't all mocked, I know there was at least one integration test that went all the way through to writing the incoming attachment out to disk and then reading it back in from the filesystem to compare for equality of UTF8 content -- but I'll add more to cover the current situation and then patch will follow... 🙇

pawelmarkowski:
attachements image/png (new ticket) and application/pdf (update) both unreadable 

User-Agent: Mozilla/5.0 (X11; Linux x86_64; rv:45.0) Gecko/20100101
 Thunderbird/45.4.0
MIME-Version: 1.0
Content-Type: multipart/mixed;

reduxionist:
I'm about to push a branch to my fork with a patch that works for me except under Django 1.8 and Python 3 (issue was reproducible with any binary attachment under any Django with Python 3). So if you can give that one a try hopefully it will work for you too while I write the missing tests to cover this scenario properly. Thanks!

reduxionist:
Sorry for the delay, it took me way, way longer to write the commit message than it should have. If this does in fact fix the corrupt binary attachments issue for you  then I will merge it here with some special-cased handling of the Django 1.8 + Python 3 scenario that would currently break the build. My apologies once again for the bug, I actually thought the present version was so much cleaner than the earlier refactoring which is now back in place... 😊

reduxionist:
[...]

PR #468, opened 2016-12-13
Ticket has illegible file attachements from email

pawelmarkowski:
regarding #466


pawelmarkowski:
TODO: fix python2.7 coding issue (probably add unicode mark)

reduxionist:
Pawel, did you reference the branch on my fork that addressed this issue? Here we seem to be revertinf to an approach we've tried in the past but shich can't work for both py3 and py2. By using dfferent methods of the multi-part message class we can selectively attach as content (good for py2 where there's no penty for treating bytes as strings if the encodings work out) or we can attach as just pathon-dsk which is good for py3 since te std lib will handle the codec operations itself when it needs to). Sorry for typos on my phone... :wink:

gwasser:
I've run into this problem before and thought it was fixed previously :-) it's definitely an encoding issue, and needs to be written so that it works on py2 and py3 (likely needs to use six support?) as encoding handling is vastly different between python 2 & 3. Thanks for starting the work @pawelmarkowski !

pawelmarkowski:
@reduxionist Regarding your firt question - yes, I did.  I fetched your branch and after couple changes and  few tests I merged it and deleted. I think there is a chance to add for example:
@python_2_unicode_compatible

and add: part_attachment.add_header('Content-Disposition', **u**'attachment; filename="{}"'.format(file_name)) symbol. Unfortunately I do not have a possibility to make a tests in py 2.7 

I have seen six.PY3 condition in your code. I thought that the message was not attached due a mimetype and headers that were missing so I changed this part of code too.

tshabs:
I have tested this on Python 3.4.2 w/ Django version 1.9.1 and it solves the issue for me.   Thanks.

reduxionist:
Thanks for the review @tshabs! 😄 

Although @pawelmarkowski's  excellent combined-and-refined patch also LGTM as the solution to this issue, unfortunately it fails to build for Django 1.8 under Py3. The issue is that Django's re-implementation of the standard library's email module complicates Py2/3 compatibility, but they didn't revert back to the standard lib's version until Django 1.9. 💢 

Since there's still another three months before Django's next LTS release, I don't feel we can make our support for 1.8 a Py2-only option just yet... Having taken a few swings at this already myself, the simplest solution I can come up with for now is to split out our 1.8/Py3 support into a separate branch until they release 1.11 this spring.  It's just the templated outbound-mail that needs this special handling, so I'm fine volunteering to handle any dual-patching maintenance-overhead that this might incur. Any thoughts from the maintainer's gallery @flinz, @gwasser et al.?

gwasser:
Hmm, complex situation. I guess my concern is how users will know to use a special branch until then? I appreciate you volunteering to upkeep the separate branch! I'm fine with that as long as we have a bug report or wiki saying to use that branch until 1.11 is out.

I think when 1.11 is out, it would be fair to branch off the 0.2.0 release separately, and have master target 1.11+ (as a 0.3.0 I guess?). So that's longer term solution.

pawelmarkowski:
are you sure is it a problem with lib not with the tests? Can anybody check it under django 1.8? Unfortunately I do not have enough time in next 2-3 weeks.

gwasser:
I merged in the patch now that Django 1.11 is out. We'll switch to 1.11 as the minimum required version, since 1.11 will not remove any features of 1.8-1.10 but fixed the mail handling issues in 1.8 that caused trouble here.

Issue #467, opened 2016-12-12
Can't assign ticket to individual


rando305:
Django 1.9.11
Python 3.4
helpdesk 0.2.0  (according to pip freeze)

in /ticket, I am trying to assign an unassigned ticket to an individual, and nothing happens.
Box is checked in the query results section, I select the person at the bottom in the 'With Selected Tickets' I choose an 'Assign to' individual and click Go.  The clock processes, but nothing happens.

Do you want me to pull a more current version and test again?

reduxionist:
While it usually doesn't hurt to re-install with the latest master to debug with, if you're on 0.20 it means you're at least on a commit that was head more recently than our last official release to PyPI, which was 0.19. I suspect it would be more helpful to get a peak at the stack through whatever method is most convenient in your environment, such as a set_trace before returning from the view, or asserting False if you're running with extensions's runserver_plus, or even just turning DEBUG on if you've got the debug-toolbar installed.
I'm currently redeploying my dev env so that I can get back to bug-fixing, but this is number two on my list after that -- sorry for the delay!

rando305:
I think I'll wait for your next release.  I think I got my production env ahead of my dev env.  That makes it a little complicated to debug the problems I am seeing in production.

reduxionist:
Understood. I'm going to leave this ticket open for now, at least until I get the chance to test this against the latest build. But you can always get back-in-touch with us about regardless od whether I've closed this ticket by then or not... 😄

gwasser:
I checked this out. Looks like the button is missing its appropriate link tag so its non-functional! Oops. Also there's two query boxes in the latest master. Maybe a merge went wonky here. I'll be submitting a PR to fix this soon.

reduxionist:
@gwasser provided a next-day fix for this, but unfortunately I took an accidental two-week absence-from-merging. Please let us know if the issue persists, and please open a new ticket for any other, different issues you may encounter.

PR #469, opened 2016-12-14
Fix form end tag placement, and remove duplicate unstyled query box

gwasser:
Fix form end tag placement, and remove duplicate unstyled query box, to resolve the issue reported in #467 .

Basically, the "Go" button was outside of the form so did nothing. Oops.


reduxionist:
My apologies for the delay on merging this one! I must have accidentally deleted the email for it without noticing because it was news to me when I saw it here after a Christmas break from the issues page... Thanks for another quick-fix @gwasser, and sorry again for the slow merge! 🙇

PR #473, opened 2017-01-08
fix mistake in helpdesk/lib.py (subject_part declared twice, second time without stripping newline)

tshabs:


reduxionist:
Thanks very much for the PR! 🎁

PR #477, opened 2017-01-13
Hotfix/imap encoding errors

reduxionist:
In case you're still available for review @flinz, this is a, perhaps temporary, fix for the two issues that cropped up in #474. No-one was using SOCKS on Python 3 yet unless they were doing all-or-nothing SOCKS anyway, so this won't break any existing installations -- we just might get a request for it if some mixed-SOCKS case wants to upgrade to Py3 in future...


reduxionist:
Thanks @gwasser !

Issue #483, opened 2017-02-09
Can not create queues with long slugs


xnegativx:
The permission_name, and slug fields of Queue model are both limited to 50 chars.
Or permission_name is built as 'helpdesk.queue_access_%s' % slug...
Fix would be  increasing max_length for permission_name

gwasser:
Slugs are typically meant to be pretty short. What do you think is an appropriate max_length?

xnegativx:
Don't disagree. I managed it locally by limiting my slugs, but was rather pointing out the inconsistency there (and you'll end up with ghost permissions in the db as hey are created before the failure)
Well, the minimum max length should be 72 chars. I guess with 100 it will be pretty safe from any update in the prepare_permission_name function.

gwasser:
Oh I understand now, thank you! Yes, the slug (50 chars) gets some stuff prepended to it and result is 72 chars, which doesn't fit in the 50 chars reserved for permission_name. Yep, definitely a bug!

I'll code up a quick patch to expand it to 72 chars to cover slug + prepend.

PR #484, opened 2017-02-10
Change queue permission_name from 50 chars to 72 chars

gwasser:
Changed from 50 chars to 72 chars to accommodate slug + prepended text, to address #483


reduxionist:
Thanks very much for taking care of this @gwasser ! 💎

Issue #481, opened 2017-01-29
Install documentation missing SITE_ID


J-A-M:
I previously mentioned a documentation issue, there is still a problem.

To get it to work I had to remove     'django.contrib.sites',  from INSTALLED_APPS.
With this missing I can login and configure (after making the change in the [previous post](https://github.com/django-helpdesk/django-helpdesk/issues/480))

However when I press submit on a new ticket I get
`RuntimeError: Model class django.contrib.sites.models.Site doesn't declare an explicit app_label and isn't in an application in INSTALLED_APPS.`

But, if I restore django.contrib.sites I get

`DoesNotExist: Site matching query does not exist.`
for /login/

J-A-M:
According to [This answer](http://stackoverflow.com/questions/35388637/runtimeerror-model-class-django-contrib-sites-models-site-doesnt-declare-an-ex/35389220) you need to add
`SITE_ID = 1` to settings.py
This seems to work

gwasser:
Yes now that you mention it, I had to use `SITE_ID = 1` on my testing project and it somehow slipped my mind to add this to documentation. Good catch!

reduxionist:
Thanks for the report @J-A-M and thanks again for the fix @gwasser! 

Closed by #485

PR #485, opened 2017-02-10
Update docs to add SITE_ID, to address #481

gwasser:
Update docs to provide basic setup for django.contrib.sites by adding SITE_ID, to address #481.

See https://docs.djangoproject.com/en/1.10/ref/contrib/sites/#module-django.contrib.sites

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -35,13 +35,19 @@ Adding To Your Django Project
         'django.contrib.auth',
         'django.contrib.contenttypes',
         'django.contrib.sessions',
-        'django.contrib.sites',  # Required for determing domain url for use in emails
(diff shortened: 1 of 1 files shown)


reduxionist:
Thanks again @gwasser, not just for providing a fix, but for taking the time to follow through with the reporter's scenario to find it in the first place! 🍰  Since we're explicit about not packaging a stand-alone project, I take it for granted that all my development and issue debugging occurs in an empty-shell django-project and so it's easy to forget all the details that go into a bare initial setup.

I think it would be useful to users considering the project if we did supply a demo project like many pluggable apps do nowadays. It could also serve as a test environment, replacing quicktest. More "normal" IMO, faster for users to get a taste, less setup issues -- which are pretty common at the moment, etc.

PR #487, opened 2017-02-14
`get_email` management command: Force message to unicode to avoid encoding issues

mikhuang:
On py2.7 I got `UnicodeEncodeError: 'ascii' codec can't encode character u'\u2013' in position 1599: ordinal not in range(128)` type errors until I explicitly declared the message to be Unicode.


gwasser:
Hi @mikhuang thanks for the bug report & patch! Looks like the patch broke Python 3 support though :-) maybe we need to use `message.encode('utf-8') if six.PY2 else message` or something similar?

Good catch on the missing namespace for logout! I hindsight I probably don't logout as much as I should and didn't notice :-)

gwasser:
@mikhuang I wonder what message you were parsing with `get_email` that caused this? I would like to see that message added as a test case if you don't mind! (or if the contents are personal, cut it to make the smallest test case that illustrates the fail). Obviously our old tests weren't catching it :-)

mikhuang:
It was the Zoho new account intro email. The particular one that gave me issues was ingested by django-helpdesk and deleted from IMAP. Now I'm not able to reproduce the error (signed up with another account) and stuck wondering if that was the original issue after all. :-/

gwasser:
Thanks for looking into it! I suppose any email with unicode characters might work. Specifically `u'\u2013'`

gwasser:
I'm trying to update our tests for this to include unicode characters. I added `\u2013` to the text string of the test email message just to see what happens, and am getting errors on both Py2 and Py3.

`test_get_email.py", line 135, in test_read_email
    self.assertEqual(ticket1.description, "This is the helpdesk comment with a unicode character \u2013 via email.")
AssertionError: 'This is the helpdesk comment with a unicode character \\u2013 via email.' != 'This is the helpdesk comment with a unicode character – via email.'
- This is the helpdesk comment with a unicode character \u2013 via email.
?                                                       ^^^^^^
+ This is the helpdesk comment with a unicode character – via email.
?`

I'm not an encoding expert so I'm not sure what to do here. It looks to me as if the unicode character's representation is getting converted to text and dropped into the Ticket description.

flinz:
Unicode consistency is actually a really needed test case for us, which we've encountered in quite a few scenarios, and which might have been fixed previously in an ad hoc fashion. I can't tell you off the board how to fix it, but I'm glad this is coming up. I'll try to have a look soon, but if @reduxionist wants to have a look he's of course welcome to.

reduxionist:
@gwasser we're mostly using the 'replace' option to handle unicode errors in this module, so I believe that all you are seeing in your tests is that we are successfully re-encoding the string using utf-8. I think that's the correct behavior, but if there's some other approach we're missing please do point it out to me. 😄  Otherwise, an adjusted version of your test should make this PR ready-to-merge!

gwasser:
I submitted PR #501 to address this. My new unit test was catching the same error message as @mikhuang in Py2.7 but I believe I fixed it. Once #501 is committed, should be able to commit this one too.

reduxionist:
Thanks for the contribution @mikhuang! With @gwasser's diligent pursuit of your lead, this PR lead to our fixing more than just the bug it patches directly, but to a handful of other encoding landmines buried in the same module as well! 🎁

PR #490, opened 2017-02-25
Update Travis CI to include Python 3.6

gwasser:
Some linux distributions (like OpenSUSE Tumbleweed) are already include Python 3.6, so makes sense to start testing against it.


flinz:
👍 thanks!

reduxionist:
3.6 has been a great Python update for me, glad you noticed to include it here @gwasser ! 😄

PR #491, opened 2017-03-08
Fix and update russian translation

Yuego:
The href attribute of the link was always empty.
Added translations to multiple lines.


gwasser:
I updated the Russian translations on our transifex account based on your contribution. Thanks again!

PR #497, opened 2017-03-17
Update es_MX translations

gwasser:
Thanks to @andresf01 for updated es_MX translations. Committing on his behalf.

For the record, since I'm new to Transifex, this is what I did:

1. Logged in to Transifex, downloaded the "Ready for Use" .po file
2. Used command `msgfmt es.po --output-file es.mo` to generate the .mo file using gettext, based on http://stackoverflow.com/questions/739314/easiest-way-to-generate-localization-files

I believe this was correct but please let me know if the process should be different!


reduxionist:
Normally I use the 'compilemessages' management command to generate the binaries myself, but AFAIK that doesn't offer any advantages other than the project directory integration, so if going straight to msgfmt is easier then by all means!

Issue #488, opened 2017-02-15
Provide a Django demo project


gwasser:
As @reduxionist mentioned in #485 , might be a good idea to have a demo project.

> I think it would be useful to users considering the project if we did supply a demo project like many pluggable apps do nowadays. It could also serve as a test environment, replacing quicktest. More "normal" IMO, faster for users to get a taste, less setup issues -- which are pretty common at the moment, etc.

gwasser:
Closing this issue now that I've submitted a pull request. Further discussion can continue on #498.

PR #498, opened 2017-03-20
Add a demo project and update docs

gwasser:
A demo project is now available to address issue raised in #488. The demo project includes a fixture with some starter queue and ticket info so that a new user may quickly get started and click around on live tickets without needing to set anything up.

Not only does it allow easier/faster demos and provide a template for starting new projects, this could be a good baseline for developers to test against.

I reworked the README to talk about the new demo project first thing. As I was re-writing it I realized there was a certain amount of redundancy between the the README, the CONTRIBUTORS file, and the docs/ files. I cut some stuff from the README and moved it to the docs/, to keep the README file shorters (more of a "quickstart" than an all-info document; IMO it should mostly just point at the other docs as like an index).


reduxionist:
Thanks for holding down the fort single-handedly all this time! :1st_place_medal: I've been keeping a nervous eye on the notifications arriving in my inbox but have been able to just guiltily sigh with relief as they're handled without me... 🙇  I will be out of crunch-mode this week (delivery was today but there's still some hand-holding to do), so I should be able to review this PR in just a couple days if no-one else gets to it first.

Thanks again for putting in all the effort single-handedly of late, it's great Ross is still around and could grant the auth'z you were missing; having many hands on deck sure makes a difference! 🙏

gwasser:
No problem, happy to be able to fill in while you're busy :-) hope you get a break from crunch-time soon!

reduxionist:
Wow, that is an awesome demo installer! Simple considering the flexibility in environments that it provides, it worked flawlessly even in my most tortured Python installation. I think the sample data provided is just enough for users to click their way through to a fast, hands-on. understanding of the system far more easily than they could before. Great job @gwasser! 👏

PR #499, opened 2017-03-26
Migration added

Yuego:
Not all changes are reflected in migrations.
Django constantly tries to create a new migration, but in my case it can not do this because of lack of access rights.
This gives me some inconvenience in development - I can`t use 

`python manage.py makemigrations`

 globally.
If you do not plan to make changes to the model in the near future, I think this would be the solution.


reduxionist:
Issue confirmed and fix verified using our new handy demo project (thanks again @gwasser !).

Thanks for the PR @Yuego, we appreciate the contribution! 😄

PR #501, opened 2017-04-03
Updated get_email tests

gwasser:
I updated our plain text email test to use unicode characters and added a MIME multipart email test for get_email script, partly to address the issue brought up in #487. I believe this will be more robust test and expand our coverage!

This test lead to a possible bug fix in get_email. For the email body, the unicode character was getting doubly escaped (\\u...) and so wasn't appearing correctly when compared against a unicode string. I fixed this with a unicode_escape encoding, and cleaned up a few unnecessary encoding steps.

I also snuck in a missing dependency to the demo project and added knowledgebase data to complete the demo. (Bad form, I know, but got committed to wrong branch, so oops :-) )


reduxionist:
You fixed more encoding issues than I was even afraid were still left! Great job coding up those badly needed new test cases, that was an important code path we hadn't covered yet. That 5% coverage bump is as high a single-commit improvement as I think we've ever had, and you brought us to the 60% mark -- awesome contribution @gwasser !! :1st_place_medal: 

p.s. Since you were sneaking in some demo updates, I took the liberty of adding to them with tiny makefile doc updates that wouldn't warrant a merge of their own... 😉

Issue #184, opened 2013-09-23
Problem installing fixture: Incorrect string value


chrisspen:
I tried installing django-helpdesk 0.1.11 from pypi, and running `manage.py migrate helpdesk` fails with:

```
Warning: Problem installing fixture '/usr/local/myproject/.env/local/lib/python2.7/site-packages/helpdesk/fixtures/initial_data.json': Incorrect string value: '\xD0\x97\xD0\xB0\xD1\x8F...' for column 'heading' at row 1
```

So I tried installing master with:

```
pip install -U https://github.com/rossp/django-helpdesk/archive/master.zip
```

but this also fails with:

```
Downloading/unpacking https://github.com/rossp/django-helpdesk/archive/master.zip
  Downloading master.zip (853Kb): 853Kb downloaded
  Running setup.py egg_info for package from https://github.com/rossp/django-helpdesk/archive/master.zip
    /usr/lib/python2.7/distutils/dist.py:267: UserWarning: Unknown distribution option: 'include_package_data'
      warnings.warn(msg)
    /usr/lib/python2.7/distutils/dist.py:267: UserWarning: Unknown distribution option: 'zip_safe'
      warnings.warn(msg)
    /usr/lib/python2.7/distutils/dist.py:267: UserWarning: Unknown distribution option: 'install_requires'
      warnings.warn(msg)
    usage: -c [global_opts] cmd1 [cmd1_opts] [cmd2 [cmd2_opts] ...]
       or: -c --help [cmd1 cmd2 ...]
       or: -c --help-commands
       or: -c cmd --help

    error: invalid command 'egg_info'
```

What are causing these errors? I'm using MySQL as my db backend. Please let me know if I'm doing anything incorrectly.

mrkiwi-nz:
As @chrisspen  says, utf8_general_ci is a safe choice for collation. This error will occur whenever you have created your db using a collation that can't handle the characters in the fixture.

For example these wont work; "big5", "cp-1251" 
These do work : utf8_general_ci or utf16_general_ci

If you really want to keep your current collation, and you don't mind editing the fixtures file, you can delete the russian and german email templates (lines 194 to 386) in the file emailtemplate.json
These templates contain a number of unicode characters which cause the error when trying to shove them into fields which have an incompatible database collation.

Maybe a line in the installation notes would help?

gwasser:
Added documentation to the installation notes as part of #507 so closing. Let us know if I missed anything!

Issue #95, opened 2011-12-19
Handle carbon copied addresses on tickets submitted via email


fatbox:
If a ticket submitted by email has CC'd recipients they are currently ignored. They should be added to the CC list for the Ticket that gets created.

See also GH-18. My feeling right now is that CC'd addresses should have _view only_ permissions to the tickets, but perhaps we can make it a configurable option.

gwasser:
This was implemented as part of #507 so closing. Let me know if you run into any bugs while testing!

Issue #72, opened 2011-09-24
django code in comment field does not get escaped


kotowicz:
I had a ticket where I had to post some django code. The exact comment was:
_"commented out the {% if new %} check for the forms, since we don't use this button anyway"_

When I try to submit this comment, I get a crash, because 'endif' is not found:
_TemplateSyntaxError: Unclosed tags: else, endif_

gwasser:
I confirmed this issue still exists in 0.2.0, and can potentially even be a security concern. Will take a look at a better way for template processing.

PR #507, opened 2017-04-16
Fix Few Bugs and Update Mail Handling

gwasser:
Grouped a few tasks into one PR:

Addressed #72 by basically commenting out Django tags with a simple search. Not a perfect solution, but it works since I think we really only want to give commenters the ability to access ticket/queue data, and that's it.

Addressed #184 by adding the suggested SQL to the install.rst file in docs/.

Implemented #95, which looks for CC'd addresses in the email headers, and adds them to the Ticket with view permissions. Also updated tests to handle CCs.

Also small tweak to Travis CI: Django 1.11 was released recently so might as well start testing against it! My understanding is that the Django 1.8 -> 1.11 shouldn't be a problem.


gwasser:
Hey @reduxionist thanks for the review! Completely slipped my mind that the DB doesn't enforce uniqueness, was thinking it already did somehow.

Reworked it to get a list of currently CC'd emails and compare to CCs in new email, and only add the actual no ones, to prevent duplicates. Extended the test case with some duplicate CC emails to ensure that the duplicates really are left out (they are).

gwasser:
Ok should be ready for review again. Thanks!

gwasser:
I think I get what you're saying: one of the CC'd emails could be a helpdesk staff/user and so want to make sure that user isn't duplicated in CC?

reduxionist:
Yes I think so: a staff member who is already subscribed to the ticket, because they are submitter or owner for example, should not get added as a CC just because they happen to have been CC'd on the message we're processing.

gwasser:
Thanks for the feedback again @reduxionist ! Modified get_email and its test to now detect Users when they are CC'd on emails, and add the User as a CC rather than their email only if they aren't already CC'd or the Assignee or Submitter.

gwasser:
OK It's ready for review again after some PEP-8 fixes (I should have just run it locally but I kept thinking I for sure had it :-) )

reduxionist:
It was a nice collection of bug fixes from the start, but I want to thank you explicitly for the patience and persistence you gave to bugs I only described in theory, that was some top-notch professionalism and dedication on your part Mr. @gwasser Sir! 🎖

gwasser:
Thanks @reduxionist ! Your reviews helped a lot. While I think the tests for get_email I added here were sufficient, I plan on cleaning them up and adding some more for improved coverage next in a new PR.

PR #515, opened 2017-06-02
Added 'es' email templates in JSON file

andresf01:
I had a pair of days free so I improve the email templates with translation to 'es' and I added to emailtemplate.json file.


gwasser:
Hi @andresf01 thanks for providing these translations! looks good to me except for one question, why did you add the angle brackets on line 776? (<assigned_cc></assigned_cc>). I think that will cause some problems as that template will not be accessible at run time. Please change back to just "assigned_cc" and then I'm happy to merge!

gwasser:
Nevermind @andresf01 I was able to push a quick patch to fix this. I'm guessing your text editor might have auto changed it to an xml tag?

andresf01:
Thank you @gwasser I think so.

Issue #521, opened 2017-06-14
Update setup.py


mrkiwi-nz:
Travis is testing django 1.8.18, 1.9.13, 1.10.7, 1.11
In the "classifiers" list, setup.py mentions
        "Framework :: Django",
        "Framework :: Django :: 1.8",
        "Framework :: Django :: 1.9",
        "Framework :: Django :: 1.10",

Should "Framework :: Django :: 1.11" be added here?

gwasser:
Yep, we should include 1.11 too. It's not a super big deal, but I'm planning on a little commit to update the third party stuff so I'll toss this in too.

gwasser:
Was resolved in PR #522.

Issue #520, opened 2017-06-14
Is mock a requirement?


mrkiwi-nz:
while running 
`python quicketst.py helpdesk`

`/helpdesk/tests/test_attachments.py", line 20`
`import mock`
`ImportError: No module named mock`

I can't see mock in either requirements.txt or requirements-testing.txt

Maybe mock is usually pulled in by another of the packages in requirements-testing.txt, but it didn't work for me.
pip install mock
fixes it for me (also pulls in pbr)

After that, 46 tests run fine.

So the fix is just adding mock to requirements-testing.txt ?

gwasser:
Based on the command you used, I think you're running this on python 2. `mock` is a standard package for python 3, so I think what happened is the requirements-testing.txt reflects python 3 dependencies.

Maybe we can make a separate requirements-testing.txt file for each version of python?

gwasser:
Actually I just tried it and it seems to work fine for python 3 to list `mock`. So I'll just make the change in my next PR. Thanks!

gwasser:
This was added in #522 so I believe it was resolved, but let me know if there's still an issue.

PR #522, opened 2017-06-14
Update statics

gwasser:
Updates the static third-party dependencies to the latest versions. Probably more clean up is needed (I think some dependencies are no longer used and should be cleaned but haven't had time to check yet).

I also added simple fixes for #520 and #521 .


gwasser:
I suspect since this only updates the static includes and not any code, there aren't any objections from the other reviewers to merge this is now for better testing :-)

Issue #579, opened 2017-12-20
improvements to public-facing helpdesk


gwasser:
From #569, we got the following comment from @yugensoft, I moved it to a new ticket for better tracking:

> Basically for public-facing customer help, the following options are needed:
> 
> -     Tickets should only be prioritizable (or due date change) and assignable by admins
> -     The email address of the creator shouldn't be shown (privacy is important)
> -     The yellow view tickets panel has css overlap issues on the text (doesn't bother a dev, but things like that hurt confidence from a customer)
> -     The interface is unclear to a customer. I.e. dashboard only makes sense (assigned to you etc) in a JIRA-like context. They are more likely to click "Tickets", but that shows a search interface (but calls it "Query" which isn't as familiar to a customer). It would be better to have an option whereby (for "customer helpdesk mode"), they only see a "Tickets" tab, and that tab just shows a list of tickets, starting with their own. The search panel could be off to the side, or in a separate view, or a button + modal. At the moment it's taking up all of the page, so a customer has to scroll down to see tickets, something that is jarring to them (they expect just to see tickets when they click tickets).
> -     The base email templates should probably be editable in the database too (otherwise the user is having to fork just to edit that template).
> 
> Hope that info was useful, cheers

arnuschky:
I second every item on this issue. Maybe they should be separated out though. The first item seems to be covered by #503 already.

arnuschky:
I just added PR #579 that adds three new settings that allow the operator to pre-define the queue, priority and due_date fields for public tickets. If one of these settings are present the corresponding input
field is hidden from the form.

Issue #503, opened 2017-04-13
Hide default fields


willstott101:
**Thanks for the awesome work here!** It would be great to see progress on the public facing web interface as well as the email system. I may do some work on this in the near future.

I noticed that there is currently no easy way to hide default fields... for our use case we wanted to hide the `due_date`, `priority`, and `assigned_to` fields from non-staff users. I ended up achieving that with the following patch. However some view code relies on the existence of the fields in the fields dictionary, so this is not complete as a PR.

Luckily the field expected by the view was the `assigned_to` field which has it's own special snowflake setting: `HELPDESK_CREATE_TICKET_HIDE_ASSIGNED_TO`

```
diff --git a/helpdesk/forms.py b/helpdesk/forms.py
index 7ab2024..d1b6994 100644
--- a/helpdesk/forms.py
+++ b/helpdesk/forms.py
@@ -317,6 +317,8 @@ class TicketForm(AbstractTicketForm):
         """
         super(TicketForm, self).__init__(*args, **kwargs)
         self._add_form_custom_fields()
+        for f in getattr(settings, 'EA_HIDDEN_TICKET_FIELDS', []):
+            self.fields.pop(f, None)
 
     def save(self, user=None):
         """
@@ -369,6 +371,8 @@ class PublicTicketForm(AbstractTicketForm):
         """
         super(PublicTicketForm, self).__init__(*args, **kwargs)
         self._add_form_custom_fields(False)
+        for f in getattr(settings, 'EA_PUBLIC_HIDDEN_TICKET_FIELDS', []):
+            self.fields.pop(f, None)
```

gwasser:
Thanks! I've been looking at the email system a bit myself but as far as I'm aware no one is working on the public interface yet, so greatly appreciate your help with patches!

Offhand I'm not sure I like the introduction of a new `EA_HIDDEN_TICKET_FIELDS` variable. Do you think we might be able to accomplish it by making a more basic form class, and then making the staff form a subclass with more features, rather than popping off stuff?

willstott101:
Yes that's indeed a possible technique. However that wouldn't really make it much more customisable. More recent versions on Django have a 'disabled' option for form fields. And there's also the HiddenField form Widget. I wonder if there's a way to implement this similar to how I've done it as a layer after building the class. But possibly at the template / widget level. Because:

1. View/Form code is expecting some of the fields to be present (refactor-able).

[...]

Issue #289, opened 2015-01-18
Hide queue selection when there's only a single queue


rossp:
For simpler helpdesk installs where there is only a single queue configured, the queue selection should be hidden (default would always be the single queue)

arnuschky:
I just added PR #579 that adds three new settings that allow the operator to pre-define the queue, priority and due_date fields for public tickets. If one of these settings are present the corresponding input
field is hidden from the form.

It's doesn't address this issue directly but solves indirectly by setting the default queue name.

DavidVadnais:
I went in and deleted the second queue from `make rundemo`.  Then I went to make ticket and edit ticket.

<img width="803" height="252" alt="Image" src="https://github.com/user-attachments/assets/6052a706-6d3f-4040-97b6-2e290d085300" />

<img width="2244" height="412" alt="Image" src="https://github.com/user-attachments/assets/c58f9069-8e24-44fe-b589-0c06f49f75a7" />

The fields do auto assign but they are there. The goal here is to not show the queue option with one queue I believe.

PR #583, opened 2018-01-09
Adds pre-defined values for public tickets

arnuschky:
This commit adds three new settings that allow the operator to
pre-define the `queue`, `priority` and `due_date` fields for public
tickets. If one of these settings are present the corresponding input
field is hidden from the form.

The settings are the following:

    HELPDESK_PUBLIC_TICKET_QUEUE = 'website'
    HELPDESK_PUBLIC_TICKET_PRIORITY = 2
    HELPDESK_PUBLIC_TICKET_DUE_DATE = ''

If the due date is set to the empty string, no due date is saved in the
ticket (analogously to the form). The other settings should be
self-explanatory.


arnuschky:
This addresses issue #503 and partially #289  and #579.

This patch was made against the 0.2.4 release as I was unsure which version I should use. I can rebase if need be.

helpdesk/views/public.py
```diff
@@ -58,6 +59,19 @@ def homepage(request):
...
+        # add pre-defined data for public ticket
+        if getattr(settings, 'HELPDESK_PUBLIC_TICKET_QUEUE', None) is not None:
+            # get the requested queue; return an error if queue not found
+            try:
+                queue = Queue.objects.get(slug=settings.HELPDESK_PUBLIC_TICKET_QUEUE)
+            except Queue.DoesNotExist:
+                return HttpResponse(status=500)
+        if getattr(settings, 'HELPDESK_PUBLIC_TICKET_PRIORITY', None) is not None:
+            initial_data['priority'] = settings.HELPDESK_PUBLIC_TICKET_PRIORITY
+        if getattr(settings, 'HELPDESK_PUBLIC_TICKET_DUE_DATE', None) is not None:
```
willstott101: What if I want the default 'due_date' to be `None`?

willstott101: Right, I missed your initial comment. Would it not just be easier to use `hasattr` as opposed to having special values?

arnuschky: Err, yes, I guess so. :) I was unhappy with the special value but couldn't think of an easy way to get rid of it.

Regarding None: As we pre-set this value via a hidden in the form, I think we always end up with an empty string. I should test this though, I am not 100% sure of Django's behavior here.

arnuschky: It's actually the other way round: the empty string always gets converted to `None` for datetime fields. I updated the PR accordingly.

gwasser:
Looks pretty good and simple. I'll wait to merge in case you're working on documentation updates. Thanks for the contribution!

Issue #576, opened 2017-12-13
Demo ist broken


TobitRE:
I tried to login with demo demo into the demo an get this error.

http://django-helpdesk-demo.herokuapp.com/view/?ticket=demo&email=demo&csrfmiddlewaretoken=mZyx0NIeJ3onk418nlZA0Q5sR1ln9u5JFvGitexUsE0kZ00pa0ebecKpdSujw0IN

<img width="1422" alt="bildschirmfoto 2017-12-13 um 14 18 41" src="https://user-images.githubusercontent.com/4648804/33940795-b6328d82-e010-11e7-9974-617ebc271b5e.png">

gwasser:
Looks like you entered demo/demo in the "View A Ticket" form on the main page. You enter a ticket ID and your email address there, it is meant for someone without an account to view only the ticket they submitted.

Please click "Log In" at the top of the page, then enter demo/demo, to log in as a staff user.

gwasser:
Actually we should more gracefully handle this error in case someone makes that same mistake. I'll reopen and look at a fix.

gwasser:
I made a quick patch for this to redirect you to the home page instead of giving an error. Not perfect but there's some more permanent fixes coming, so I'll go ahead and close this for now. I'll update the demo soon when version 0.2.4 is released. Thanks for the comment!

PR #586, opened 2018-01-12
Fix for Email Issue while pulling CC and ValueError for public homepage

ddwaraka:
It fixes the way some email providers format the CC field and catch Exception for ValueError at public view ticket function.

### DIFF
--- a/helpdesk/management/commands/get_email.py
+++ b/helpdesk/management/commands/get_email.py
@@ -412,7 +412,7 @@ def ticket_from_message(message, queue, logger):
         for ccemail in new_cc:
             tcc = TicketCC.objects.create(
                 ticket=t,
-                email=ccemail,
+                email=ccemail.replace('\n', ' ').replace('\r', ' '),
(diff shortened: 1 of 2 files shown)


ddwaraka:
No problem at all! Thank you guys for all the great work you've been doing! 
Also, this commit addresses [this issue,](https://github.com/django-helpdesk/django-helpdesk/issues/576) at least for now, when someone enters a string or anything that is not in the format of a general ticket ID. 

It'll accept these inputs in the format of:

1. queue_name-ticket.id_integer - Ex: billing-2527
2. ticket_id.integer - Ex: 2527

So, its just not when the value is a non integer, I think it's the '[' and ']' around the ticket that is causing it. Ex: Entering [billing-2527] will cause it to crash. 

Now, I don't know how good this fix is, but it should be a placeholder for now.

gwasser:
Good point, should help a little until a more thorough overhaul of the public form is done. Thanks again!

PR #630, opened 2018-08-19
Don't redirect to staff only ticket submission page

timthelion:
When HELPDESK_SUBMIT_A_TICKET_PUBLIC is set to True, users's should be able to click the "Submit a Ticket" button without being asked to log in.

### DIFF
--- a/helpdesk/forms.py
+++ b/helpdesk/forms.py
@@ -301,21 +301,27 @@ class TicketForm(AbstractTicketForm):
         help_text=_('This e-mail address will receive copies of all public '
                     'updates to this ticket.'),
     )
-
     assigned_to = forms.ChoiceField(
-        widget=forms.Select(attrs={'class': 'form-control'}),
-        choices=(),
+        widget=forms.Select(attrs={'class': 'form-control'}) if not helpdesk_settings.HELPDESK_CREATE_TICKET_HIDE_ASSIGNED_TO else forms.HiddenInput(),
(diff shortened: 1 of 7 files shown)


gwasser:
I'm not sure that redirecting to `home` is the best solution here either. Our `submit` page should probably know how to handle a logged-in user vs a non-logged-in user. The `develop` branch is still under active work so maybe we can overhaul the permissions.

timthelion:
Currently the `create_ticket` view which appears in `staff.py` uses the `TicketForm` form. Where-as the form presented on the public homepage uses a `PublicTicketForm`. Futhermore, in `staff.py` and `public.py` you will find that there is a relatively large amount of code in the `create_ticket` and `homepage` functions that is not shared and quite a bit more that differs in subtle ways. I'll have to think more on it, but I'm not yet convinced that combining these forms and functions makes sense.

However, it would make sense to me to create a working and independent public ticket submission because it would make sense to be able to link to a specific queue's form.

gwasser:
There's definitely some refactoring and clean up that needs to happen, I agree. especially around the forms. A lot of stuff started out hardcoded and can be either turned into a library or replaced with some third party library.

timthelion:
Just pushed a new version of this pull request. Its a rather large change and there are tones of fiddlybits. So many different settings.py options :( so I'll bet $10 that I broke something for someone. But it works for me :)

PR #631, opened 2018-08-19
Add user profile links for submiters who have profiles

timthelion:
This adds a user profile link (visible only to admins) to the view ticket page. The link points to the django admin user profile edit view.

link

The link is shown only if the email address is the same as that of a user in the system.

### DIFF
--- a/helpdesk/models.py
+++ b/helpdesk/models.py
@@ -9,6 +9,7 @@ models.py - Model (and hence database) definitions. This is the core of the
 
 from __future__ import unicode_literals
 from django.contrib.auth.models import Permission
+from django.contrib.auth import get_user_model
 from django.contrib.contenttypes.models import ContentType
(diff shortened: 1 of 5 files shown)


timthelion:
Does the develop branch test suit succeed? It shows it as passing in the README flair, but it doesn't successfully run even without my changes on my local system.

timthelion:
Note: I just changed the profile button from btn-warning to btn-primary, so the screenshot is out of date.

gwasser:
Hello @timthelion I don't think `develop` quite passes tests yet, I recall there being a couple remaining bugs to squash that I haven't gotten to yet. The README flair I think is connected with `master` still and incorrect, I'll have to fix that.

This seems like a pretty simple change you can also add to `master` if you'd like for the 0.2.x release.

Issue #633, opened 2018-08-20
RSS feeds should be staff only and not available to any logged in user


timthelion:
If I go to `helpdesk/rss/queue/<slug>/` I can see all tickets in the entire system even if I'm a non-staff user.
https://github.com/django-helpdesk/django-helpdesk/blob/master/helpdesk/urls.py#L172

gwasser:
Shouldn't the `login_required` decorator make you login? Right now the permissions system is very weak and doesn't do a good job differentiating staff from other users unfortunately, that's a to-be-updated in the next major release.

gwasser:
Fixed in #636 so closing the ticket. Thanks for the contribution!

PR #636, opened 2018-09-03
Only staff users should be given access to the RSS feeds

timthelion:
Currently, non-staff users can view rss feeds, which is a big security flaw IFF you have non-staff users who shouldn't have access to everything that flows through the helpdesk.
This patch fixes that.


timthelion:
Fixes: https://github.com/django-helpdesk/django-helpdesk/issues/633

gwasser:
Thanks! This is great.

Issue #647, opened 2018-09-26
Are the attachments protected?


johanovic:
I see that html emails are stored in the media folder (which is nice), but I do have a question. The url is

> /media/helpdesk/attachments/email-44/119/email_html_body.html

Does that mean these attachments are all public? That would be a major problem. Anyone aware of the name of the ticket queue could start reading other people's email. 

Or am I missing something?

gwasser:
Confirmed using the demo, I could navigate directly the to the attachment URL even though I wasn't logged in. Thanks for reporting! This is definitely a potentially big issue since the URL is largely guessable. A persistent user could simply keep trying values until found one that worked, and at least read some emails.

We've been working on cleaning up the user permissions and access code lately in the `develop` branch so I'll mark this as a bug/security issue and we'll try to think of a solution before the next release. It seems like we could add some user checks on any attachment URL to ensure you must be logged in to view.

johanovic:
I can't find it right now (under time pressure) but whilst navigating through the code/PR's yesterday I saw that someone had added a UUID to secure the public ticket display. Something ?ticket=...?email=...?key=...  

This type of unique key for a ticket would make a lot of protection easier.

I've used an alternative solution with send_file (or its alternative on nginx) to serve protected files, which is also an option, but which does not work out of the box. 

Yet another alternative is to create an protected media folder which serves the attachments in a django view. That has the advantage of working out of the box, but the disadvantage that django actually handles the static files. For email this might not be such a huge issue given the limited size of attachments. Also for a helpdesk there won't be 1000s of simultaneous requests anyway.

johanovic:
Found it, and... you wrote it :)

https://github.com/django-helpdesk/django-helpdesk/commit/a3bc6ed46bb1a00ce4664dfeac155e4b11b842a8#diff-1bf25f6a5d5de1ed7c39dbbd61760137R488

gwasser:
I can't take credit, I just did the merge request :-) this was a great start by @timthelion.

We might be able to use the extra URL keys to protect it as you suggested.

timthelion:
This might be a very easy fix. Simply change:

https://github.com/django-helpdesk/django-helpdesk/blob/develop/helpdesk/models.py#L784

To

    
    path = 'helpdesk/attachments/%s/%s' % (instance.followup.ticket.ticket_for_url, instance.followup.secret_key)

And write a test ;)

I haven't tried it though.

gwasser:
[...]

PR #649, opened 2018-10-06
Include ticket secret in path to new attachments in order to reduce URL guessability.

timthelion:
I didn't write tests because I'm not imaginative enough to think up a meaningful one.


timthelion:
Fixes #647

gwasser:
Pretty simple change to the URL, thanks! Will have to think on some good tests to ensure this works as expected.

PR #650, opened 2018-10-15
Added optional serverside processing on datatables that lists all tic…

ddwaraka:
I've added serverside processing for datatables on the tickets_list.html page. I've made optional via a flag in settings.py. It is set to True by default. 

Datatables will do an ajax request to fetch the queryset after the page has loaded, which is why I'm using django's low level cache API to store it and and then retrieve it for datatables. 

This is my first major commit to github, I may have made mistakes - logical or syntactical. I have tested the feature, it worked well. Please let me know if there is a lack of documentation, comments or any other mistakes, I am happy to learn!

Searching is now different on the tables for date fields (created and due_date), because the naturaltime format is not stored on the DB, it is not possible to search it with that string. Instead, users will have to use "2018-10-4" as the search string to search for tickets created on October 4, 2018.

### DIFF
--- a/helpdesk/templates/helpdesk/ticket_list.html
+++ b/helpdesk/templates/helpdesk/ticket_list.html
@@ -6,32 +6,100 @@
 
 <script src='{% static "helpdesk/filter.js" %}'></script>
 <script>
-$(document).ready(function() {
+    {% if server_side %}
+        $(document).ready(function()
+        {
+          let url = ""
+            //DataTables Initialization
+            let tasks_table = $('#ticketTable').DataTable({
+                "language": {
+                    "emptyTable": "{% trans 'No Tickets Match Your Selection' %}"
+                },
+                "processing": true,
+                "serverSide": true,
+                "ajax": {
+                    "url": "/datatables_ticket_list/",
+                    "type": "GET",
+                },
+                createdRow: function( row, data, dataIndex )
+                {
+                    $( row ).addClass(data.row_class);
+                },
 
-    $('#ticketTable').DataTable({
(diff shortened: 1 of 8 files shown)


gwasser:
Thanks! This looks great and I think gets us on the way to using django REST framework (to address #198)

gwasser:
This is only in `develop` right now. If you can send the diff or submit a new PR I'll merge it. I can try to pull out the patches myself later but don't have time at the moment. There's another proposed change to our current branch so I'll probably merge these two together into `master` and make another release.

ddwaraka:
That's okay, we can wait for the newer release, probably come with more fixes/features! Better for us as well! Thanks!

gwasser:
I'm hoping to have the newer release out by the end of November, but please reply and bug me about it if you don't see much progress by end of month.

helpdesk/templates/helpdesk/ticket_list.html
```diff
@@ -6,32 +6,100 @@
...
+                        return data
+                      }
+                  },
+                  {"data": "priority"},
+                  {"data": "title",
+                    "render": function (data, type, row, meta)
+                    {
+                      if (type === 'display')
+                      {
+                        data = '<b><a href="' + url + '" >' + data + '</a></b>';
```
timthelion: This is not the correct url. The url set in line 35 is out of scope at this point. https://github.com/django-helpdesk/django-helpdesk/issues/681

gwasser: Merged pull #699 to fix this, thanks.

PR #651, opened 2018-10-17
To make sure that, a single user can't vote more than once (KBItem)

### LINKED ISSUE #638: You can upvote/downvote kb articles as many times as you want
Each user should only have one vote.

shashwat1002:
fixes #638 , by adding a new field to the KBItem model  (ManyToMany), to keep  track of all users who have voted.  The vote view checks whether a user has voted.


gwasser:
Thank you for your work! But this commits things to a build directory. Can you close this and resubmit as a commit diff on the actual files? I can't tell what was changed here. Thanks!

PR #652, opened 2018-10-18
To make sure that, a single user can't vote more than once (KBItem) 

### LINKED ISSUE #638: You can upvote/downvote kb articles as many times as you want

shashwat1002:
fixes #638 , by adding a new field to the KBItem model (ManyToMany), to keep track of all users who have voted. The vote view checks whether a user has voted.
@gwasser , does this  work?


gwasser:
Thanks! This is much better.

So a question I have is that this patch seems to assume that only a registered user clicks the vote link. What happens when the user isn't logged in?

helpdesk/views/kb.py
```diff
@@ -46,9 +46,11 @@ def vote(request, item):
...
-        item.votes += 1
-        if vote == 'up':
-            item.recommendations += 1
-        item.save()
+        if request.user not in item.voted_by: 
+
+            item.votes += 1
+            if vote == 'up':
+                item.recommendations += 1
+            item.save()
```
timthelion: What happens if the user downvotes?

gwasser: Hmm I think it used to have a very similar behavior of recording down votes. Wonder if it was lost in a past refactoring? We'll have to add it back.

gwasser:
I'm merging this into the `develop` branch for the 0.3 release, I don't want to mess with the 0.2.x release too much anymore. Thanks!

Issue #622, opened 2018-06-15
Getting a AttributeError: 'NoneType' object has no attribute 'email' while doing get_email.


ddwaraka:
I have a cron job scheduled to do the get_email task to pull from the tickets inbox every 5 mins. That task recently failed to fetch tickets for a particular queue. When I manually ran the command, I got this error:

```
  File "helpdesk/management/commands/get_email.py", line 388, in ticket_from_message
    current_cc_users = [x.user.email for x in current_cc]
AttributeError: 'NoneType' object has no attribute 'email'
```

I printed out some of the values there and:
`current_cc_users = [u'<someone@gmail.com>']`
`current_cc = [<TicketCC: <someone@gmail.com> for Update on Order 12345678>]`

Am I doing something wrong or is this a bug?

gwasser:
This is probably some kind of logic error/bug. It looks like the `TicketCC` class does not necessarily have to have a non-null `user` property, the `_display` function checks for this and simply returns an email if there is no `user`. I'll add in a fix and release something new soon!

gwasser:
Actually are you using the latest `django-helpdesk` release? Based on the line number shown in your error, I don't think you're on the latest release since I see that around line 446 in the current `master` branch.

ddwaraka:
You're right, I am not using the latest release, my last update was easily a couple of releases ago.

gwasser:
OK, please let me know if the update fixes the issue.

gwasser:
I'm closing the ticket as part of the cleanups toward the 0.3 release but please reopen if this still a problem for you. Thanks!

tboulogne:
hello,

I'm using django 0.2.10 and getting :
````
  File "/home/tboulogne/.virtualenvs/apiprint/lib/python3.6/site-packages/helpdesk/management/commands/get_email.py", line 240, in process_queue
    ticket = ticket_from_message(message=full_message, queue=q, logger=logger)
  File "/home/tboulogne/.virtualenvs/apiprint/lib/python3.6/site-packages/helpdesk/management/commands/get_email.py", line 404, in ticket_from_message
    body = body.text
AttributeError: 'NoneType' object has no attribute 'text'
````

Any idea to fin please ?

gwasser:
Do you have an example email that causes this error message?

tboulogne:
Just send an html mail without anything into body, just object. I think you will reproduit this.

If not, i will backward my change and reproduit for you.

PR #661, opened 2018-11-01
Update get_email to prevent error on empty body message

tboulogne:
Pull request Regarding #622
Avoid the error when message body is empty

### DIFF
--- a/helpdesk/management/commands/get_email.py
+++ b/helpdesk/management/commands/get_email.py
@@ -401,10 +401,10 @@ def ticket_from_message(message, queue, logger):
         if ">" in mail.text:
             body = mail.find('body')
             try:
-              body = body.text
-              body = body.encode('ascii', errors='ignore')
+               body = body.text
+               body = body.encode('ascii', errors='ignore')
             except AttributeError:
-              body = ''
+               body = ''
         else:
             body = mail.text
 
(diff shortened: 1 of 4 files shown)


helpdesk/management/commands/get_email.py
```diff
@@ -400,8 +400,11 @@ def ticket_from_message(message, queue, logger):
         mail = BeautifulSoup(part.get_payload(), "lxml")
         if ">" in mail.text:
             body = mail.find('body')
-            body = body.text
-            body = body.encode('ascii', errors='ignore')
+            try:
+               body = body.text
+               body = body.encode('ascii', errors='ignore')
+            except AttributeError:
+               body = ''
```
timthelion: I think it would be better to default to `body = mail.text`.

tboulogne: In a first time i think like you and make `body = mail.text` but i got html code into the ticket.
so i set body to `''`.

timthelion: It seems to me like that's better than showing nothing at all.

gwasser: We could maybe use BeautifulSoup such as `mail.get_text()` to try to filter out any html present.

PR #666, opened 2018-11-07
Add celery task for get_email

timthelion:
Allows you to get_email even without cron, thus enabling the use of django-helpdesk of cronless PaaS services.


timthelion:
I'll admit that I'm too lazy to test this, but it should be enough...

PR #671, opened 2018-11-18
Forms' staff overriding

tboulogne:
Due form calling inside views into views/staff.py, it's difficult to override forms (for adding some special fields, create fieldsets, ....).
So it could be interesting to have a setting for this. I try this solution and works great.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@ docs/doctrees/*
 .project
 .pydevproject
 .directory
+.vscode
 
 # ignore demo attachments that user might have added
 helpdesk/attachments/
--- a/helpdesk/management/commands/get_email.py
+++ b/helpdesk/management/commands/get_email.py
@@ -401,10 +401,10 @@ def ticket_from_message(message, queue, logger):
         if ">" in mail.text:
             body = mail.find('body')
             try:
-              body = body.text
-              body = body.encode('ascii', errors='ignore')
+               body = body.text
+               body = body.encode('ascii', errors='ignore')
             except AttributeError:
-              body = ''
+               body = ''
         else:
             body = mail.text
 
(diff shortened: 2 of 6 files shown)


gwasser:
Thanks! I'm not sure if adding a bunch more settings is the best approach. Are the custom fields built-in not enough? What could we do to extend them to make them more useful to you?

tboulogne:
Hello thanks for your answer.

Add this setting help customize django-helpdesk design (add fieldsets to group fields, change field type, add custom styles on fields). Overriding as Django let's do with less effort.

Could see below :-).

image

Result with just the code below... and i add autocomplete fields into the form to find external data on custom fields.

```
class CustomTicketForm(TicketForm):
    if settings.HDL_SEARCH_FIELDS:
        for f in settings.HDL_SEARCH_FIELDS.keys():
            vars()[f] = AutoCompleteSelectField('{}'.format(f))

    if settings.HDL_FIELDSET_CREATE:
        field_order = set_order(settings.HDL_FIELDSET_CREATE)

    def __init__(self, *args, **kwargs):
        super(CustomTicketForm, self).__init__(*args, **kwargs)
        if settings.HDL_SEARCH_FIELDS:
            self.fields = set_custom_fields(
                self.fields, settings.HDL_SEARCH_FIELDS)

        if settings.HDL_FIELDSET_CREATE:
            self.fields = set_fieldset(
                self.fields, settings.HDL_FIELDSET_CREATE, False)

        if settings.HDL_FIELDSET_HIDDEN:
            self.fields = set_hidden_fields(
                self.fields, settings.HDL_FIELDSET_HIDDEN)
```

image


So thanks to consider a way to overriding form.

Regards

alligatorbait:
Moving logic out of staff.py, and probably a few other files that are part of the views, and into the models.py is the preferred approach. I have also found overriding items in staff.py is not possible because it is loaded in the view, loading the view when starting Django is a violation, where loading models and overriding is acceptable.

timthelion:
This is better implemented by transitioning staff.py to using class based views where form overriding is trivial.

PR #679, opened 2018-12-19
Email fix

timthelion:


timthelion:
There seems to be a bug in github ajax right now which makes it so that the wrong version of pull requests are submitted when changing the target branch :P

PR #680, opened 2018-12-19
Don't crash when mail body is empty

timthelion:


timthelion:
This is a continuation of @tboulogne 's PR https://github.com/django-helpdesk/django-helpdesk/pull/661

Issue #677, opened 2018-12-13
You should be able to change the assignee of a ticket in the top section of the staff view ticket view


timthelion:
Currently, you can do this by pressing the "Change further details button" 

![untitled](https://user-images.githubusercontent.com/1391608/49953643-dc8a4e80-fefe-11e8-85da-54c1d05ad436.png)

But you cannot do so in the top box.

![1](https://user-images.githubusercontent.com/1391608/49953774-25420780-feff-11e8-85f3-43447a76ebec.png)

I would fix this by adding a pencil icon https://fontawesome.com/icons/user-edit?style=solid which linked to an anchor which would be inside the "Change further details button" section.

timthelion:
I'd be happy to do this, but I'm a bit hesitant to work on the UI until #635 hits.

gwasser:
I've thought about that change too. Sorry, I have some time off coming up soon and I would like to finish the UI so I'm not holding up further work :-)

DavidVadnais:
The implementation of this is slightly different then mentioned here but we now have a drop down in the ticket_desc_table that allows users to select a staff member and save them as the owner of ticket.  I would say this issue is solved.

PR #705, opened 2019-02-02
Add pencil buttons to ticket view for easier assignment and priority …

timthelion:
…setting

assign

Right now the buttons just jump down to the bottom section, which isn't very modern (modern web design would make the field transform into an editable one) but I actually find it to be good, because then you're down right next to the submit button when preforming a quick change.


timthelion:
Fixes  https://github.com/django-helpdesk/django-helpdesk/issues/677 I guess.

PR #706, opened 2019-02-03
Display only queues for the logged-in user

jachym:
In the submission form (public and staff), display only queues, which should be available for the logged-in user only.

### DIFF
--- a/helpdesk/tests/test_ticket_submission.py
+++ b/helpdesk/tests/test_ticket_submission.py
@@ -70,6 +70,8 @@ class TicketBasicsTestCase(TestCase):
         self.assertEqual(email_count + 3, len(mail.outbox))
 
     def test_create_ticket_private(self):
+        from helpdesk import settings as helpdesk_settings
+        helpdesk_settings.HELPDESK_ENABLE_PER_QUEUE_STAFF_PERMISSION = True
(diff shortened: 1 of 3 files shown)


jachym:
There is little change in the behaviour while creating ticket via e-mail.

Not sure, why Python 2 refuses :-/

jachym:
@gwasser I'm moving to `development` branch. One issue I have: the logic moved to `forms.py`, I need to `__init__` method of `TicketForm` class

My problem is, that I create list of queues based on logged-in user. Not sure, how to get logged-in user in the `__init__` method (in `master` in `views/staff.py` this is part of the `request` object). Any hint?

gwasser:
Hmm offhand not sure, will have to think about it. Possibly the right place is to look at where the form is generated in `views.py`, or maybe adapt the template that it is rendering.

jachym:
See pull request for develop branch in pr #715

gwasser:
Thank you! I merged #715. I hesitate to merge this change into the 0.2 release if it doesn't work well on both Python 2 and 3. So I'm going to suggest let's focus on getting 0.3 done :-)

jachym:
This was addressed in the develop branch - closing.

PR #707, opened 2019-02-03
AttributeError: 'Command' object has no attribute 'option_list'

JayTurnr:
There is a bug in the command in which option_list doesn't exist and thus, cannot be added to.


JayTurnr:
Thank you! I plan on fixing the ticket interface on mobile at some point. Unless someone else is already working on that?

gwasser:
We've switched to an entirely bootstrap-4-based UI for the next release (version 0.3) that should be more mobile friendly. You're welcome to jump into the `develop` branch to help finishing testing so we can release it!

PR #715, opened 2019-02-05
User queues develop

jachym:
Queues only visible to user with corresponding rights

### DIFF
--- a/helpdesk/forms.py
+++ b/helpdesk/forms.py
@@ -275,8 +275,11 @@ class TicketForm(AbstractTicketForm):
         """
         Add any custom fields that are defined to the form.
         """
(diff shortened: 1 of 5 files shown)


gwasser:
Thank you for your work on this!

jachym:
not sure, if you like it @gwasser 

I believe, that usage of the `get_form_kwargs` method is the most elegant way, how to pass data to the form and could be used for further adjustments (e.g. list of users, you can assign the ticket to etc..)

Not sure, if the behaviour of `HELPDESK_ENABLE_PER_QUEUE_STAFF_PERMISSION` switch works as expected.

gwasser:
Yes I think it's fine, thanks. We'll run it through some more testing!

Issue #700, opened 2019-01-30
develop branch: If an emails body is empty but there is an attachment, then the attachment is included as text


timthelion:
Related: https://github.com/django-helpdesk/django-helpdesk/pull/680

![no-body](https://user-images.githubusercontent.com/1391608/51988848-fcbd9f00-24a5-11e9-99b4-f0b8a8c8e968.png)

johanovic:
I've been working on this bug for a while without success. It is hard to replicate. My guess is that it might have to do with "application/ms-tnef" maintype's of the message.

But you might actually be right and this might have to do with any messages with attachment, but no body. Testing it right now.

johanovic:
Just tested it and the problem indeed occurs when messages without body and with attachment are imported. 

Bug confirmed.

gwasser:
Thanks for testing @johanovic! Could you please add your test email to our unit testing framework so we can make sure we properly handle this situation in the future? Email is notoriously hard to get right; the email processing is probably an older part of the code base and can probably use some clean up to make sure we're using the latest features of python email libraries, etc.

timthelion:
@johanovic Is this happening on master as well, or only develop?

johanovic:
I wish I had a proper test, but I've mostly been testing it step by step. 

I have tried to make the changes of https://github.com/django-helpdesk/django-helpdesk/commit/eee82015f787c303acdec85d667a58296093cac2#diff-01c02c5839198cb2c27ec0ef4dc09a16

but sadly they do not solve the issue.

timthelion:
Here is a failing test case: https://github.com/django-helpdesk/django-helpdesk/pull/720

gwasser:
Thanks @timthelion! I just pushed a potential fix. It looks like we were using the last known part of the mime message for processing rather than the message as a whole, so I updated it to use the full message and keep the description empty unless an HTML part is discovered.

johanovic:
I tried this in production and I confirm that it works for emails without a body and without as subject, but with an attachment. Thanks!

gwasser:
Thanks for testing @johanovic! Glad to have production confirmation of the fix :-)

PR #720, opened 2019-02-15
Add failing test case for issue-700 emails without bodies

timthelion:


gwasser:
Thanks! Looks like a good test. Will talk about a fix in #700

PR #722, opened 2019-02-21
Error not raised when user change the HELPDESK_PUBLIC_TICKET_QUEUE in…

altimore:
… the admin.

Please do not return blank HttpResponse(status=500) it is a nightmare to locate/debug when a user change the list name in the admin.

Either let it raise its exception or give it a default value to display the form.

I separated the slug so it displays its values in the logger.

### DIFF
--- a/helpdesk/views/public.py
+++ b/helpdesk/views/public.py
@@ -67,11 +67,8 @@ def homepage(request):
 
         # add pre-defined data for public ticket
         if hasattr(settings, 'HELPDESK_PUBLIC_TICKET_QUEUE'):
-            # get the requested queue; return an error if queue not found
-            try:
-                queue = Queue.objects.get(slug=settings.HELPDESK_PUBLIC_TICKET_QUEUE)
-            except Queue.DoesNotExist:
-                return HttpResponse(status=500)
+            slug=settings.HELPDESK_PUBLIC_TICKET_QUEUE
+            queue = Queue.objects.get(slug=slug)
         if hasattr(settings, 'HELPDESK_PUBLIC_TICKET_PRIORITY'):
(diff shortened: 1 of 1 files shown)


altimore:
Hello @gwasser,  I don't know how to display this ticket in the Issue list, did you see it ?

gwasser:
Hi! I did see it and it apparently got lost under tickets, sorry about that. It looks like you just removed the 500 error, which is fine, but I think we still need to handle the "queue does not exist" situation somehow. Do you have any ideas?

altimore:
Hello, 

Thanks for you fast answer. 

On my side the 500 error is handled by Django itself. 
The exception is Raised normally and handled by the logging sub-system (text, db, sentry, etc...) while the user is having and standard 500 page displayed in production or displaying the error in debug mode.

I didn't try to run your project as a standalone but i guess setting up the 500 page should do the trick : https://docs.djangoproject.com/en/2.2/topics/http/views/#customizing-error-views

By default Dango looks for the template ./500.html in the template root directory, I can try to create one on the demo project and test it after work tonight.

In any case we cannot return a blank HttpResponse instead of an exception.

If you really want to customize the exception message you can raise a custom one but a object.DoesNotExist is usually clear enough.

altimore:
It works with just the 500.html.

I made a branch with the template added and the settings to recreate the error, then you run the server and open the dashboard in your browser.
https://github.com/altimore/django-helpdesk/commit/e8201e512927a50d99fa30725ab418d6525d7ee6

You can add the 500.html in your helpdesk app, in your demo app or in any project your use helpdesk with. The order of the apps in settings will determine which template is used if you got multiple 500.html files.

Issue #732, opened 2019-03-08
Develop branch: Emails with 8bit transfer-encoding and utf8 charset don't work


timthelion:
Getting pretty glad I did that refactor to make it easy to add broken test messages ;)

Failing test coming right up.

gwasser:
Thanks for the test case! Will investigate.

gwasser:
Appears to be fixed in PR #747. Thanks @timthelion !

PR #733, opened 2019-03-08
Add failing test for #732

timthelion:
Failing test for #732


gwasser:
Thanks! even though the test fails, I'm going to merge it so we can work on it on the `develop` branch.

PR #738, opened 2019-03-10
adding some boostrap styles

jachym:
I'm certainly not web designer. Still, even django-helpdesk seems to be working pretty well sofar, the design could be done much better.

This set of changes introduces some slight bootstrap modifications, which should make the design of some elements slightly "cleaner". 

Overall: buttons are smaller, tables are smaller. Headers set where I thought, it makes sense.

Tables are slightly restructure and it should be more easy to get oriented at the first sight.

I hope, you like it, comments are of course welcomed - I assume bigger discussion around this pull request.


jachym:
tests are ok, except for the one, broken by #733

gwasser:
Thanks!

PR #742, opened 2019-03-14
Fix command rundemo

CedricCarrard:
We get an error if we run the command  ```sudo make rundemo```

Here is the error in the image : 

error

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -83,8 +83,8 @@ readme:
 #: demo - Setup demo project using Python3.
 .PHONY: demo
 demo:
-	$(PIP) install -e .
-	$(PIP) install -e demo
+	$(PIP) install -e . --user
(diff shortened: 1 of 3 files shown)


gwasser:
Using `--user` is probably better than a system-wide install.

Looks like your error message is because `pip3` isn't found. What linux distribution do you use? I had thought pretty much all of them used pip3 to go with python3.

CedricCarrard:
Indeed I think it is always better to use ```--user```.

I use Ubuntu 18.10 and I also use ```pip3 ``` the command works without the  ```sudo make rundemo``` but does not work with the sudo

gwasser:
Hmm interesting, Ubuntu must be doing something odd with its sudo command. In any case the `--user` solution is probably preferred. Thanks!

PR #749, opened 2019-04-01
fix markdown bug

jachym:
* Typo introduced by recent commit
* Add markdown also to Ticket resolution field


jachym:
there is not much to be said about the tests :-\

gwasser:
For some reason it doesn't seem to have actually run the tests automatically.

PR #764, opened 2019-07-04
Add setting for HTTPS in email links

floatingpurr:
A quick commit in order to enable links with HTTPS in emails.


gwasser:
Thanks! Simple patch, makes sense to me.

PR #765, opened 2019-07-06
More gui updates

jachym:
This pull request:

1. updates font awesome
2. fixes and cleans some time formatting 
3. redesigns slightly ticket detail view, inspired by redmine
3. redesigns slightly ticket filter view, inspired by redmine

Comments:

I was wondering why django-helpdesk ticket view and filter view are so un-arranged and why they take so much space, yet the orientation at first sight is still not the best. At the end it came to my mind, that Redmine has clean, nice and tidy views on both - ticket detail as well as ticket table with filters view. I tried to put something similar together - without too deep changes in the templates and I hope, you find the proposal nicer than current view.

For comparing few screenshots:

Ticket table with filters view current `develop`
obrazek

Ticket table with filters view `gui_updates`
obrazek

Ticket detail view current `develop`
obrazek

Ticket detail view `gui_updates`
obrazek


jachym:
Tests failed, but I believe, this is due to previously introduced features and has nothing in common with the gui update.

gwasser:
This looks great, thank you! I especially like the filters redesign. The original design goes back to the 0.1.x series and so didn't look quite as nice with the new theme. I'm not a great graphic designer :-) I agree that there were already some test failures I need to fix up, I don't think your redesign affects the tests.

PR #769, opened 2019-07-31
Search custom field values.

willstott101:
I would be interested to see if anyone has any performance issues with this PR.

Tests and documentation updates are probably needed before merging.


gwasser:
Thanks! Wow that seems simple. I agree we should add some tests and see if we can spot a performance problem, but it's probably ok.

willstott101:
Yeah, gotta love the Django ORM!

PR #780, opened 2019-10-18
Remove hardcoded pk from test suit

timthelion:


timthelion:
Don't merge until the test suit passes

PR #784, opened 2019-11-01
Add OR syntax to search

timthelion:


gwasser:
Seems reasonable to me thanks! I do wonder if this query could end up being very slow if there's lots of ORs, but I think it's fine for now unless someone complains.

PR #790, opened 2019-12-17
fix<migrations>: correct import of b64decode in migration 0020

koriaf:
I didn't cover it with tests and didn't even run them (sorry), but it was broken for me and now it's not, so might be helpful.


koriaf:
test fail seem to be unrelated to the changes.

gwasser:
Yes there is one failing test that I don't quite understand why, it used to work until not long ago. Maybe a change in python and/or django?

Thank you for this patch! The `master` branch targets Django 2.0+ and so is now exclusively Python 3, no need for the Py2 compatibility in future patches.

PR #791, opened 2019-12-25
fix error format time_spent

stilet:
fixed time spent display in tickets

### DIFF
--- a/helpdesk/lib.py
+++ b/helpdesk/lib.py
@@ -161,9 +161,9 @@ def format_time_spent(time_spent):
     """
 
     if time_spent:
-        time_spent = "{0:02d}h:{0:02d}m".format(
-            int(time_spent.total_seconds() // 3600),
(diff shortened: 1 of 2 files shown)


helpdesk/lib.py
```diff
@@ -161,9 +161,9 @@ def format_time_spent(time_spent):
     """
 
     if time_spent:
-        time_spent = "{0:02d}h:{0:02d}m".format(
-            int(time_spent.total_seconds() // 3600),
-            int((time_spent.total_seconds() % 3600) / 60)
+        time_spent = "{0:02d}h:{1:02d}m".format(
+            time_spent.seconds // 3600,
+            time_spent.seconds // 60
```
krylataya: % is missing here. So it displays full number of minutes without substraction hours (e.g. 02h:140m instead of 02h:20m)

PR #801, opened 2020-02-06
fix<public>: Public ticket creation form improvements

koriaf:
Update forms.py and views/public.py to physically delete form fields
instead of just hiding them if default values for queue, date and priority are provided.

+ some bugfixes
+ code and stale imports removal
+ fix pep8 errors

The original issue I was facing was being unable to create public ticket when default queue is not configured correctly (either not allowing public submissions or having different slug). Now there are warnings (going to sentry if configured) and 500 errors so it can't be forgotten setting the new setup for users. And some security improvement by not accepting data from the user when we don't want to do so.

### DIFF
--- a/helpdesk/forms.py
+++ b/helpdesk/forms.py
@@ -6,11 +6,10 @@ django-helpdesk - A Django powered ticket tracker for small enterprise.
 forms.py - Definitions of newforms-based forms for creating and maintaining
            tickets.
 """
-
+import logging
 
 from django.core.exceptions import ObjectDoesNotExist
 from django import forms
-from django.forms import widgets
 from django.conf import settings
 from django.utils.translation import ugettext_lazy as _
 from django.contrib.auth import get_user_model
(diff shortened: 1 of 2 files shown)


koriaf:
There is one side-effect of that change: any page where anon ticket form is displayed will be raising 500 error for anon users until it's configured correctly (there is the queue with public tickets creation enabled and this queue is equal to the one defined in settings). 
Applicable only for cases when HELPDESK_PUBLIC_TICKET_QUEUE is present and not empty.

So this approach helps the person configuring the setup not to forget to configure queue correctly as well. Otherwise this may go unnoticed and form is shown fine but unable to be sent. It's a political decision if such workflow should be applied or not.

gwasser:
I think that sort of change is fine, but we should definitely update the documentation if so to clearly explain the change and how to properly configure.

helpdesk/forms.py
```diff
@@ -364,15 +375,42 @@ def __init__(self, hidden_fields=(), readonly_fields=(), *args, **kwargs):
             'due_date': 'HELPDESK_PUBLIC_TICKET_DUE_DATE',
         }
 
-        for field in self.fields.keys():
-            setting = field_hide_table.get(field, None)
-            if (setting and hasattr(settings, setting)) or field in hidden_fields:
-                self.fields[field].widget = forms.HiddenInput()
-            if field in readonly_fields:
```
timthelion, line 371: What happened to disabling read-only fields?

PR #803, opened 2020-02-11
Teams

timthelion:
This allows you to assign KBItems to teams. Tickets are only displayed on the dashboard if they are either unassigned/associated with any KBItem or they are assigned to a KBItem which one of your teams is responsible for.


gwasser:
Thank you for your work on this! I do think whenever we add dependencies, we should step back and consider if (1) the dependency is really necessary, and (2) if it is the _right_ dependency for the job (for example, is there a different equivalent library that may be better for some reason?).

I don't know much about pinax so that's something we'd have to confirm. A quick look at the pinax-teams repo suggests that there hasn't been much work for over 6 months, and the README suggests that it may not have been fully ported to Djangos above 2.0 (maybe not 2.1/2.2, definitely not 3.0, etc.). I'd hate to add a dependency that then holds `django-helpdesk` back from migrating to Django 3.0, for example.

What do you think? Is pinax required? Can we implement teams without using pinax?

timthelion:
@gwasser It seems to me that the best solution is 2. I don't see a good reason to write duplicate code just to get rid of a dependency. In the worst case, we can copy the repo and maintain it ourselves. That should be no worse than writing the code from scratch.

gwasser:
I mostly actually wanted to check that there wasn't some other more well
tested and well maintained library for teams management before we settled
on pinax. But if pinax is the best option for now then sure.

timthelion:
OK, I *think* we've tested this enough that it can be merged now.

phoebebright:
I'm looking at turning off this teams dependancy for the reasons in the first comment.  Pinax teams will require changes to move to django 3 as will all it's dependancies.  While teams is a nice to have, I will try to build a version where it can be built without teams.  Any suggestions welcome!

gwasser:
Python 2 is now end-of-life as of January, so the next major release is going to target Python 3 only anyway. I definitely recommend starting any new projects with Python 3 if at all possible.

This doesn't affect the current 0.2.x release if you'd prefer to use that (it still supports Python 2 and does not depend on pinax)

PR #805, opened 2020-03-03
Fill followup user field if authenticated user submits ticket

nitr0man:
Currently follow up field is not filled even if authenticated user submits ticket. So it's impossible to generate user's ticket list.


gwasser:
Thank you! Could you double check that the Python 2.x tests pass too?

gwasser:
The issue breaking the Python 2 code looks like it was fixed in #813.

PR #806, opened 2020-03-12
Optionally hide Saved Queries from Frontend

morlandi:
At the beginning, you have to manage a few tickets, and need to familiarise with the workflow.
In that context, Saved Queries are useless, and rather intimidating


timthelion:
Hi,
we're trying to reduce the number of configuration options that can be put in settings, not increase them. In this case, such an option is not necessary. You can accomplish this by creating new named blocks in the template and over-riding those blocks in your own new versions of the templates.

morlandi:
Thank you for feedback; I understand you point ;)

PR #830, opened 2020-06-01
Feature/dashboard pagination

Bukowskaii:
simple change to the dashboard view and html files to allow pagination on the ticket tables generated by tickets.html.

staff.py

- read user ticket pp setting (used in the ticketing view), lets just recycle this instead of making a separate setting (or settings) per table. (line 96-97)

- expect new http get args for each table for user submitted tickets (atrbcu_page), user assigned tickets (ut_page), and user resolved tickets (utcr_page). If not passed, default to 1.  (line 99-102)

- use django core paginator functionality to split the ticket queryset(s) into pages, and pass only the requested page to dashboard.html (line 154-187)

helpdesk/templates/helpdesk/dashboard.html

- pass the correct http get arg to the include for three tables (lines 19-20, 25-26, 32-33)

helpdesk/templates/helpdesk/include/tickets.html

- add pagination controls for each table if it has more than the current page

Results:

No pagination required looks like this
image

W/ pagination required will look like this
image

### DIFF
--- a/helpdesk/templates/helpdesk/dashboard.html
+++ b/helpdesk/templates/helpdesk/dashboard.html
@@ -16,18 +16,21 @@
 
 {% if all_tickets_reported_by_current_user %}
 {% trans "All Tickets submitted by you" as ticket_list_caption %}
-{% include 'helpdesk/include/tickets.html' with ticket_list=all_tickets_reported_by_current_user ticket_list_empty_message="" %}
+{% trans "atrbcu_page" as page_var}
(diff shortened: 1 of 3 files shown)


Bukowskaii:
Need to better create pagination html, currently will create a object for every page which gets out of control really quickly

 
image

Bukowskaii:
okay thats nicer

image

gwasser:
This looks great, thank you!

PR #831, opened 2020-06-03
Fix html format error in french email templates fixtures

Benbb96:
When using french templates, there were missing parts.

After investigating, I found that the fixtures had an error with html closing tags.

`>\"` --> `\">`


Benbb96:
Ok, I created a new PR #833 !

gwasser:
I'm going to close this one since merged PR #833, thanks! This will get merged into master eventually when we make a new bugfix release.

PR #836, opened 2020-06-12
A ticket cannot depends on itself or on a ticket already depending on it

Benbb96:
I added a fix in the ticket dependency form in order to prevent creating a dependency on itself or with a ticket which is already depending on the current ticket, since it would block the process (though we can always delete the dependency after).


gwasser:
Excellent, thanks for catching this bug!

PR #837, opened 2020-06-19
Fix javascript buttons for bulk select all/none/inverse in ticket_list

Benbb96:
Javascript buttons for bulk select weren't working properly with `.attr()` function.


gwasser:
Thanks!

PR #838, opened 2020-06-29
Fix Knowledge Base item score

Benbb96:
Display the overall rating as a float out of 10.


Benbb96:
I pushed too many files so I close in favor of #839 which is the good PR.

PR #840, opened 2020-07-07
Fix Keyword Filter Box in Ticket list

Benbb96:
The _Keyword_ filter box didn't show up even if it is in saved query because a variable `query `was used whereas it wasn't available in the context.


gwasser:
Thank you!

PR #844, opened 2020-07-20
fix<process_email>: free the log file handler after it's not used anymore

koriaf:
it's questionable to have file handler at all given some people will be running it in either long-living Docker environment with root file system read-only or very short-living with logs colected by the container manager, but anyway


gwasser:
Thanks!

gwasser:
I backported this to the 0.2.x branch.

PR #845, opened 2020-07-20
Demo-related bugfixes

koriaf:
See the commits for detailed descriptions. I'm not 100% sure about --user flag here because this fix just works for me with recent pip/python and virtualenv - but may cause more problems for people not using either docker or virtualenv. But they probably know that they are doing.

also such cruel Pinax upgrade without any testing doesn't look good as well, but, given it wasn't working anyway...

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -78,8 +78,9 @@ readme:
 #: demo - Setup demo project using Python3.
 .PHONY: demo
 demo:
-	$(PIP) install -e . --user
-	$(PIP) install -e demo --user
+	# running it with and without --user flag because it started to be problematic for some setups
+	$(PIP) install -e . --user || $(PIP) install -e .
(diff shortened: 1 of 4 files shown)


gwasser:
Thanks! Yeah I haven't had a chance to work on master in a while; someone else had contributed the pinax code so it's not well tested yet.

koriaf:
btw this Pinax code is introducing interesting problem - it marries the helpdesk library to the pinax account library, and me for example using django-allauth and they are conflicting; means helpdesk library is not a simple library to integrate to existing projects anymore but more like a standalone application

gwasser:
Yeah there's definitely a design tension here. I think the codebase is moving more toward standalone rather than add-on to other projects. I'm also not super stuck on pinax if there's an alternative; we had one person propose it for doing better auth management, but there could be a better idea somewhere if you know any. Thanks!

PR #859, opened 2020-08-18
Html custom field

chris-hellberg:
Adds a new custom field type called HTML multi-line. When editing or creating a ticket that has a custom field of this type, the field includes a javascript window that is a WSYWIG HTML editor. When the ticket is saved or updated, the basic HTML is saved to the database (without any headers  or embellishments). When displaying the ticket, the HTML rendering is shown on the ticket page.

The {{ form.media }} modification to the base.html file is needed to call any HTML rendering.

Finally the ticket_desc_table.html modifications shift from ifequal to the more updated if xx == yy format as part of the other changes to the file.

The following changes need to be made to helpdesk's configuration/environment to enable this functionality:
1. Add: `path('tinymce/', include('tinymce.urls')),` to urlpatterns in urls.py
2. Install the package: `django-tinymce4-lite`

Open to ideas whether this functionality is made optional by default to avoid installing yet another package.


timthelion:
It should be optional I think...

hellcheese:
That's fine by me. The next question is, where's the optimal place to have a single flag/option to enable/disable.

gwasser:
Since it does introduce a new dependency (TinyMCE) we should think about it a little if we want that the default or not. Honestly we should probably be using TinyMCE in all the fields regular users interact with but that's a much bigger project than just this PR, but I bring it up because we might want this option by default if it is indeed the long term goal.

Benbb96:
I agree with @gwasser, it would be great if all fields like the `description` and `resolution` of the ticket, as well as the follow up `comment` and the preset reply `body`, were `HTMLField`, using the TinyMCE widget. The sent mail would then include HTML part and that would be nice for formatting.

I have already implemented that on [my fork](https://github.com/Benbb96/django-helpdesk/blob/310104835e1bb21b47582032e7dc39e1632ee4da/helpdesk/models.py#L512). Maybe I should try to open a PR ?

gwasser:
Yes definitely, PRs welcome!

hellcheese:
@Benbb96 The tricky bit may be how to migrate an existing non-HTML DB column to one of type HTMLfield

uhurusurfa:
Most decent HTML editors support being activated in a form field based on a CSS selector. You just have to include the relevant WYSIWIG editor library and optionally some configuration override and the field will be targeted based on the CSS selector.
Apart from TinyMCE noted in the comments above, there are others like Summernote, Quill, CKEditor etc and each has its own benefits so allowing the choice of HTML editor should be left up to the implementation.
The intiialisiation code to be included in the web page can be configurably injected based on Django settings.
So defining a standard CSS selector for Helpdesk that can be added to fields that are known to benefit from HTML would mean that the end user can choose the editor of their liking and activate the HTML editor by adding the appropriate startup code into the Django settings for their app.

gwasser:
I'd be open to keeping it more open-ended on support @uhurusurfa . I do wonder if standardizing one might help with consistency of visual style/behavior? But also not stuck on it.

Mte90:
Wondering if it is better a markdown (it is already support) editor maybe something like https://github.com/agusmakmun/django-markdown-editor

As the system is also open by public or by email. 
This pr add only the html field but there isn't any sanitization so people can inject javascript and I don't think that is something we want.

gwasser:
A Markdown editor would make a lot of sense too. I think Markdown can be converted to HTML fairly easily/directly, there's got to be library functions that do that, so could allow better sanitzation as well as HTML email support.

Mte90:
So to move on the discussion maybe it's the case to close this PR, discuss what we want to do and how (markdown, sanitization, probably parameters to define what HTML tags to strip, what libraries etc).

timthelion:
This PR only implements this for custom fields and does not pertain to follow up bodies. It seems like there is more discussion about how best to implement this. In any case, I don't think that the code as it is will be merged or is usable with the current helpdesk version.

Issue #866, opened 2020-08-25
Validate ticket.assigned_to is set before attempting comparison in staff.py


alligatorbait:
**Describe the bug**
Non-staff and Non-superuser attempts to access a ticket that is not assigned / has no Owner.
```
File "lib/python3.6/site-packages/helpdesk/views/staff.py", line 82, in _is_my_ticket
    if user.is_superuser or user.is_staff or user.id == ticket.assigned_to.id:
AttributeError: 'NoneType' object has no attribute 'id'
```

**To Reproduce**
See bug description.

**Expected behavior**
PermissionDenied being Raised is currently the expected result.

 - django-helpdesk Version 0.2.22

**Additional context**
Current Code: staff.py:79
```
def _is_my_ticket(user, ticket):
    """Check to see if the user has permission to access
    a ticket. If not then deny access."""
    if user.is_superuser or user.is_staff or user.id == ticket.assigned_to.id:
        return True
    else:
        return False
```

Suggested Code checks if ticket.assigned_to is not None before attempting to access ticket.assigned_to.id: staff.py:79
```
def _is_my_ticket(user, ticket):
    """Check to see if the user has permission to access
    a ticket. If not then deny access."""
    if user.is_superuser or user.is_staff or (ticket.assigned_to and user.id == ticket.assigned_to.id):
        return True
    else:
        return False
```

PR #869, opened 2020-08-25
Verify ticket.assigned_to is set before attempting to access id property

alligatorbait:
Validate ticket.assigned_to is set before attempting comparison in staff.py #866


gwasser:
Thanks, I'll apply this to `master` too.

PR #879, opened 2020-10-08
Update configuration.rst

typonaut:
Adds documentation as to the function of "E-Mail Check Interval" and its interaction with a crontab.


gwasser:
I do not understand why codecov is saying coverage is down when this just changes documentation? Weird. I'll merge it and apply the patch to `master` too, thanks!

Issue #878, opened 2020-10-05
Contribute to django-helpdesk 0.3x


alligatorbait:
Discussion to help work the 0.3x release and progress this project towards compatibility with the current Django version.

There are tasks to complete the 0.3x release, the purpose of this ticket is to have a place to discuss current goals and help newcomers know where they can put effort and pull requests to move support the next version release.

Would one of the main contributors please post some primary items that need attention that I might be able to work on?

Thanks.

gwasser:
Thank you very much @alligatorbait ! I haven't had the time to work on this like I had hoped.

The `master` branch is the 0.3.x work. As far as I know it mostly needs to be updated for Django 3.x. There's some kind of bug in it that I possibly introduced when merging a bugfix from 0.2.x, but I haven't gone through to figure out where the error is.

You can also check the tickets under the 0.3.0 milestone. It's possible some of them should be deferred to 0.4.x, we need someone to check them and triage them and figure out only the minimum needed to get the 0.3.x release out.

Thanks!

gwasser:
I didn't look too deeply, but it looks like many of the "bug" tickets on the 0.3.x milestone are related to Django 3.x support so I think that would be the top priority for the next release. Thanks again for your help!

Benbb96:
Hi all !

I believe the first bug to fix is the import of an old function which doesn't exist anymore in master : `apply_query` (in `staff.py` line 30). Each time I run on master, I'm having trouble with that error. I can try to understand the new `__Query__` class and fix it in a PR.

At the same time, I will take a look into the things that are missing in order to make **django-helpdesk** running with the third version of Django. But as I'm not using it my other projects, I'm not sure I will manage to do it..

Then I would like to work on replacing some fields (ticket `description` and `resolution`, followup `comment` and preset-reply `body`) into HTML Fields with the use of the **TinyMCE** library in order to have better formatting and also keep the HTML from email (instead of having an attachment with _email_html_body_). See #859 for reference.

Another thing I would like to add is a **Merging** option on the ticket list : select some tickets and then, on an intermediate page, choose which one will receive all the followups of the other ones, also select which information for each ticket's fields. If an email is sent to a merged ticket, the followup has to be attached to the ticket to which it was merged.

[...]

PR #881, opened 2020-10-08
Fix errors on master

Benbb96:
As part of my contribution for #878, here is my PR to fix errors I had on master.  
I've taken the opportunity to clean things up and remove useless imports (read commit history for more details).

I had to set an absolute path for `ROOT_URLCONF` and `WSGI_APPLICATION`in order to run my local server. I don't know if that could break things for others or the tests (which I still don't know how to run correctly).


gwasser:
Tests are run with `python quicktest.py helpdesk`. It's a pretty old testing harness so I'd fully support modernizing it if feasible (although maybe we'll push that to the 0.4.x release simply to not hold up the 0.3.x any longer).

gwasser:
There's a few interesting errors in the travis-ci testing here. Some of the template tags are crashing for not being "iterable". Other than that looks like maybe the `get_email.py` script doesn't have the right imports or something.

gwasser:
I merged the bugfixes from the not yet released 0.2.23 bug fix release into `master`, please do a pull and check that your PR syncs with latest `master`! sorry if this caused you some trouble.

Benbb96:
No problem, I think I'll do it today.

PR #882, opened 2020-10-10
Missing Migrations for 0.2.22

TurnrDev:
Fixes a missing migrations messages in 0.2.22 which is caused by the careless changing of strings in models


gwasser:
Thanks!

PR #883, opened 2020-10-12
Updates required to work with Django 3.x

alligatorbait:
Updates for a working version with Django 3.x


alligatorbait:
This is to contribute on #878 getting the project up to the level to work with Django 3.x.

PR #884, opened 2020-10-13
update saved_queries imports, filter, and authentication checks, correct created field reference and remove Django version decrement

alligatorbait:
#878 Updates to correct some test failures in saved_queries and remove Django version decrement.


gwasser:
Thanks! That seems to have fixed the crashes, just need to update the tests now.

Issue #864, opened 2020-08-23
Way to integrate with sites base template


LenCole:
Would like a way (or if there is I can't find it) to integrate with site so that the sites base.html template is used (to provide sites - header, footer, etc.)

alligatorbait:
> Would like a way (or if there is I can't find it) to integrate with site so that the sites base.html template is used (to provide sites - header, footer, etc.)

I second this one. Would be very nice to have a way to create a directory in Django with templates that wrap the django-helpdesk templates. This is something I have done with djangorestframework.

gwasser:
Thank you for the suggestion. The templates currently in use are pretty old and definitely need some refactoring for the next major release.

alligatorbait:
I looked over the templates and have an easy way to add an override for a navigation header above the django-helpdesk navigation. I will put in a pull request on master soon.

timthelion:
This is fixed right?

PR #886, opened 2020-10-14
customizable navigation header template for override and documentation update

alligatorbait:
#878 contribution and applies to #864 (although not exactly as requested applies in principle to intent). This provides a blank html template to override and create a custom header above navigation.


gwasser:
Makes sense to me, simple solution. Thanks!

PR #887, opened 2020-10-14
Update Django Versions in documentation and correct 2 referenced before set bugs

alligatorbait:
One documentation change and two minor bug fixes.


gwasser (approved):
Looks good, thanks!

PR #895, opened 2020-10-30
New feature:  Merging tickets

Benbb96:
This new feature let staff users merge tickets from the ticket list after selecting tickets and choose option **Merge** from the mass update list select : 

image

One thing to note is that I limited the merge feature up to **four** tickets at the same time but it could be more if needed.
Also, tickets from different queues cannot be merge since it could result in some conflicts for permission access.

An intermediate page shows a table to the user with all information for each attributes and custom fields for each tickets : 

image

User has to choose which information to keep and also select the main ticket, in which everything will be gathered. All follow-ups from other tickets will be moved to the main ticket (with a new title to understand where the followup comes from) : 

image

Same applies for CC. Submitter email and assigned users on other tickets will be added to main ticket's CC if not already present.

Once a ticket has been merged to another one, it set as DUPLICATE and it is not possible to add another follow-up (even for staff users). A message display in which ticket it has been merged to : 

image

If somebody send an email to a merged ticket, the followup will be transfer on the ticket it has been merged to.

---

I added one test for the merge view but it could be a good thing to also test if follow-up is well added to the main ticket when a mail is sent on a merged ticket.

I also updated french translations.

If you have time to review/test this, I know it is a long PR but I think it will be a useful feature for everybody using this library 😃


gwasser:
Awesome! We've long needed the ability to merge tickets, thanks for starting this. Let me know when you're ready for a review.

Benbb96:
I think it's all good for me, @gwasser !
@alligatorbait, if you have some time to take a look at it too 😉

alligatorbait:
I have been offline for over a week, won't be able to look for another few
days.

gwasser:
Looks great! I'll merge this so we can include it as part of the beta release for testing. Thanks!

gwasser (approved):

alligatorbait:
I added a comment in the code review. There is an else block that looks to be indented a level too deep.

gwasser:
It's part of a try-catch-else statement; not a best practice to use exception handling for control flow, but I think it's ok for now.

Benbb96:
@gwasser Oh I didn't know it was a bad practice to use exception handling. How should I have done it then ?

gwasser:
Exception handling can result in some slower code and a bit more difficult to follow the code on what's happening. Using a normal if-then and object checking is the preferred method, but like I said, not a huge deal here.

Benbb96:
Alright, I'll know it for later. Thanks for your answer !

timthelion:
In this case exception handling is used correctly. if-then object checking is wrong here, because the DB is concurrent and checking-then getting would lead to a race condition. Checking then getting also leads to two round trips, which is much slower. Generally, in python, @Benbb96's pattern is a good one:

https://devblogs.microsoft.com/python/idiomatic-python-eafp-versus-lbyl/

Benbb96:
Thanks for the article @timthelion, that was an interesting read.
Sometimes the _if-then_ pattern is the way to go but I've seen many example with a _try-catch_ in Django's documentation when retrieving an object from database in order to handle the case when it doesn't exist or when it returns multiple objects. That's why I think we should keep it like this.

gwasser:
I am getting old and exception handling was more of a problem in C++, I think you're both right it's the more standard way of handling things in Python. Nevermind me, it's fine :-)

alligatorbait:
I appreciate the lessons as well, I grew up with embedded C.

PR #901, opened 2020-11-10
fix(migrations): Fix the 0034 migration for Postgres

koriaf:
although it works fine for other databases Postgres needs special care in this case. The solution looks stupid but works.


helpdesk/migrations/0019_ticket_secret_key.py
```diff
@@ -8,20 +8,22 @@ def clear_secret_keys(apps, schema_editor):
     db_alias = schema_editor.connection.alias
 
     for ticket in Ticket.objects.using(db_alias).all():
-         ticket.secret_key=''
-         ticket.save()
+        ticket.secret_key = ''
+        ticket.save()
```
Benbb96, line 12: Indentation was not good here ?

Benbb96 (commented):
I didn't know this could cause an issue on PostgreSQL.

But I don't understand why you renamed the migration 18 ?

helpdesk/migrations/0019_ticket_secret_key.py
koriaf: yes, it was indentation, something like 9 spaces instead of 8 (x4). My editor did it and stopped complaining

and regarding the migration - it seems that some time ago that migration has been called as I've made it or something, so on old database of working project in the `django_migrations` library it was called `0018_ticket_secret_key.py` so once I upgrade the code these migrations don't work. After rename it's fine; new projects will not notice the difference, the only issue are recently created ones but not 100% new.

gwasser (approved):
Sounds good, thanks!

helpdesk/migrations/0019_ticket_secret_key.py
alligatorbait: When you say recently created databases will have a problem, you mean like this?
`django.db.migrations.exceptions.InconsistentMigrationHistory: Migration helpdesk.0019_ticket_secret_key is applied before its dependency helpdesk.0018_ticket_secret_key on database 'default'.
`

koriaf: yes, exactly, it's just different first migration name. I guess you are having this issue now, because your database created after the migration has been renamed for the first time but before my fix?

My initial idea is that before first rename there are 1000 setups and after it is 100 so it's better to make 100 people angry than 1000; you have to do some `update django_migrations set name = xx where name = yy` to fix it for your database but I'm not sure if it's possible to do it somehow transparently for user

gwasser: seems like there's a way to check the name of migration history, look for either the old or new name, and adjust appropriately with if-then statements. I haven't done this before though, just hopeful.

alligatorbait:
Mine was just a development database, so I just deleted it and started
over.

Issue #867, opened 2020-08-25
Customer/User (non-staff non-superuser) unable to view or edit their submitted tickets


alligatorbait:
**Describe the bug**
An authenticated user creates a ticket (who is not a superuser or staff) and is unable to view their own ticket.

**To Reproduce**
Django Settings:
SITE_ID = 1
HELPDESK_REDIRECT_TO_LOGIN_BY_DEFAULT = True
HELPDESK_ALLOW_NON_STAFF_TICKET_UPDATE = True
HELPDESK_KB_ENABLED = False
HELPDESK_VIEW_A_TICKET_PUBLIC = False
HELPDESK_SUBMIT_A_TICKET_PUBLIC = False
HELPDESK_MAX_EMAIL_ATTACHMENT_SIZE = 0
HELPDESK_DEFAULT_SETTINGS = {
        'use_email_as_submitter': False,
        'email_on_ticket_assign': True,
        'email_on_ticket_change': True,
        'login_view_ticketlist': True,
        'tickets_per_page': 25
        }

1. Authenticate user
2. Submit a ticket, validate email address populated field for email.
3. After submission redirected to /helpdesk/tickets/[ticket number] and page displaying 403 FORBIDDEN.
```
2020-08-25 18:31:24,922 WARNING log PID: 25421 ThreadID: 140086437427008 Forbidden (Permission denied): /helpdesk/tickets/6/
Traceback (most recent call last):
  File "lib/python3.6/site-packages/django/core/handlers/exception.py", line 34, in inner
    response = get_response(request)
  File "lib/python3.6/site-packages/django/core/handlers/base.py", line 115, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "lib/python3.6/site-packages/django/core/handlers/base.py", line 113, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "lib/python3.6/site-packages/helpdesk/decorators.py", line 45, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "lib/python3.6/site-packages/helpdesk/views/staff.py", line 287, in view_ticket
    raise PermissionDenied()
django.core.exceptions.PermissionDenied
``` 

**Expected behavior**
Customer is able to view their own ticket after submission, as stated in 3. below from the Docs.
- User is directed to the /helpdesk/tickets/[ticket number] but should be directed to /helpdesk/view/... where they can view the ticket they entered.
[...]

PR #905, opened 2020-11-12
Allow users to edit tickets if HELPDESK_ALLOW_NON_STAFF_TICKET_UPDATE enabled

alligatorbait:
#867 branch 0.2.x


alligatorbait:
The travis-ci checker was giving an over indent error on a line continuation. However, without the over indent a PEP error would be created. So, I split the if, else statement into if, elif, else to bypass the line continuation solution and associated indentation errors/warnings.

gwasser:
Thanks!

PR #906, opened 2020-11-14
timeline3 offline content

alligatorbait:
Timeline 3 was added, however it requires Internet access as implemented. Conforming to the CDN pattern used for all other CSS and JS files the content has been installed and a setting must be set in order to use CDN directly from the Internet Source. This content comes from https://timeline.knightlab.com/
The entire content, not just JS and CSS files are needed for multi-language support and if any non-default fonts are used (according to the documentation).
License information: TimelineJS is released under the Mozilla Public License (MPL), version 2.0. That means that TimelineJS is free to "use, reproduce, make available, modify, display, perform, distribute" or otherwise employ. You don't need our permission to publish stories with TimelineJS and you don't need to pay us any fees or arrange any further license beyond the MPL. To read more about what you can do with TimelineJS, read our license page.


gwasser:
Thanks great idea! For future reference, I found the zip distribution of TimelineJS at this link: http://timeline.knightlab.com/docs/instantiate-a-timeline.html.

gwasser (approved):

Issue #908, opened 2020-11-24
Log file handler on the process email code


koriaf:
Description: I'm always frustrated when it creates a physical file instead of just logging things to the console (like most modern container environments are used to work with). The code is https://github.com/django-helpdesk/django-helpdesk/blob/master/helpdesk/email.py#L79

Is there any specific reason to do that? If that functionality is needed, is if fine if somebody implement setting value to not doing that (so existing setups are untouched)? We are having problems with these files on our docker setup - they are pointless there.

gwasser:
We could certainly add an option to disable the log file. Some folks prefer a record of just the email processing log for later audits or whatever.

timthelion:
I'm not sure why this code is setting up that specific handler. Typically Django's logger is highly configurable: https://docs.djangoproject.com/en/3.1/topics/logging/#examples

koriaf:
> I'm not sure why this code is setting up that specific handler. Typically Django's logger is highly configurable: https://docs.djangoproject.com/en/3.1/topics/logging/#examples

yes, my initial thinking is just to get a custom logger `queue_logger = logging.getLogger(queue.name)` and configure it on project-level if needed; but I think it's too late to change that functionality due to backwards compatibility (somebody may rely on these files already). At least without changing the major version number.

timthelion:
Well now is the perfect time to change this, because the latest release is 0.2 and master is 0.3...

gwasser:
Yes we're about to do a major version change to 0.3 so PRs welcome!

uhurusurfa:
Closing as the PR mentioned above provided the desired functioality

PR #911, opened 2020-11-25
ifix(email): Create the log file only if required

koriaf:
+ update some translations
+ update some tests to correctly mock unlink as well
+ fix flake8 errors and sort the imports in email module
+ update some log messages to pass base message + parameters instead of rendered string (works better with Sentry)

This is relatively breaking change, if someone has the logging enabled (level is set) but the directory empty they may expect files created in the standard log directory, and once it's merged they won't appear there anymore. Although console output will still be present if configured.
 #908

@gwasser @timthelion


koriaf:
Can I assume that the coverage is not blocker of this PR merging?

gwasser:
Yes coverage is not usually a blocker for me if we're making good cleanup changes. Thanks!

gwasser:
There's a typo in the word "mechanisms" here but not a huge deal

PR #920, opened 2021-01-04
Fix some login errors

Benbb96:
When making my PR #919 , I encountered issues to login on the demo app. So I try some fixes in the demo settings and also in `urls.py` where I removed a duplicate entry with name='login', which I think was causing the 302 error (too many redirects).

But since, I'm not too sure about this one, I leave it as a draft so that other can review it.


gwasser (approved):
Thanks! I believe this should fix it. I encountered the same problem and didn't have a chance over the holidays to fix it yet.

PR #924, opened 2021-01-12
Update German translation

bikeshedder:
I have also requested access on the transifex website. Is this still used?

I'm currently no active user of `django-helpdesk` and some Translations might be a bit wonky. Apart from two strings that I didn't really get what they mean it should be a decent translation nonetheless.


gwasser:
Transifex is still used. I didn't know you requested access, I did not receive a message. I'll have to check the settings. I just added you to the German team on Transifex @bikeshedder . Thank you!

PR #940, opened 2021-02-04
Tool tips improved, KB Contact Human Button replaced

alligatorbait:
New tooltips added for some icons - assign, edit user, filter tickets by user, and add email to ignore list.
Replace Knowledge Base "Contact a Human" button with "Create New Ticket" and Queue name with plus-circle icon

### DIFF
--- a/helpdesk/templates/helpdesk/kb_category_base.html
+++ b/helpdesk/templates/helpdesk/kb_category_base.html
@@ -36,7 +36,7 @@
 
                 {% endif %}
                 <a href='{% if iframe %}{% url 'helpdesk:submit_iframe' %}{% else %}{% url 'helpdesk:submit' %}{%endif%}?{% if category.queue %}queue={{category.queue.pk}};_readonly_fields_=queue;{%endif%}kbitem={{item.id}};{{query_param_string}}' class="col-sm">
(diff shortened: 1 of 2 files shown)


gwasser (approved):

timthelion:
@alligatorbait What was your usecase that lead to wanting this change? In our case we embed an iframe in a help dialog in a publicly facing application. The user does not know that they are interacting with a ticketing system, they just know that they have a problem and need help. In that case telling them what queue they're going to end up in does make much sense, and neither does "creating a ticket". All they know is that they want to contact a human, so the origional text and graphics are better for our usecase. Of course, we have no problem over-riding the template with our own, but I wonder if the original textation and graphics weren't better in most cases.

alligatorbait:
I was looking purely from interacting on the ticket system web interface,
and "Contact Human" did not provide the context to what the button was
doing within the ticket system and at which level (queue) the button would
be applied.

timthelion:
I think that maybe the IFrame view needs a different text from the one in the ticketing system, because "create a ticket" doesn't make much sense if you are in a web application's help dialogue. What do you think?

alligatorbait:
I'll defer to your judgement, I had only used it from one point of view.
Having the Queue name or KB Item etc on the button made sense from what I
was working on. Now that you have my context for the change you can take
that into account.

PR #941, opened 2021-02-04
Disabled merged_to field in edit ticket + Fix datepicker

### LINKED ISSUE #938: merged_to field needs help_text
Some users do not find the "Merged to" title to be intuitive enough.

Add help_text to the `merged_to` field in the `class Ticket` of `models.py` and update migrations as necessary.
`help_text=_('If selected, this ticket will be merged into the selected ticket.'),`

image

Renaming the field was another option, but this seems like a simple solution that will work well with minimal effort.

Benbb96:
Here is my PR to fix #938 

I override `merged_to` field in `EditTicketForm` in order to add a help text and set it as disabled since it make no sense to change that field without using the bulk action from the ticket list.

I also refactored the view and reformat the template for the edit ticket form. I noticed I could fix the datepicker (there was an error with the order of jquery import).

### DIFF
--- a/helpdesk/forms.py
+++ b/helpdesk/forms.py
@@ -83,6 +83,12 @@ class CustomFieldMixin(object):
 
 
 class EditTicketForm(CustomFieldMixin, forms.ModelForm):
+    merged_to = forms.ModelChoiceField(
+        label=_('Merged to'),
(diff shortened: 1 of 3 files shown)


gwasser (approved):
Thank you!

PR #949, opened 2021-03-04
feat(forms): Ability to provide custom public ticket form

koriaf:
Works for me with custom form; and the tests are broken the same way they were before my changes so it's fine.

probably won't be useful for many people but django-allauth uses the same behaviour which means it's useful for at least some

not sure if you have feature freeze already in place - in this case this PR may wait for the next release


gwasser (approved):
Simple enough change! Thanks.

PR #958, opened 2021-04-15
fix(email): Add ability to attach full first email text to avoid losi…

koriaf:
…ng forwards, and to save .eml files for any incoming mesages, plus fix tests and some minor bugs

I haven't tested it on all python and django versions but works fine for me

### DIFF
--- a/helpdesk/email.py
+++ b/helpdesk/email.py
@@ -21,6 +21,7 @@ from os.path import isfile, join
 from time import ctime
 
 from bs4 import BeautifulSoup
+from django.conf import settings as django_settings
 from django.contrib.auth import get_user_model
 from django.core.exceptions import ValidationError
 from django.core.files.uploadedfile import SimpleUploadedFile
@@ -394,7 +395,7 @@ def create_object_from_email_message(message, ticket_id, payload, files, logger)
         title=_('E-Mail Received from %(sender_email)s' % {'sender_email': sender_email}),
         date=now,
         public=True,
-        comment=payload['body'],
+        comment=payload.get('full_body', payload['body']) or "",
         message_id=message_id
     )
 
@@ -505,6 +506,7 @@ def object_from_message(message, queue, logger):
         ticket = None
 
     body = None
+    full_body = None
     counter = 0
     files = []
 
@@ -523,7 +525,19 @@ def object_from_message(message, queue, logger):
                 if part['Content-Transfer-Encoding'] == '8bit' and part.get_content_charset() == 'utf-8':
                     body = body.decode('unicode_escape')
                 body = decodeUnknown(part.get_content_charset(), body)
-                body = EmailReplyParser.parse_reply(body)
+                # have to use django_settings here so overwritting it works in tests
+                # the default value is False anyway
+                if ticket is None and getattr(django_settings, 'HELPDESK_FULL_FIRST_MESSAGE_FROM_EMAIL', False):
+                    # first message in thread, we save full body to avoid losing forwards and things like that
+                    body_parts = []
+                    for f in EmailReplyParser.read(body).fragments:
+                        body_parts.append(f.content)
+                    full_body = '\n\n'.join(body_parts)
+                    body = EmailReplyParser.parse_reply(body)
+                else:
+                    # second and other reply, save only first part of the message
+                    body = EmailReplyParser.parse_reply(body)
+                    full_body = body
                 # workaround to get unicode text out rather than escaped text
                 try:
                     body = body.encode('ascii').decode('unicode_escape')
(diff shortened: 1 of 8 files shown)


helpdesk/email.py
```diff
@@ -536,13 +550,23 @@ def object_from_message(message, queue, logger):
...
-                payload = """
-<html>
-<head>
-<meta charset="utf-8"/>
-</head>
-%s
-</html>""" % email_body
+                if not body and not full_body:
+                    # no text has been parsed so far - try such deep parsing for some messages
+                    altered_body = email_body.replace("</p>", "</p>\n").replace("<br", "\n<br")
```
Benbb96, line 555: Maybe there is a missing `>` for the last argument here ?

Benbb96 (approved):
This looks ok for me ! I will try to apply these changes because I also encountered the issue with eml files.

helpdesk/email.py
```diff
@@ -554,22 +578,22 @@ def object_from_message(message, queue, logger):
...
+            #     payload = payload.pop().as_string()
+            # # payloadToWrite = payload
+            # # check version of python to ensure use of only the correct error type
+            # non_b64_err = TypeError
+            # try:
+            #     logger.debug("Try to base64 decode the attachment payload")
+            #     # payloadToWrite = base64.decodebytes(payload)
+            # except non_b64_err:
+            #     logger.debug("Payload was not base64 encoded, using raw bytes")
+            #     # payloadToWrite = payload
```
timthelion, line 596: What is going on here? If the code is dead, then kill it. This is like gangrenous, rotting, zombie code at this point.

helpdesk/email.py
```diff
@@ -581,18 +605,33 @@ def object_from_message(message, queue, logger):
         if beautiful_body:
             try:
                 body = beautiful_body.text
+                full_body = body
             except AttributeError:
                 pass
         if not body:
             body = ""
 
+    if getattr(django_settings, 'HELPDESK_ALWAYS_SAVE_INCOMING_EMAIL_MESSAGE', False):
```
timthelion, line 614: Currently we already save a `email_html_body.html`. If that file isn't enough sometimes, then it would be better not to save `email_html_body.html` at all and save the `.eml` file by default. The only point of looking at `email_html_body.html` is that you would need to see missing formatting. In which case the origional `.eml` file should be even better. Indeed, I don't see the need for this to be a setting, it should be on by default.

helpdesk/email.py
```diff
@@ -523,7 +525,19 @@ def object_from_message(message, queue, logger):
                 if part['Content-Transfer-Encoding'] == '8bit' and part.get_content_charset() == 'utf-8':
                     body = body.decode('unicode_escape')
                 body = decodeUnknown(part.get_content_charset(), body)
-                body = EmailReplyParser.parse_reply(body)
+                # have to use django_settings here so overwritting it works in tests
+                # the default value is False anyway
+                if ticket is None and getattr(django_settings, 'HELPDESK_FULL_FIRST_MESSAGE_FROM_EMAIL', False):
```
timthelion, line 530: Does this need to be a setting? Can it not be on by default? If people find the forwarded parts of the messages are too long, we can add javascript later to hide them.

helpdesk/views/staff.py
```diff
@@ -409,7 +409,7 @@ def return_ticketccstring_and_show_subscribe(user, ticket):
     counter_all_ticketcc = len(all_ticketcc) - 1
     show_subscribe = True
     for i, ticketcc in enumerate(all_ticketcc):
-        ticketcc_this_entry = str(ticketcc.display)
+        ticketcc_this_entry = str(ticketcc.display).strip()
```
timthelion, line 412: Why is this needed and why should the strip be here and not here: https://github.com/django-helpdesk/django-helpdesk/blob/012ba4f42d55282e81419735f15fd81fd44f50b8/helpdesk/models.py#L1708

helpdesk/email.py
koriaf: I hope that one day someone who understands what happens gets it back; if removed then no chance to see it done

helpdesk/email.py
koriaf: my initial thinking about all that settings was to keep backwards compatibility and don't surprise existing users who don't need that; if it's not a problem then yes, I'd replace `email_html_body.html` by these `.eml` files because they are bulletproof source of truth and make this behaviour default.

helpdesk/email.py
koriaf: the same; it uglifies the first message very much, not sure if everyone is going to be happy about it. If no backwards compatibility is issue then making this default is cool.

helpdesk/views/staff.py
koriaf: agree

helpdesk/email.py
```diff
@@ -536,13 +550,23 @@ def object_from_message(message, queue, logger):
...
+                    mail = BeautifulSoup(str(altered_body), "html.parser")
+                    full_body = mail.get_text()
+
+                if "<body" not in email_body:
+                    email_body = f"<body>{email_body}</body>"
+
+                payload = (
+                    '<html>'
+                    '<head>'
+                    '<meta charset="utf-8"/'
```
Benbb96, line 565: There is a missing `>` here

koriaf: fixed with rebase

sei-gwassermann (approved):
Since @Benbb96 reviewed and approved and looks like @koriaf fixed issues brought up in comments, I'm merging this into master. Thanks!

gwasser (approved):
Whoops was logged in with the wrong account, my bad.

PR #963, opened 2021-08-01
fix(emails): Avoid 'value too long for type character varying(200) ' …

koriaf:
…error when incoming message has too long subject


Benbb96 (approved):

gwasser (approved):
Looks good, thanks!

PR #967, opened 2021-08-19
Iframe defaults fix

timthelion:
Really not sure what happened here but it seems that django stopped supporting `;` as a query arg delimiter and now only supports `&`.


gwasser (approved):
Ampersands are definitely the right way to do these queries. Thanks!

PR #970, opened 2021-09-13
Kb base iframe - fix query param syntax and wording.

timthelion:


gwasser:
I think the "Contact a Human" button was removed on purpose at one point. Does it need to be there for some code to work?

timthelion:
@gwasser It was and that change still stands, this is for the iframe view only, where the 'create a ticket' language doesn't make any sense. See https://github.com/django-helpdesk/django-helpdesk/pull/940

Btw, including the name of the queue is very wordy, I'm not sure if it is good UX at all. Even within the helpdesk itself...

gwasser (approved):
OK that makes sense then, thanks.

Issue #983, opened 2021-11-17
validators=[validate_file_extension] not working on ticket attachment


noobpk:
HI, I'm trying to fix the bug of uploading arbitrary attachments on a ticket. 
That resulted in the Stored XSS vulnerability being disclosure at: https://huntr.dev/bounties/4d7a5fdd-b2de-467a-ade0-3f2fb386638e/

When uploading attachments, the application will call def `process_attachments` then call class `FollowUpAttachment`
https://github.com/django-helpdesk/django-helpdesk/blob/91b37f6d73e31dc312c8e3a54b9578b0a0851624/helpdesk/lib.py#L131-L148
https://github.com/django-helpdesk/django-helpdesk/blob/91b37f6d73e31dc312c8e3a54b9578b0a0851624/helpdesk/models.py#L1087

These parameters will be passed to the class `Attachment`. And at `file` it must be checked by `validators=[validate_file_extension]`
https://github.com/django-helpdesk/django-helpdesk/blob/91b37f6d73e31dc312c8e3a54b9578b0a0851624/helpdesk/models.py#L1017-L1029
https://github.com/django-helpdesk/django-helpdesk/blob/91b37f6d73e31dc312c8e3a54b9578b0a0851624/helpdesk/validators.py#L7-L15
But somehow it got through and resulted in being able to upload arbitrary attachments.
<img width="553" alt="image" src="https://user-images.githubusercontent.com/31820707/142244486-f82978b9-31e2-4083-ae63-3d5e42fdc06b.png">

Can someone help me with this issue?

gwasser:
I'm not certain why the filter failed either. Is it correctly calculating the file extensions?

noobpk:
> I'm not certain why the filter failed either. Is it correctly calculating the file extensions?

it just doesn't work with ticket attachments but KB attachments it works fine

gwasser:
oh that's interesting. maybe silly, but is this an existing instance, and if so, have you run all the migrations? I remember that the Attachments class was updated recently to serve as a base for ticket and KB attachments, and maybe that change caused some issues somewhere.

Benbb96:
I believe the issue is that in the `process_attachments` function, the `FollowUpAttachment` instance is only saved in database and that doesn't trigger the validators : https://github.com/django-helpdesk/django-helpdesk/blob/91b37f6d73e31dc312c8e3a54b9578b0a0851624/helpdesk/lib.py#L148

[...]

PR #984, opened 2021-11-18
Add `att.full_clean()` before saving

noobpk:
Fix issue https://github.com/django-helpdesk/django-helpdesk/issues/983 as well as fix bug stored XSS disclosure: https://huntr.dev/bounties/4d7a5fdd-b2de-467a-ade0-3f2fb386638e/

Reference: 
> https://docs.djangoproject.com/en/1.10/ref/models/instances/#validating-objects
> https://stackoverflow.com/a/40888943/8439435

### DIFF
--- a/helpdesk/lib.py
+++ b/helpdesk/lib.py
@@ -145,6 +145,7 @@ def process_attachments(followup, attached_files):
                 'application/octet-stream',
                 size=attached.size,
             )
+            att.full_clean()
             att.save()
 
             if attached.size < max_email_attachment_size:


gwasser:
oops, looks like this works now, but it breaks the tests since we can't upload files with those extensions anymore. Probably just renaming the test text files to a `.email` extension or something not in the filter would work.

gwasser (approved):

noobpk:
> oops, looks like this works now, but it breaks the tests since we can't upload files with those extensions anymore. Probably just renaming the test text files to a `.email` extension or something not in the filter would work.

I often fix the bug that XSS can execute in .svg or .html files using CSP. Don't know if this can be used.

gwasser:
I'm going to merge it in the interests of the patch getting out sooner, but will have to work on fixing the tests. thanks everyone!

PR #991, opened 2022-01-27
Show active page in sidebar depending on the request path url

Benbb96:


gwasser:
Seems like a good simple improvement! thanks.

Issue #1000, opened 2022-03-11
Ticket dependency delete confirmation - don't delete button click


afranjin:
**Describe the bug**
After add dependency to ticket and then want remove the confirm modal append and on do not delete 404 not find. Fixed wrang path on button click.

**To Reproduce**
Ticket dependency - do not delete 404
Steps to reproduce:
1.Log in with user(admin-role)
2.Helpdesk
3.All Tickets
4.Click any Ticket
5.Add Ticket Dependency
6.Delete Ticket Dependency
7.Don't Delete

Page not found (404)
Request Method: | GET
-- | --
http://127.0.0.1:8080/tickets/12/dependency/

**Expected behavior**
Need to redirect to ticket edit after button do not delete is clicked.

**Desktop (please complete the following information):**
 - OS:MacOs - M1
 - Browser chrome
 - Version 98.0.4758.109 (Official Build) (arm64)

gwasser:
Thank you! This should be fixed now that #1001 is merged, so I will close this ticket.

PR #1001, opened 2022-03-11
Fix wrong href.

afranjin:
The href path of don't delete button needs to be ../../../

#1000

### DIFF
--- a/helpdesk/templates/helpdesk/ticket_dependency_del.html
+++ b/helpdesk/templates/helpdesk/ticket_dependency_del.html
@@ -7,7 +7,7 @@
     <a href="{% url 'helpdesk:list' %}">{% trans "Tickets" %}</a>
 </li>
 <li class="breadcrumb-item">
-    <a href="{% url 'helpdesk:list' %}{{ ticket.id }}/">{{ ticket.queue.slug }}-{{ ticket.id }}</a>
+    <a href="{% url 'helpdesk:list' %}{{ dependency.ticket.id }}/">{{ dependency.ticket.queue.slug }}-{{ dependency.ticket.id }}</a>
 </li>
 <li class="breadcrumb-item active">{% trans "Delete Ticket Dependency" %}</li>
 {% endblock %}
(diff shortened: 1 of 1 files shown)


helpdesk/templates/helpdesk/ticket_dependency_del.html
```diff
@@ -18,7 +18,7 @@ <h2>Delete Ticket Dependency</h2>
 <p>Are you sure you wish to remove the dependency on this ticket?</p>
 {% endblocktrans %}
 
-<p><a href='../../'><button class="btn btn-primary btn-lg">{% trans "Don't Delete" %}</button></a></p>
+<p><a href='../../../'><button class="btn btn-primary btn-lg">{% trans "Don't Delete" %}</button></a></p>
```
Benbb96, line 21: Maybe we should use `{{ ticket.get_absolute_url }}` instead of a a relative URL.

afranjin: Changed in 3ea2d1dd501b249c43f1c2ed2ce67f09c6e615d6

helpdesk/templates/helpdesk/ticket_dependency_del.html
```diff
@@ -7,7 +7,7 @@
     <a href="{% url 'helpdesk:list' %}">{% trans "Tickets" %}</a>
 </li>
 <li class="breadcrumb-item">
-    <a href="{% url 'helpdesk:list' %}{{ ticket.id }}/">{{ ticket.queue.slug }}-{{ ticket.id }}</a>
+    <a href="{% url 'helpdesk:list' %}{{ dependency.ticket.id }}/">{{ dependency.ticket.queue.slug }}-{{ dependency.ticket.id }}</a>
```
Benbb96, line 10: Good catch, I didn't see that ticket was not available in the template. You can replace `{% url 'helpdesk:list' %}{{ dependency.ticket.id }}/` by `{{ dependency.ticket.get_absolute_url }}` 😉

Benbb96 (approved):

gwasser:
Thank you @afranjin and @Benbb96 for working on this! Makes it easy for me to look at and approve.

PR #1003, opened 2022-03-18
Disable KB without Django errors

Mte90:
So if you set as false `HELPDESK_KB_ENABLED` in the settings there are various python errors because doesn't find stuff.
This patch helps on those errors.


gwasser:
Thank you for your work on this!

PR #1007, opened 2022-03-31
Create an API endpoint to manage tickets (ignoring custom fields)

Benbb96:
This one is simpler and it works well but there is no support for custom fields.


gwasser:
Work toward #198

PR #1011, opened 2022-04-15
Disable timeline on ticket list

Mte90:
As per title, PR very simple but disable the timeline to simplify the interface

### DIFF
--- a/demo/demodesk/config/settings.py
+++ b/demo/demodesk/config/settings.py
@@ -114,6 +114,8 @@ HELPDESK_SUBMIT_A_TICKET_PUBLIC = True
 # Should the Knowledgebase be enabled?
 HELPDESK_KB_ENABLED = True
 
+HELPDESK_TICKETS_TIMELINE_ENABLED = True
+
 # Allow users to change their passwords
 HELPDESK_SHOW_CHANGE_PASSWORD = True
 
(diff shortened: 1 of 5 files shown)


timthelion:
It is possible to over-ride templates in django by creating a new template in your project with the same name and making sure your template dirs are ordered properly. This gives you full control over the template's appearance. It is much cleaner to over-ride the template than to have endless options for customizing them. The disadvantage of over-riding the templates is that if context variables change, then your custom templates could break, but the maintenance burden of supporting a million SETTINGS is too great for this project.

Benbb96 (approved):

Mte90:
> It is possible to over-ride templates in django by creating a new template in your project with the same name and making sure your template dirs are ordered properly. This gives you full control over the template's appearance. It is much cleaner to over-ride the template than to have endless options for customizing them. The disadvantage of over-riding the templates is that if context variables change, then your custom templates could break, but the maintenance burden of supporting a million SETTINGS is too great for this project.

Yes I had the same doubt, it wasn't clear to me how a setting is defined, I mean when it is the case to enable a setting to change something.

gwasser:
A custom template could accomplish this same thing pretty easily, just copy the existing template and remove the stuff you don't want. We can create our own settings whenever it seems useful to do so. I don't mind a setting if it's a common request.

gwasser:
The main thing is that any new settings should be properly documented in the `docs/` folder so that new users know how to use it and what it does. Could you add some documentation to this PR, @Mte90 ? Thanks!

Mte90:
I will update asap with the docs stuff :-)

Mte90:
Documentation updated :-)

PR #1014, opened 2022-05-02
Create user by REST endpoint

Mte90:
It's the first time that I work with them but I was able to create an endpoint that works. 

I am not sure about naming convention and so on but works :-)

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -46,6 +46,16 @@ Here is an example of a cURL request to create a ticket (using Basic authenticat
     --header 'Content-Type: application/json' \
     --data-raw '{"queue": 1, "title": "Test Ticket API", "description": "Test create ticket from API", "submitter_email": "test@mail.com", "priority": 4}'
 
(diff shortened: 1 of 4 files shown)


timthelion:
Is there some kind of permissions check?

Mte90:
I was thinking that all the rest endpoints there require the token so also this one.  
I used the web interface on http://127.0.0.1:8080/api/users/ that allows just POST.

gwasser:
Thanks! I mostly want to make sure we consider security here. Is this a necessary thing for the API? Should the admin log into the admin panel to create users instead?

timthelion:
The other thing that occures to me is that this isn't a helpdesk specific task so might fit better somewhere not in the helpdesk code...

Mte90:
I will update the PR with that change.

About the implementation usually ticket platforms have this so you can create it.

gwasser:
@Mte90 merged! I will likely wrap this up into the next release this weekend when I have a little more time. thank you for your work on this!

Issue #1025, opened 2022-06-25
can't follow install.html


bbrendon:
https://django-helpdesk.readthedocs.io/en/latest/install.html

I went through this and depending on if the view is added to urls.py, I get these errors.

Is there a missing step regarding startapp ?

It seems something is missing in those instructions, but I'm not sure what.

```
❯ python manage.py migrate helpdesk

SystemCheckError: System check identified some issues:

ERRORS:
helpdesk.KBItem.team: (fields.E300) Field defines a relation with model 'pinax_teams.Team', which is either not installed, or is abstract.
helpdesk.KBItem.team: (fields.E307) The field helpdesk.KBItem.team was declared with a lazy reference to 'pinax_teams.team', but app 'pinax_teams' isn't installed.
```

```
$ python manage.py migrate helpdesk
...
    return import_module(self.urlconf_name)
  File "/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.8/lib/python3.8/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1014, in _gcd_import
  File "<frozen importlib._bootstrap>", line 991, in _find_and_load
  File "<frozen importlib._bootstrap>", line 975, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 671, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 783, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/Users/bbrendon/p/django-helpdesk/my_d_proj/my_d_proj/urls.py", line 21, in <module>
    path("helpdesk/", include("helpdesk.urls")),
NameError: name 'include' is not defined
```

bbrendon:
after adding `from django.conf.urls import include`
I'm back to https://github.com/django-helpdesk/django-helpdesk/issues/1020

gwasser:
thanks, are you using Django 4? must have been a recent change.

PR #1026, opened 2022-06-25
update install for noobs like me

bbrendon:


bbrendon:
fixes https://github.com/django-helpdesk/django-helpdesk/issues/1025

gwasser:
Thank you @bbrendon!

PR #1029, opened 2022-07-12
Formatting

martin-marty:
Auto format using pycodestyle


helpdesk/settings.py
```diff
@@ -25,6 +25,9 @@
 
 HAS_TAG_SUPPORT = False
 
+# Use international timezones
+USE_TZ: bool = True
```
Benbb96, line 29: I don't understand why it's necessary to add this setting. It should already be set in the project django configuration using django-helpdesk.

gwasser: It can be changed in the project configuration but this would set it as the default value for new projects, so seems reasonable to me.

martin-marty: The default value has changed in Django 4, this will prevent the deprecation warning from being emmiited.

gwasser:
Looks like there's a conflict in the tests. You should be able to do a pull from stable and do formatting on the missing function, and do another commit here to fix it. Thanks for your work on this @martin-marty !

helpdesk/settings.py
Benbb96: Alright, thank you for the explanation.

gwasser:
Thanks for your work on this @martin-marty .

PR #1030, opened 2022-07-20
GitHub workflow

martin-marty:
Fixed up unstable to have the latest changes from stable.

I had to all but remove the `helpdesk.models.EscapeHtml.exendMarkdown` method, there's unfortunately no documentation on what it's trying to achieve and with the latest version of Markdown, it was throwing errors.  If you can let me know what this _should_ be doing, I'll be happy to fix up.

In the meantime though, there is an initial github workflow added which is just performing the 'quicktest.py' at the moment to get this going.


gwasser (approved):
Thank you! Honestly I am not sure what `extendMarkdown` is supposed to do either, I suspect it's a workaround for an old version of markdown libraries that may simply not be needed anymore. I think what you did here is sensible for now, if we don't catch any weird output or broken tests suddenly we can totally remove it later.

Issue #969, opened 2021-09-10
Redundant cc code in email.py?


ebeers-png:
**Describe the bug**
I'm doing some work on helpdesk/email.py for a fork of Helpdesk. I noticed that the work done in [lines 507-516](https://github.com/django-helpdesk/django-helpdesk/blob/master/helpdesk/email.py#L507) don't seem to have any effect, and when run, the block doesn't parse the emails from `message.get_all('cc', None)` properly in one case.

The block seems to be leftover from some rework commits, since the variable `cc` isn't used elsewhere; later on, the line 360 creates a list of addresses correctly and in fewer lines.

Is there a reason for this code block to be here? Or can I remove it without issue?

**To Reproduce**
Here's a reproduction based on one email I imported. For this email, on line 507, `message.get_all('cc', None)` produces this list of length 1 (I've randomized the names/domain):
```
[
    '"Brunet, Angelus (ABC)" <Angelus.Brunet@abc.com>, "Durant, Lawrence ABC)"\r\n\t<Durant.Lawrence@abc.com>, "Abrami, Henri (ABC)" <henri.abrami@abc.com>, \r\n\t"Barta, Kai (ABC)" <Kai.Barta@abc.com>'
]
```

Using that value for `cc`, and replacing line 510's `message.get_charset()` with `None`, the final result on line 516 splits on the commas inside the names, and it doesn't strip `\r\n\t` from one of the strings.
```
{
    '"Barta',
    'Kai (ABC)" <Kai.Barta@abc.com>',
    '"Durant', 
    'Angelus (ABC)" <Angelus.Brunet@abc.com>',
    '"Abrami',
    'Henri (ABC)" <henri.abrami@abc.com>',
    '"Brunet',
    'Lawrence ABC)"\r\n\t<Durant.Lawrence@abc.com>'
}
```

**Expected behavior**
I would expect the block to instead produce the following:
```
{
    '"Barta, Kai (ABC)" <Kai.Barta@abc.com>', 
    '"Brunet, Angelus (ABC)" <Angelus.Brunet@abc.com>', 
    '"Abrami, Henri (ABC)" <henri.abrami@abc.com>',
    '"Durant, Lawrence (ABC)" <Lawrence.Durant@abc.com>'
}
```

**Additional context**
My fork of email.py is up to date with the latest commit in the master branch. I'm debugging email.py locally, by running `./manage.py get_email` with a real email address using IMAP.

timthelion:
[...]

PR #1036, opened 2022-07-22
Remove left over 'cc' code

martin-marty:
Fixes https://github.com/django-helpdesk/django-helpdesk/issues/969


gwasser (approved):
Great! Easy fix, I was worried there was more stuff hiding in the function. Thanks!

Issue #976, opened 2021-09-26
attachments are not deleted


dev72:
Hello,
when I delete an issue attachments are still in media folder.
I think they should also be deleted.

timthelion:
Seems like maybe the first answer here could be coopted: https://stackoverflow.com/questions/16041232/django-delete-filefield but the reason this was removed from django (as explained in the answer) remains. If the transaction is rolled back, you end up in an inconsistent state. Perhaps, instead, files should be cleaned up during the mail getting task... Not sure really.

gwasser:
Exactly, files are not removed automatically/by default in order to allow
the database to remain in a consistent state should you decide to
"undelete" or roll back. I think even the tickets and such don't actually
get deleted, only marked as "deleted" in the database so they don't show up
in queries. We could create an admin tool to search the database and
filesystem for outdated data and files (say, older than 30 days?) and
permanently remove it but right now nothing like that exists.

dev72:
there is no way to undelete or rollback after successful deletion: tickets are deleted from table.
Should simply delete attachments after the deletion transaction ended without errors.

gwasser:
I'll have to look into this more to see exactly what the code does. maybe
we changed the default over all the work the last year. if it is not
removing attachments automatically with the ticket, we should fix that.

uhurusurfa:
Probably best to make it a configurable option globally for now and use this project:
https://github.com/un1t/django-cleanup

The documentation can be updated to specify the settings file to including the library explicitly or  exclude it if there is an environment variable set something like:
```
    if "true"== os.getenv("DELETE_ATTACHMENTS_ON_FILE_SYSTEM", "true"):
        INSTALLED_APPS = INSTALLED_APPS + ("django_cleanup.apps.CleanupConfig",)
```
The solution becomes a documentation one.

gwasser:
django cleanup sounds like the right solution, thanks @uhurusurfa

PR #1037, opened 2022-07-23
Add django-cleanup app to support deleting file system files

uhurusurfa:
This relates to this issue:
https://github.com/django-helpdesk/django-helpdesk/issues/976

If deleting files is not desired, the end user can remove the app from INSTALLED_APPS


gwasser (approved):
Thanks! Does the helpdesk code need updating to implement or does this mostly happen automatically now?

Issue #1039, opened 2022-07-24
complexity reduction


martin-marty:
Reduce the complexity of the following:

- [x] ./helpdesk/email.py:371:1: C901 'create_object_from_email_message' is too complex (21)
- [x] ./helpdesk/email.py:519:1: C901 'object_from_message' is too complex (29)
- [x] ./helpdesk/views/staff.py:484:1: C901 'update_ticket' is too complex (43)
- [x] ./helpdesk/views/staff.py:910:1: C901 'merge_tickets' is too complex (27)
- [x] ./helpdesk/views/staff.py:1059:1: C901 'ticket_list' is too complex (26)
- [x] ./helpdesk/views/staff.py:1420:1: C901 'run_report' is too complex (36)

martin-marty:
@gwasser https://github.com/django-helpdesk/django-helpdesk/pull/1041 for cross-reference

gwasser:
Looks like all the functions were updated in #1041 so closing this. Thanks!

PR #1041, opened 2022-07-25
Unstable

martin-marty:
@gwasser Hi, addressed issues here:

https://github.com/django-helpdesk/django-helpdesk/issues/1039

Also, updated the workflow with the flake8 complexity so values > 20 will be rejected.


gwasser (approved):
Looks good, thanks!

PR #1050, opened 2022-08-15
Fix PyCharm warnings in staff.py

Benbb96:
I had some warnings in my PyCharm IDE so I tried to fix them up.

I also ran the quicktest.py and get 3 failures + 1 error but they doesn't seem to be related to my changes.


gwasser (commented):
Thanks! Hmm it looks like it's doing a lot of stuff. I'm mostly concerned if there's any issues with for example I saw something changed from str to int and if that affects what goes into the database and so could cause errors later that we're not currently testing for. I won't merge this in quite yet as I want a little more time to look more carefully and have some other eyeballs look with me. But I imagine it is most likely fine.

helpdesk/views/staff.py
```diff
@@ -356,7 +356,7 @@ def view_ticket(request, ticket_id):
         )[1]
 
         if show_subscribe:
-            subscribe_staff_member_to_ticket(ticket, request.user)
+            subscribe_to_ticket_updates(ticket, request.user)
```
uhurusurfa, line 359: Nice rationalisation of the code.

helpdesk/views/staff.py
```diff
@@ -471,30 +471,19 @@ def subscribe_to_ticket_updates(ticket, user=None, email=None, can_view=True, ca
...
             can_update=can_update
         )
-        ticketcc.save()
-
-        return ticketcc
-
-
-def subscribe_staff_member_to_ticket(ticket, user, email='', can_view=True, can_update=False):
-    """used in view_ticket() and update_ticket()"""
-    return subscribe_to_ticket_updates(ticket=ticket, user=user, email=email, can_view=can_view, can_update=can_update)
```
uhurusurfa, line 480: Nice rationalisation of the code.

helpdesk/views/staff.py
```diff
@@ -508,20 +497,14 @@ def get_ticket_from_request_with_authorisation(
...
-            )
-
-        if not ticket:
-            return HttpResponseRedirect(
-                '%s?next=%s' % (reverse('helpdesk:login'), request.path)
+                submitter_email__iexact=request.POST.get('mail'),
+                secret_key__iexact=request.POST.get('key')
             )
+        except (Ticket.DoesNotExist, ValueError):
+            return redirect_to_login(request.path, 'helpdesk:login')
```
uhurusurfa, line 508: This change may not be ideal.
It should probably check if "key" and "email" are in the request and retiurn a different error message than just redirecting to the login page. 
I am not familiar with the UI usage of this form so it may be acceptable.

helpdesk/views/staff.py
```diff
@@ -562,10 +545,10 @@ def get_due_date_from_request_or_ticket(
 
 
 def get_and_set_ticket_status(
-    new_status: str,
+    new_status: int,
     ticket: Ticket,
     follow_up: FollowUp
-) -> typing.Tuple[str, str]:
+) -> typing.Tuple[str, int]:
```
uhurusurfa, line 551: This looks correct since there is a cast in the calling method:
`new_status = int(request.POST.get('new_status', ticket.status))`

helpdesk/views/staff.py
```diff
@@ -273,8 +274,7 @@ def followup_edit(request, ticket_id, followup_id):
             'time_spent': format_time_spent(followup.time_spent),
         })
 
-        ticketcc_string, __ = \
-            return_ticketccstring_and_show_subscribe(request.user, ticket)
+        ticketcc_string = return_ticketccstring_and_show_subscribe(request.user, ticket)[0]
```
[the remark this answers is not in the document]
Benbb96: You are right, this will be better with a unit test for this case. But my change shouldn't break anything since I get first element of the tuple with `[0]`. I wanted to get rid of the useless variable `__`

helpdesk/views/staff.py
Benbb96: Indeed, I simply refactored this part of the code but an error handling would be better. Maybe I can only catch `Ticket.DoesNotExist` here and let the ValueError be handled in the calling view (which is `update_ticket`) ?

helpdesk/views/staff.py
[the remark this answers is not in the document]
uhurusurfa: Ah! ... I missed that [0]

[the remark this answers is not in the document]
uhurusurfa: NOTE: If you use a single undescore instead of a double underscore, the linters will not complain about an unused variable.

helpdesk/views/staff.py
gwasser: Good catch! There might be a need for some staff specific logic but we can always add that back later when needed since this function doesn't seem to have done anything.

helpdesk/views/staff.py
gwasser: good point. does that make the cast redundant and able to be removed?

helpdesk/views/staff.py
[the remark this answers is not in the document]
gwasser: I missed the array indexing too, thanks. sounds good. do we need to update to single underscore to pass tests?

helpdesk/views/staff.py
gwasser: does `update_ticket` do any error catching for ValueError? I don't have time at the moment to investigate.

Benbb96: No it is not currently handling this type of Exception. But I'm still wondering what would be the best. It seems to be easier to do error handling directly in that view since we have access to the request in the parameter

gwasser: I think you're probably right, it just means we might need to do some extra work on `update_ticket` too as part of this PR. thanks!

Mte90: Maybe something frontend with an alert?

uhurusurfa:
It is probably worth refactoring the code around get_ticket_from_request_with_authorisation() a bit.
Given that if "public" is set to True it just tries to get a ticket and the only time it is set to true is via the update_tcicket() call in helpdesk/views/public.py, it is pointless calling update_ticket from helpdesk/views/public.py at all and it should just call get_object_or_404(Ticket, id=ticket_id) directly.
That would simplify the logic in the get_ticket_from_request_with_authorisation() somewhat and "public" can be removed from the method signature.

The `if` statement seems to check stuff that is implicitly already checked as well and could be simplified:
`    if not (public or (
            request.user.is_authenticated and
            request.user.is_active and (
                is_helpdesk_staff(request.user) or
                helpdesk_settings.HELPDESK_ALLOW_NON_STAFF_TICKET_UPDATE))):
`
The user can only be authenticated if the user is active  (this is done in the authenticator module) so `and request.user.is_active ` can be removed.

If the above removal of "public" is implemented then only authenticated user requests will ever invoke the get_ticket_from_request_with_authorisation() method so forcing the user to the login screen does not really make sense.
The reason why the ticket may not be found could be a configuration error in the app or it could be someone trying to hack the system and modify tickets they do  not have access to so it should possibly log the user Id used to try and access the record with a message like f"Possible attempt to access unauthorised ticket by user {request.user}" and return an error screen to the user.
Instead there should be a screen that

uhurusurfa (approved):
Tests all good so change accepted.

PR #1053, opened 2022-09-08
Attachment file name handling,  email containing comma in "real name" and ignored emails

uhurusurfa:
PR has the following changes:

- Refactored the object_from_message() method to make it easier to reationalise about and more easily testable
- Changed the way the filename is extracted from the attachment to use the standard email module mechanism.
- Fixed the workaround to handle a comma in the real name part of an email address. Instead of pre-decoding the header, inoke the email,utils.parseaddr method that does all the work for us and handles the extraction compliant with RFC2822
- Added support for the "keep_in_mailbox" field on the IgnoreEmail model. The handling of the ignored emails was enhanced to specifically handle ignored emails instead of folding them into general handling processes by raising specific exceptions that allow targeted handling in the caller methods instead of returning True/False when the method is supposed to return a Ticket object.
- Add a utils.py module for testing that provides helper methods to more easily extend the testing of emails


gwasser (approved):
Looks like good refactoring and fixes, thanks very much for your work on this!

PR #1055, opened 2022-10-09
Fix fail email check when an attachment has a wrong extension

Benbb96:
I used a try except in the `email.py` command in order to mute the ValidationError. I added one test to validate it.

I also add the validator on the files field of the `AbstractTicketForm`.

### DIFF
--- a/helpdesk/email.py
+++ b/helpdesk/email.py
@@ -478,12 +478,16 @@ def create_object_from_email_message(message, ticket_id, payload, files, logger)
 
     logger.info("[%s-%s] %s" % (ticket.queue.slug, ticket.id, ticket.title,))
 
-    attached = process_attachments(f, files)
-    for att_file in attached:
-        logger.info(
-            "Attachment '%s' (with size %s) successfully added to ticket from email.",
-            att_file[0], att_file[1].size
-        )
+    try:
+        attached = process_attachments(f, files)
+    except ValidationError as e:
+        logger.error(str(e))
+    else:
+        for att_file in attached:
+            logger.info(
+                "Attachment '%s' (with size %s) successfully added to ticket from email.",
+                att_file[0], att_file[1].size
+            )
 
     context = safe_template_context(ticket)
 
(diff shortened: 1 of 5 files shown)


uhurusurfa (approved):
👍 Looks good tome.

ilikerobots:
@Benbb96  / @uhurusurfa : I admit I haven't verified this with a run, but from reading the PR, it appears that if a email comes in with N attachments, any any 1 of them is invalid, the resulting ticket will have zero attachments.  Is this correct and, if so, is this the desired behavior?

uhurusurfa:
You are correct - it probably needs to catch each exception on an attachment individually and allow the valid attachments to be processed.
@Benbb96 - can you raise a PR to change the behaviour?

Benbb96:
@ilikerobots yes you're right, I didn't think of this.. Thanks for pointing it out. As you said @uhurusurfa, it may be best to find a way to only mute errors from invalid files and let the other be saved.

Yes I think I'll be able to rework on that this evening.

Benbb96:
Hi @uhurusurfa, I'm not sure if you've seen my last PR #1056 where I fixed the problem pointed by @ilikerobots 

By the way, I was wondering if _django-helpdesk_ is participating in the [HacktoberFest](https://hacktoberfest.com/) event? It could be nice to get the rewards from our contributions and maybe attract new contributors 😃

uhurusurfa:
Hi Ben - yes sorry. I have a lot on my plate over the next week or so
having returned from an extended vacation recently and then should have the
bandwidth to take a close look at the code. Since you have not rolled back
any of the previous code changes, I need to verify that there is no edge
case where it would exclude an attachment.

uhurusurfa:
@Benbb96 Regarding Hacktoberfest, I think we would have to have a well defined set of requirements for people to build stuff on and I personally do not have the time over the next 3-4 weeks to do that. If you feel there are well enough spec'd additions somewhere for the Helpdesk app then please feel free to register the project for help.

Benbb96:
Alright, no problem about that !

Yes, I think I agree with you, we would have need to do a triage of issues and label `good-first-issues` (even for no-code contributions) to call for help.  
But now I believe it's too late to do it so maybe next year :wink:

Issue #1066, opened 2023-01-24
Due date change detected when no change made


finnertysea:
**Describe the bug**
In the current stable release, if a ticket has a due date set then saving the form on the view ticket page results in a (false) due date change being detected. 
The view then adds a TicketChange and FollowUp to record the (false) due date change, despite no change actually being made.

**To Reproduce**
1. Create a ticket with a due date
2. Save
3. Edit any detail about the ticket (e.g. add a comment) via the view ticket page
4. A note is automatically added reporting that due_date was changed, but the due date itself remains the same

**Expected behaviour**
Only the edit or comment made appears in the thread.

**Diagnosis**
These POST requests are handled by the update_ticket view.  This view currently assumes the POST data for due_date always has the format mm/dd/yy.  A comment in the code, consistent with some custom re logic, reveals this assumption:

NOTE: jQuery's default for dates is mm/dd/yy --  very US-centric but for now that's the only format supported until we clean up code to internationalize a little more. But the ticket.html template uses datepicker with format yy-mm-dd:
`$( "#id_due_date" ).datepicker({dateFormat: 'yy-mm-dd'});`
 With standard default input date formats, Django correctly transforms the POST data into the correct datetime object.

**Proposed Fix**
Short-term we can improve this by simply removing the custom re logic and using Django's built-in date parsing. Long-term, we should be using a form in this view to do validation and transformation and implement consistent date handling to support I18N and L10N package-wide. Pull request following with short-term fix.

**Desktop (please complete the following information):**
 - OS: MacOS Ventura 13.0.1
 - Browser: Safari
 - Version 16.1

PR #1067, opened 2023-01-24
Bug fix for #1066 - detected due date change when no change was made

finnertysea:
This change is a patch for issue #1066.  It utilizes the current surrounding architecture and interferes earlier in the process to transform the POST data item into a tz aware datetime object so that no false change is detected upon comparison to the stored due date.

A more thorough fix would involve re-factoring this view to use Django's builtin form handling logic. Probably best to do as part of a larger effort to support I18N and L10N  package-wide.
I don't see any reason this shouldn't work in practice, and have tested it in a variety of scenarios. However, I couldn't think of a good way to add a unit test because the bug is caused by mismatch between client and server side of the application.  It would be good if someone has an idea for a unit test that might help in future with regression testing.

### DIFF
--- a/helpdesk/views/staff.py
+++ b/helpdesk/views/staff.py
@@ -23,10 +23,10 @@ from django.http import Http404, HttpResponse, HttpResponseRedirect, JsonRespons
 from django.shortcuts import get_object_or_404, redirect, render
 from django.urls import reverse, reverse_lazy
 from django.utils import timezone
-from django.utils.html import escape
-from django.utils.translation import gettext as _
 from django.utils.dateparse import parse_datetime
+from django.utils.html import escape
 from django.utils.timezone import make_aware
+from django.utils.translation import gettext as _
 from django.views.decorators.csrf import requires_csrf_token
 from django.views.generic.edit import FormView, UpdateView
 from helpdesk import settings as helpdesk_settings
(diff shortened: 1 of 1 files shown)


uhurusurfa (changes requested):
Please fix the ordering of imports per the CI/CD fail

finnertysea:
isort failure is in helpdesk/lib.py. lib.py hasn't been changed since Oct 2022. Possible CI/CD change? 

Same issue is causing #1070 to fail.

uhurusurfa:
@finnertysea - possibly a library upgrade that fixed a bug or changed core functionality in isort. If you could fix eventhough it is not your PR that is causing the failure it would be appreciated

finnertysea:
I couldn't determine why this started happening this PR since I'm not super familiar with CI/CD processes. I did find that isort stumbles on lib.py in the following segment.
Screenshot 2023-02-14 at 17 47 56
The change I've made on the right clears isort, but changes the behaviour a little, although I think it should be fine.

I would have expected isort to start having issues with this file after revision ce95949c50110abc0c1aba45c1b8952b902bd91b which changed the namespacing from `helpdesk.akismet` to just `akismet`, thereby violating isort's alphabetical import rule.

Let me know if you'd like me to add it to the PR or make a change yourself.

uhurusurfa (approved):
Whilst it could be done using Django function per other review, the end result is the same so all good in the end.
Thanks for the fix.

Issue #1069, opened 2023-01-27
Inconsistent username regex in urls.py


finnertysea:
**Describe the bug**
The regex to match user_name in re_path for rss_user is much more restrictive than the regex for user_name in rss_user_queue. This leads to, for example, usernames containing hyphens to raise an error for the former but not the latter.

Compare regex in urls.py on line 154 vs 159 in current stable branch.

**To Reproduce**
1. Logged in as a user with a hyphen (or any character that is non-alphanumeric, _, or .)
2. Go to /rss
3. Click on "My Open Tickets"
4. See "ERROR (EXTERNAL IP): Internal Server Error: /rss/"

**Expected behavior**
Should be redirected to /rss/user/<username>/. Note that clicking on any of the queue-specific rss links under "My Open Tickets" works fine. 

**Desktop (please complete the following information):**
 - OS: macOS Ventura 13.0.1
 - Browser: Safari
 - Version 16.1

**Additional context**
Modern Django uses "path converters" instead of hard-coded regex. See https://docs.djangoproject.com/en/3.2/topics/http/urls/#path-converters/

Maybe a future improvement could switch to using this method for urls?

uhurusurfa:
Merged PR

PR #1070, opened 2023-01-27
#1069 - make regex for username more permissive for rss_user to match rss_user_queue

finnertysea:
As a simple fix for #1069, matching existing code. This changes the regex for rss_user to match the more permissive regex for rss_user_queue.

This fixes the behaviour seen in #1069, but does not address the problem of django-helpdesk urls being out of date with modern Django url practices using path converters. Since the builtin Django slug converter does not allow "." characters, a custom username converter could be built to allow 
```
re_path(
    r"^rss/user/(?P<user_name>[a-zA-Z0-9\_\.]+)/", 
    helpdesk_staff_member_required(feeds.OpenTicketsByUser()),
    name="rss_user",
),
```

to become:

```
path(
    "rss/user/<username:user_name>/",
    helpdesk_staff_member_required(feeds.OpenTicketsByUser()),
    name="rss_user",
),
```

### DIFF
--- a/helpdesk/urls.py
+++ b/helpdesk/urls.py
@@ -151,7 +151,7 @@ urlpatterns += [
 
 urlpatterns += [
     re_path(
-        r"^rss/user/(?P<user_name>[a-zA-Z0-9\_\.]+)/",
+        r"^rss/user/(?P<user_name>[^/]+)/",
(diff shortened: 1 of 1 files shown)


uhurusurfa (approved):
Seems sensible

Issue #1075, opened 2023-03-17
Fix for #1066 is incomplete


finnertysea:
**Describe the bug**
PR #1067 introduces another error because the fix expects a datetime. The model field is a datetime field, but the JS datepicker in the ticket form supplies a date. Easy fix coming in another PR.

PR #1076, opened 2023-03-17
#1075 - make view agnostic to format of incoming due date

finnertysea:
Fix for #1075


uhurusurfa (commented):
Clearly we are missing a unit test that catches this.
Can you add one?

uhurusurfa (approved):
👍  - thanks for the fix. Will look at making a unit test in a separate PR

Issue #1079, opened 2023-03-21
quicktest test_email_with_quoted_printable_body fail


brucegibbins:
Hi. This is possibly more of a question than an issue.

Platform: Windows 11
Locale: Australia
IDE: PyCharm

I am trying to prep my dev environment so that I can run tests and potentially contribute but I am getting an issue when running quicktest. Basically the test fails at line 82 with the following assertion. The reason for the failure is that ticket.description has a trailing '\n' (linefeed) and therefore does not match the assertion test. 

```
        self.assertEqual(ticket.description,
                         "Tohle je test českých písmen odeslaných z gmailu.")
```

Tracing the code leads me to the extract_part_data() function that does a number of  things but after line 749 (below) part_body variable contains the string with the trailing linefeed after extracting the part from the message object. 

`part_body = part.get_payload(decode=True)
`

I am just wondering if this related to some type of Windows Specific management of the carriage return line feed or something to do with utf-decoding which i am not familiar with?

I have tested this within PyCharm so that I could trace line execution but get the same issue from the CLI as well.

Thank-you in advance.

timthelion:
The text says, 'this is a test of czech characters sent from gmail. Nothing to do with linefeeds. I think that the testis still perfectly valid if you just strip the description when passing it to asstert.

brucegibbins:
Hi. Thanks. Yes I agree that the test is still valid. I wasn't actually
concerned about the text itself which in the assertion test does not
include a linefeed. But when the description field is extracted, at least
in my case, it includes a single trailing linefeed which it probably
collected from the test message. Therefore when the comparison occurs they
are identical except for the inclusion of the trailing linefeed after the
message body is extracted.

So I think the test is more than valid. But am trying to isolate why in my
case all of the trailing linefeeds were dropped except for one?

Sorry for the confusion.

brucegibbins:
I have attached a couple of screenshots to show what I mean...

The first image shows the end-of-line characters (in my case \r\n)

![eml](https://user-images.githubusercontent.com/16873240/226559023-0a5f7bd0-28bf-4573-a661-42295b74902e.png)

The second image is what is shown during debug. As you can see there is a trailing linefeed which on the face of it is not a problem except that the test expects that there are no trailing end of line markers.  Which again is logical. But of course the test will fail in this case.

![email](https://user-images.githubusercontent.com/16873240/226560008-1cd96c8f-6a47-4dd4-b56d-da8630903a03.png)

[...]

PR #1082, opened 2023-03-27
Force URTF-8 Encoding on File Open

brucegibbins:
Added encoding argument to open context manager to force utf-8 encoding to address issue raised in #1079


uhurusurfa (approved):
👍

PR #1088, opened 2023-04-15
Add IMAP OAUTH Mailbox Type

brucegibbins:
Have updated email.py to include a new function imap_oauth_sync() that uses OAUTH to authenticate against an IMAP server. This code has only been tested against an O365 Mailbox. There are some settings needed to carry the OAUTH token url, secret and a few other items. These should be setup as a Dict in settings.py.

Two extra packages were required to support the change (oauthlib and requests_oauthlib) . These have been included in requirements.txt 

A migration has also been created to allow for the extra mailbox type.

I am not familiar with how to setup a test for this given the need for credentials. But since it uses the exact same logic to extract the message from the IMAP service (in this case O365) I am guessing the existing test is ok. However, I am guessing the best approach is to somehow Mockup the OAUTH authentication. I am afraid I have no experience do this at this stage.


helpdesk/email.py
```diff
@@ -225,6 +226,101 @@ def imap_sync(q, logger, server):
...
+        oauth = requests_oauthlib.OAuth2Session(client=client)
+        token = oauth.fetch_token(
+            token_url=settings.HELPDESK_OAUTH["token_url"],
+            client_id=settings.HELPDESK_OAUTH["client_id"],
+            client_secret=settings.HELPDESK_OAUTH["secret"],
+            include_client_id=True,
+        )
+
+        # TODO: Somehow link this to the debug level set within Django settings logging
+        server.debug = 4
```
uhurusurfa, line 255: The IMAP library has a multi-level logging setting that does not really correlate to the Django LOG_LEVEL values.
Add a settings entry to allow the user set the debug level for IMAP independently of the Django log level:
`HELPDESK_IMAP_DEBUG_LEVEL = int(os.environ.get("HELPDESK_IMAP_DEBUG_LEVEL", "0"))`

Then change this statement to:
`server.debug = settings.HELPDESK_IMAP_DEBUG_LEVEL`

helpdesk/email.py
```diff
@@ -225,6 +226,101 @@ def imap_sync(q, logger, server):
...
+        server.logout()
+        sys.exit()
+
+    except ssl.SSLError:
+        logger.error(
+            "IMAP login failed due to SSL error. This is often due to a timeout. "
+            "Please check your connection and try again."
+        )
+        server.logout()
+        sys.exit()
```
uhurusurfa, line 277: Change these excpetions to this format so that we get the exceptions reason for the problem rather than a generic and possibly misleading message. Adding **exc_info-True**  will also show the stack trace to aid in debugging:
```
    except imaplib.IMAP4.abort as e1:
        logger.error(f"IMAP authentication failed in OAUTH: {e1}", exc_info=True)
        server.logout()
        sys.exit()
    except ssl.SSLError as e2:
        logger.error(
            f"IMAP login failed due to SSL error. (This is often due to a timeout): {e2}", exc_info=True
        )
        server.logout()
        sys.exit()
```
I am not sure why they used sys.exit() in this module since it crashes the Django app and there is a more gracefull way to handle it so that it might recover but will not crash the entire Django app that Hlpdesk might be integrated into but this can be addressed in a separate PR.

uhurusurfa (changes requested):
Looks good - just some minor changes requested and then I think this is ready to merge.

brucegibbins:
I will take a look tonight
Thanks

helpdesk/tests/test_get_email.py
```diff
@@ -552,258 +420,436 @@ def test_read_email_with_template_tag(self):
...
+# RwcDf9O7MvKSV2kSPmryLnUF7D+2fva+Cy+CvJDVJCo5zr4ucXPXZ4htpI6Pjpd5
+# oPHvbqxSCMJrQ7eAFTYmBNGauSyr0XvGM1qmHBZD/laQEJHYgLT2ILrymZhVDHtK
+# W7tXhGjMoUvqAxiKkmG3UHFqN4k3EYo13PwoOWyJHD1M9ArbX/Sk9l8DDguCh3DW
+# a9eiiQ+3V1v+7wWHXCzq
+# =6JeP
+# -----END PGP SIGNATURE-----
+# """)
+#             # should this be 'application/pgp-signature'?
+#             # self.assertEqual(attach1.mime_type, 'text/plain')
+#
```
Benbb96, line 852: Just to know, why did you commented out these tests ?

brucegibbins:
I was wondering if someone could explain the failure in the test. It seems to be related to the timestamp returned. during the test and thus raising a negative assert as the two values are different.

Also, is there a way I can simply run the test suite for the PR again? I suspect it will pass

test

brucegibbins:
Discovered how to rerun jobs and as suspected they ran sucessfully.

uhurusurfa (approved):
🚀

brucegibbins:
Fantastic. Thanks. Do I need to do anything from here?

PR #1095, opened 2023-06-12
Fix custom commands

Benbb96:
Removes warning when running `manage.py` scripts


Benbb96:
I'm wondering if I should try to fix the `escalate_tickets` command too.

brucegibbins:
Hi @Benbb96. My 10cents worth would be that if you have the time and resource it would be good to see this corrected as well. My experience with HD was that I did seem to come across a number of legacy issues and was never sure if the downstream impact. Thanks again for whay you have already achieved.

PR #1098, opened 2023-06-16
prepare for 0.5.0 release to pypi

gwasser:


gwasser:
seems weird that django 4 on python 3.9 failed but works on other pythons. some kind of internal build error?

gwasser:
ok all passed that time, guess a weird internal github error. passes for all python versions, I added the latest python 3.11 to the testing. need a review to get it merged. can you do so @uhurusurfa ? i guess it requires a review from someone that isn't the submitter?

Benbb96 (approved):
For me, it all seems good

PR #1099, opened 2023-06-16
move development to minimum of django 4.2 LTS

gwasser:
Updating version numbers and README text so that development proceeds on Django 4.2 now that it has reached LTS status. Django is designed to easily migrate from LTS to LTS so time to wind down 3.2 support in favor of 4.2.


uhurusurfa:
@gwasser Since Django 3 has support for at least another year I am not sure it is the right time to stop supporting it.
Remember that it took quite a bit of effrot for me to get the dependent projects upgraded to Django 4 before we could properly upgrade Helpdesk.
We could start work on supporting Django 5 now that it is officially released.

gwasser:
Yeah I suppose I didn't realize 3.2 was going to be supported for that much longer so I suppose no need to rush it. Adding Django 5 support is a good thing to start on though.

PR #1116, opened 2023-09-02
Use SECURE_SSL_REDIRECT as a default value for HELPDESK_USE_HTTPS_IN_EMAIL_LINK setting

Benbb96:
Following discussion #1115, I believe we could use the setting [SECURE_SSL_REDIRECT](https://docs.djangoproject.com/fr/4.2/ref/settings/#std-setting-SECURE_SSL_REDIRECT) as a way to know if HTTPS has been set up in the project, and thus avoid any surprise in outbound emails.


brucegibbins (approved):
Makes sense to me. Cheers

uhurusurfa (approved):
🚀

Issue #826, opened 2020-05-19
Provide a Dockerfile to help config a setup plz


danijeljw:
**Is your feature request related to a problem? Please describe.**
I am not a django user, but would like to see this spun up inside a Docker container. Can you please adda Dockerfile to the project to use?

**Describe the solution you'd like**
Having a pre-configured Dockerfile to see what the install instructions mean. I didn't understand them.

**Describe alternatives you've considered**
I have tried to follow your steps. I don't undersatnd.

**Additional context**
Having a Dockerfile would give me visibility on how to setup the project.

gwasser:
This is a great idea. I'm not super familiar with Docker so I invite folks to submit a PR with a dockerfile. Might be better to use this eventually than the bash testing script currently in use.

danijeljw:
@gwasser Can you provide me the bash script you use to setup this project? I would be happy to look at how to create a Dockerfile for you as I'm better with that.

I just need to see what steps - even just the basics - put together to setup this project and I will look into it for you.

timthelion:
Have you found this README and associated Makefile? 

https://github.com/django-helpdesk/django-helpdesk/blob/master/demo/README.rst

https://github.com/django-helpdesk/django-helpdesk/blob/master/Makefile

pupattan:
This will be helpful

vzankov:
https://github.com/clwluvw/Django-HelpDesk

PR #1117, opened 2023-09-07
dockerize, fixes #826

simevo:
just a quick Dockerfile

PR #1119, opened 2023-09-25
Standalone helpdesk installation configuration

timthelion:
Fixes https://github.com/django-helpdesk/django-helpdesk/issues/826

You can easily configure themes and alternate email senders such as SES. This is documented in the new standalone Readme.


timthelion:
I see another helpful contributor was working on this at the same time I was. I believe my version is better though because it has more production ready features such as using gunicorn as well as support for receiving email using a cron job...

https://github.com/django-helpdesk/django-helpdesk/pull/1117/files

Benbb96:
I didn't have time to review it but I hope I'll be able to do it soon..

Maybe you can also update the README like on the other PR, in order to show the commands to use the docker-compose ?

standalone/Caddyfile
```diff
@@ -0,0 +1,3 @@
+localhost {
+        reverse_proxy helpdesk:8000
```
Benbb96, line 2:
```suggestion
        reverse_proxy django-helpdesk:8000
```
I had to change this to make it match the container name in the docker-compose file to make it work

standalone/docker-compose.yml
```diff
@@ -0,0 +1,28 @@
...
+ django-helpdesk:
+   build:
+    context: ..
+    dockerfile: standalone/Dockerfile
+   user: root
+   volumes:
+    - /tmp/django-helpdesk-data:/data/
+    - ./custom_navigation_header.html:/opt/django-helpdesk/helpdesk/templates/helpdesk/custom_navigation_header.html:r
+   env_file: docker.env
+
```
Benbb96, line 23:
```suggestion
   env_file: docker.env
   depends_on:
     - postgres

```
In order to be sure db is ready to run migrations

Benbb96 (commented):
I encountered some errors while trying on Windows (setup.sh used wrong end of line but that could be due of PyCharm), then I tested on WSL Debian and it worked (after fixing Caddyfile).

It seems good for me. Only media files didn't work for me. But I didn't dig too much to make them work.

timthelion:
What did you have to do to fix the Caddyfile?

standalone/Caddyfile
Benbb96: @timthelion Here I had to make it match the name of the container from the `docker-compose.yml`

Benbb96 (approved):

helpdesk/models.py
```diff
@@ -207,7 +207,7 @@ class Queue(models.Model):
     email_box_ssl = models.BooleanField(
         _('Use SSL for E-Mail?'),
         blank=True,
-        default=False,
+        default=True,
```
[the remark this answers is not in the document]
uhurusurfa: This probably will require a migration.

uhurusurfa:
I have a feling the expunge call is messing with the unit tests.

timthelion:
Interesting thought, just reverted that to see if it helps...

timthelion:
Turns out it was that bad default.
We should still change the default though...

timthelion:
But in another PR

timthelion:
I don't know if you noticed but I also added a crappy icon ;) So now we have a logo...

Benbb96 (approved):

PR #1124, opened 2023-10-17
Fix PyPi release script. Support ReadTheDocs automated release.

uhurusurfa:
Since the release is publsihed in the UI and is more easily managed with appropriate documentation there is no need for doing the release build into Githiub so that has been removed from the PyPi release.
Added the .readthedocs.yaml to enable automation of the documentation update to readthedocs when a release is made.


brucegibbins (approved):
LGTM

PR #1128, opened 2023-10-24
Catch exceptions on precess email loop

uhurusurfa:
Allows all queues to be processed even if there is an unexpected exception in proceessing a queue.
Current implementation will crash the moment any queue has an unexpected processing exception.

Also adds an additional invocation flag to the get_email management command to allow seeing additional information when running from a terminal window to more easily debug the problem.

### DIFF
--- a/helpdesk/email.py
+++ b/helpdesk/email.py
@@ -57,17 +57,16 @@ STRIPPED_SUBJECT_STRINGS = [
 HTML_EMAIL_ATTACHMENT_FILENAME = _("email_html_body.html")
 
 
-def process_email(quiet: bool =False, debug_to_stdout=False):
+def process_email(quiet: bool = False, debug_to_stdout: bool = False):
     if debug_to_stdout:
         print("Extracting email into queues...")
-    q: Queue() # Typing ahead of time for loop to make it more useful in an IDE
+    q: Queue()  # Typing ahead of time for loop to make it more useful in an IDE
     for q in Queue.objects.filter(
             email_box_type__isnull=False,
             allow_email_submission=True):
         log_msg = f"Processing queue: {q.slug} Email address: {q.email_address}..."
         if debug_to_stdout:
-            print(log_msg )
-            
+            print(log_msg)
         logger = logging.getLogger('django.helpdesk.queue.' + q.slug)
         logging_types = {
             'info': logging.INFO,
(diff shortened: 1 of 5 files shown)


helpdesk/email.py
```diff
@@ -84,16 +90,24 @@ def process_email(quiet=False):
...
+        try:
             queue_time_delta = timedelta(minutes=q.email_box_interval or 0)
             if (q.email_box_last_check + queue_time_delta) < timezone.now():
                 process_queue(q, logger=logger)
                 q.email_box_last_check = timezone.now()
                 q.save()
+                log_msg: str = f"Queue successfully processed: {q.slug}"
+                logger.info(log_msg)
+                if debug_to_stdout:
+                    print(log_msg)
```
timthelion, line 104: Doesn't this mean that if the django logger is configured to log to console then the message will be output twice?

helpdesk/email.py
```diff
@@ -56,11 +57,16 @@
...
+        print("Extracting email into queues...")
+    q: Queue()  # Typing ahead of time for loop to make it more useful in an IDE
     for q in Queue.objects.filter(
             email_box_type__isnull=False,
             allow_email_submission=True):
-
+        log_msg = f"Processing queue: {q.slug} Email address: {q.email_address}..."
+        if debug_to_stdout:
+            print(log_msg)
         logger = logging.getLogger('django.helpdesk.queue.' + q.slug)
```
timthelion, line 70: I would prefer to configure this logger here to output to stdout if the flag is set, then duplicate the lines.

helpdesk/email.py
uhurusurfa: It would but this is meant for production debugging and in production iot makes no sense at all to have console logging enabled unless you are redirecting console logs to some other persistant log store.
This mechanism is intended for quick and easy debugging in a production setup and is intended not to depend on any production configuration at all.

helpdesk/email.py
uhurusurfa: I could but from a code perspective this is much cleaner as the debug_to_stdout flag is specifically only for manually running the 
The same arguments apply here as for the earlier comment regarding why this is treated separately.

helpdesk/email.py
```diff
@@ -84,16 +90,24 @@ def process_email(quiet=False):
...
             if (q.email_box_last_check + queue_time_delta) < timezone.now():
                 process_queue(q, logger=logger)
                 q.email_box_last_check = timezone.now()
                 q.save()
+                log_msg: str = f"Queue successfully processed: {q.slug}"
+                logger.info(log_msg)
+                if debug_to_stdout:
+                    print(log_msg)
+        except Exception:
+            logger.error("Queue processing failed: {q.slug} -- {e}", exc_info=True)
```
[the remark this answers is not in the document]
uhurusurfa: Well caught - thanks. Change made

helpdesk/email.py
```diff
@@ -84,16 +90,24 @@ def process_email(quiet=False):
...
             if (q.email_box_last_check + queue_time_delta) < timezone.now():
                 process_queue(q, logger=logger)
                 q.email_box_last_check = timezone.now()
                 q.save()
+                log_msg: str = f"Queue successfully processed: {q.slug}"
+                logger.info(log_msg)
+                if debug_to_stdout:
+                    print(log_msg)
+        except Exception as e:
+            logger.error("Queue processing failed: {q.slug} -- {e}", exc_info=True)
```
Benbb96, line 106: Does this work without the `f` at the beginning of the string?

uhurusurfa: Good catch. I have fixed it and added unit tests

Benbb96 (approved):

PR #1131, opened 2023-11-09
changed UserSerializer to utilize custom user classes (didn't work wi…

Naggafin:
The `UserSerializer` class under `serializers.py` for REST framework APIs does not work for custom user classes because it was specified only to work with the Django User model. If a custom User class is specified, this can cause an OperationalError given that the table for the Django User model does not exist. I simply updated the code to use `get_user_model()` from the Django auth package so that it always returns an accurate reading of what model is in use for a User type.


uhurusurfa (approved):
🚀 Thanks foir the contribution.

PR #1140, opened 2023-11-15
Simplify teams enable/disable and fix unassigned kbitems not visible when teams not active

uhurusurfa:
Make teams enable disable a simple boolean flag.
Fix error when teams is disabled and unassigned tickets are attempted to be retrieved.


demo/demodesk/config/settings.py
```diff
@@ -39,9 +33,10 @@
 # an internal demo you don't need such security, but please
 # remember when setting up your own development / production server!
 
+# Default teams mode to enabled unless overridden by an environment variable set to "false"
+HELPDESK_TEAMS_MODE_ENABLED=False if "false" == os.getenv("HELPDESK_TEAMS_MODE_ENABLED", "true").lower() else True
```
Benbb96, line 37: I would invert the condition to make it more readable:

```suggestion
HELPDESK_TEAMS_MODE_ENABLED=False if os.getenv("HELPDESK_TEAMS_MODE_ENABLED", "true").lower() == "false" else True
```

Benbb96: And why not even do it like this:

```suggestion
HELPDESK_TEAMS_MODE_ENABLED=os.getenv("HELPDESK_TEAMS_MODE_ENABLED", "true").lower() == "true"
```

docs/teams.rst
```diff
@@ -29,10 +29,23 @@ If you want to use a different team app or implement your own team based app, yo
...
 
 
-Disabling Teams Functionality
+Configuring Teams Functionality
 -----------------------------
-If you do not wish to use team functionality, you can disable teams by setting the following settings:
+Teams functionality is enabled by default but can be disabled using this entry in your ``settings.py``::
+      HELPDESK_TEAMS_MODE_ENABLED=False 
+
+If you do not disable teams functionality then you must add additional apps into the ``INSTALLED_APPS`` tuple in your ``settings.py``use team functionality, you can disable teams by setting the following settings:
```
Benbb96, line 37:
```suggestion
If you do not disable teams functionality then you must add additional apps into the ``INSTALLED_APPS`` tuple in your ``settings.py``.
```
I end of sentence doesn't make sense to me

helpdesk/settings.py
```diff
@@ -231,12 +228,24 @@
 HELPDESK_USE_HTTPS_IN_EMAIL_LINK = getattr(
     settings, 'HELPDESK_USE_HTTPS_IN_EMAIL_LINK', settings.SECURE_SSL_REDIRECT)
 
-HELPDESK_TEAMS_MODEL = getattr(
-    settings, 'HELPDESK_TEAMS_MODEL', 'pinax_teams.Team')
-HELPDESK_TEAMS_MIGRATION_DEPENDENCIES = getattr(settings, 'HELPDESK_TEAMS_MIGRATION_DEPENDENCIES', [
+# Default to True for backwards compatibility
+HELPDESK_TEAMS_MODE_ENABLED = False if hasattr(settings, 'HELPDESK_TEAMS_MODE_ENABLED') and getattr(
+    settings, 'HELPDESK_TEAMS_MODE_ENABLED', True) is False else True
```
Benbb96, line 233: Same, we can make it more readable:

```suggestion
HELPDESK_TEAMS_MODE_ENABLED = getattr(settings, 'HELPDESK_TEAMS_MODE_ENABLED', True)
```

helpdesk/settings.py
```diff
@@ -231,12 +228,24 @@
...
+    HELPDESK_KBITEM_TEAM_GETTER = getattr(
+        settings, 'HELPDESK_KBITEM_TEAM_GETTER', lambda kbitem: kbitem.team)
+else:
+    HELPDESK_TEAMS_MODEL = settings.AUTH_USER_MODEL
+    HELPDESK_TEAMS_MIGRATION_DEPENDENCIES = []
+    HELPDESK_KBITEM_TEAM_GETTER = lambda _: None
+
+# show knowledgebase links?
+# If Teams mode is enabled then it has to be on
+HELPDESK_KB_ENABLED = True if HELPDESK_TEAMS_MODE_ENABLED else getattr(settings, 'HELPDESK_KB_ENABLED', True)
```
Benbb96, line 248: Isn't it possible to have Teams mode enabled and not wanting to have KB links enabled as well?

Benbb96 (changes requested):
I see a test unrelated to Teams mode, that would be great to have a test in order to validate the new setting `HELPDESK_TEAMS_MODE_ENABLED`

uhurusurfa:
> I see a test unrelated to Teams mode, that would be great to have a test in order to validate the new setting `HELPDESK_TEAMS_MODE_ENABLED`

@Benbb96  Agreed but I do not have the time right now to create a test. I will aim to add one when I get time.

helpdesk/settings.py
uhurusurfa: I have not looked into it too deeply but the HELPDESK_KB_ENABLED setting seems to deeply disable KB functionality which teams relies on because of the way it is implemented and relies on being able to set up and mange KB items.
@timthelion - do you know unequivocally if HELPDESK_KB_ENABLED must be True for teams to work?

timthelion:
I would persnally not auto-enable kb if teams are enabled because in the future we might add teams support on a per -queue basis and then this auto-enabling of kb would not make sense and could be super confusing.

uhurusurfa:
So my understanding from "**With KB dissabled teams are more or less a noop I think. They neither do anything nor harm anything.."** means that if Teams is enabled and KB is not then teams will not work which supports the configurtion forcing KB to be enabled if teams is enabled.

Whether or not a future change is made to how teams operates is not relevant right now and we can revisit this when and if teams functionality is enhanced or channged in the future.

uhurusurfa:
@Benbb96  - can you approve please?

demo/demodesk/config/settings.py
```diff
@@ -39,9 +33,10 @@
 # an internal demo you don't need such security, but please
 # remember when setting up your own development / production server!
 
+# Default teams mode to enabled unless overridden by an environment variable set to "false"
+HELPDESK_TEAMS_MODE_ENABLED=False if os.getenv("HELPDESK_TEAMS_MODE_ENABLED", "true").lower() == "false" else True
```
Benbb96, line 37: I think we can explicitly choose to use Teams in the demodesk in order to show that feature
```suggestion
HELPDESK_TEAMS_MODE_ENABLED=True
```

Benbb96: Sorry, I read the comment above after. We can keep that but simplify it like this:
```suggestion
HELPDESK_TEAMS_MODE_ENABLED=os.getenv("HELPDESK_TEAMS_MODE_ENABLED", "true").lower() == "true"
```

docs/teams.rst
```diff
@@ -29,10 +29,23 @@ If you want to use a different team app or implement your own team based app, yo
...
+        'account',  # Required by pinax-teams
+        'pinax.invitations',  # required by pinax-teams
+        'pinax.teams',  # team support
+        'reversion',  # required by pinax-teams
+    ])
 
-            HELPDESK_TEAMS_MODEL='auth.User',
-            HELPDESK_TEAMS_MIGRATION_DEPENDENCIES=[],
-            HELPDESK_KBITEM_TEAM_GETTER=lambda _: None,
+Alternatively just add the 4 apps listed above into the INSTALLED_APPS tuple.
```
Benbb96, line 46: INSTALLED_APPS can be a list too
```suggestion
Alternatively just add the 4 apps listed above into the ``INSTALLED_APPS``.
```

docs/teams.rst
```diff
@@ -29,10 +29,23 @@ If you want to use a different team app or implement your own team based app, yo
...
 
 
-Disabling Teams Functionality
+Configuring Teams Functionality
 -----------------------------
-If you do not wish to use team functionality, you can disable teams by setting the following settings:
+Teams functionality is enabled by default but can be disabled using this entry in your ``settings.py``::
+      HELPDESK_TEAMS_MODE_ENABLED=False 
+
+If you do not disable teams functionality then you must add additional apps into the ``INSTALLED_APPS`` tuple in your ``settings.py``.
```
Benbb96, line 37:
```suggestion
If you do not disable teams functionality then you must add additional apps into the ``INSTALLED_APPS`` in your ``settings.py``.
```

docs/teams.rst
```diff
@@ -29,10 +29,23 @@ If you want to use a different team app or implement your own team based app, yo
...
+The following can be pasted into your settings.py BELOW the INSTALLED_APPS tuple definition::
```
Benbb96, line 38:
```suggestion
The following can be pasted into your settings.py BELOW the INSTALLED_APPS definition::
```

docs/install.rst
```diff
@@ -52,30 +53,36 @@ If you're on a brand new Django installation, make sure you do a ``migrate``
...
-
-   Note: you do not need to use pinax-teams. To disable teams see the :doc:`teams` section.
-
-   Your ``settings.py`` file should also define a ``SITE_ID`` that allows multiple projects to share
-   a single database, and is required by ``django.contrib.sites`` in Django 1.9+.
-   If you aren't running multiple sites, you can simply add a default ``SITE_ID`` to ``settings.py``::
+1. Edit your ``settings.py`` file add the following entries:
+   - add ``helpdesk`` to the ``INSTALLED_APPS`` tuple along with some other required entries in the ``django.contrib`` package.
+
+     An example of the core  INSTALLED_APPS requirements for this app are shown below::
```
Benbb96, line 59:
```suggestion
   - add ``helpdesk`` to the ``INSTALLED_APPS`` along with some other required entries in the ``django.contrib`` package.

     An example of the core  ``INSTALLED_APPS`` requirements for this app are shown below::
```

Benbb96 (approved):

PR #1143, opened 2023-11-23
Add my tickets view for logged in non-staff users

timthelion:
image

This includes a breaking change as I forced pagination on the API.

Implements: https://github.com/django-helpdesk/django-helpdesk/pull/323

I also ended up switching to class based views for the public views.


uhurusurfa (commented):
Code looks good to me - giv en there are no unit tests to verify the functionality does what it should I will have to find time toi manually test this ... hopefully on the weekend.

timthelion:
I added unit tests for the API endpoint ;)

helpdesk/templates/helpdesk/navigation-sidebar.html
```diff
@@ -68,6 +68,15 @@
...
         </li>
+        {% if user.is_authenticated %}
+        <li class="nav-item{% if 'my-tickets' in request.path %} active{% endif %}">
+            <a class="nav-link" href="{% url 'helpdesk:my-tickets' %}">
+                <i class="fas fa-fw fa-tasks"></i>
+                <span>{% trans "My Tickets" %}</span>
+            </a>
+        </li>
+
+        {% endif %}
```
uhurusurfa, line 79: You will have to add this to both parts of the if/then otherwise if someone sets `HELPDESK_NAVIGATION_ENABLED` to Tue or the app is used internally by the company to provide a helpdesk and someone is an admin, they cannot use the "My Tickets" view.

uhurusurfa (changes requested):
Seems to work as described - just the one consideration noted.

helpdesk/templates/helpdesk/navigation-sidebar.html
timthelion: Added above.

uhurusurfa (approved):

PR #1145, opened 2023-11-29
API: Actually send out emails when creating followups via the API

timthelion:


docs/integration.rst
```diff
@@ -1,6 +1,9 @@
 Integrating django-helpdesk into your application
 -------------------------------------------------
```
Benbb96, line 2: Maybe make this line with equals to mark a difference between title and subtitle

Benbb96 (approved):
Nice refactoring

PR #1146, opened 2023-12-02
Webhooks

timthelion:
This PR implements the ability to add webhooks which get notified when tickets are created and followups are added. I am implementing this for a client who needs to create bots that can automatically interact with the helpdesk. https://github.com/django-helpdesk/django-helpdesk/pull/1145 needs to be merged before this one as the two PRs build on each-other.

### DIFF
--- a/docs/webhooks.rst
+++ b/docs/webhooks.rst
@@ -3,8 +3,10 @@ Registering webhooks to be notified of helpesk events
 
 You can register webhooks to allow third party apps to be notified of helpdesk events. Webhooks can be registered in one of two ways:
 
-1. Setting comma separated environement variables; ``HELPDESK_NEW_TICKET_WEBHOOK_URLS``& ``HELPDESK_FOLLOWUP_WEBHOOK_URLS``.
+1. Setting the following environement variables to a comma separated list of URLs; ``HELPDESK_NEW_TICKET_WEBHOOK_URLS``& ``HELPDESK_FOLLOWUP_WEBHOOK_URLS``.
 
 2. Adding getter functions to your ``settings.py``. These should return a list of strings (urls); ``HELPDESK_GET_NEW_TICKET_WEBHOOK_URLS`` & ``HELPDESK_GET_FOLLOWUP_WEBHOOK_URLS``.
+
+3. You can optionally set ``HELPDESK_WEBHOOK_TIMEOUT`` which defaults to 3 seconds. Warning, however, webhook requests are sent out sychronously on ticket update. If your webhook handling server is too slow, you should fix this rather than causing helpdesk freezes by messing with this variable.
 
 Once these URLs are configured, a serialized copy of the ticket object will be posted to each of these URLs each time a ticket is created or followed up on respectively.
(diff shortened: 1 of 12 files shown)


helpdesk/settings.py
```diff
@@ -277,3 +277,19 @@
 # Attachment directories should be created with permission 755 (rwxr-xr-x)
 # Override it in your own Django settings.py
 HELPDESK_ATTACHMENT_DIR_PERMS = int(getattr(settings, 'HELPDESK_ATTACHMENT_DIR_PERMS', "755"), 8)
+
+def get_followup_webhook_urls():
+    urls = os.environ.get('HELPDESK_FOLLOWUP_WEBHOOK_URLS', None)
+    if urls:
+        return urls.split(',')
```
uhurusurfa, line 284: Might be a good idea to cater for spaces before and after the comma here and in the new ticket function:
`return re_split(r"[\s],[\s]", urls)`

docs/webhooks.rst
```diff
@@ -0,0 +1,10 @@
+Registering webhooks to be notified of helpesk events
+-----------------------------------------------------
+
+You can register webhooks to allow third party apps to be notified of helpdesk events. Webhooks can be registered in one of two ways:
+
+1. Setting comma separated environement variables; ``HELPDESK_NEW_TICKET_WEBHOOK_URLS``& ``HELPDESK_FOLLOWUP_WEBHOOK_URLS``.
+
+2. Adding getter functions to your ``settings.py``. These should return a list of strings (urls); ``HELPDESK_GET_NEW_TICKET_WEBHOOK_URLS`` & ``HELPDESK_GET_FOLLOWUP_WEBHOOK_URLS``.
+
```
uhurusurfa, line 9: Would be good to add the HELPDESK_WEBHOOK_TIMEOUT setting description here as well.

docs/webhooks.rst
```diff
@@ -0,0 +1,10 @@
+Registering webhooks to be notified of helpesk events
+-----------------------------------------------------
+
+You can register webhooks to allow third party apps to be notified of helpdesk events. Webhooks can be registered in one of two ways:
+
+1. Setting comma separated environement variables; ``HELPDESK_NEW_TICKET_WEBHOOK_URLS``& ``HELPDESK_FOLLOWUP_WEBHOOK_URLS``.
```
uhurusurfa, line 6: This statement is unclear what it means. It would be best to say a comma separate list of webhook URL's for each of the environment variables.
If you do not cater for a space in the URL list then it should be stated here that there must be no spaces before or after commas.

uhurusurfa (changes requested):
Looks good overall.
Just some minor suggestions around the documentation and handling the split of URL's.

If possible, it wouyld be good if there was a test that verified it works for multiple URL's for a new or followup ticket.

timthelion:
Thank you for the thorough review. Updated.

Benbb96 (approved):
Looks good me too.  
Good job, and also well done for the triage of issues 🙂

uhurusurfa (approved):
🎸

PR #1150, opened 2023-12-27
WIP: Fix custom field sending in new ticket webhook

timthelion:
There was a race condition with custom field value object creation where the webhook was being called before the custom values were set in the DB. If anyone has a better idea on how to fix this race then please speak up. This seems to work, but I'm still testing thus the WIP status.

### DIFF
--- a/helpdesk/tests/test_webhooks.py
+++ b/helpdesk/tests/test_webhooks.py
@@ -1,5 +1,6 @@
 from django.contrib.auth.models import User
-from helpdesk.models import Queue
+from helpdesk.models import Queue, CustomField, TicketCustomFieldValue, Ticket
+from helpdesk.serializers import TicketSerializer
 from rest_framework.status import (
     HTTP_201_CREATED
 )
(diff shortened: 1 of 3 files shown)


timthelion:
This seems to work, so can be merged now.

helpdesk/tests/test_webhooks.py
```diff
@@ -116,6 +124,35 @@ def test_create_ticket_and_followup_via_api(self):
...
+             'status',
+             'submitter_email',
+             'title']
+        )
+        with open("/tmp/request", "w") as fd:
+            import json
+            json.dump(handled_webhook_requests['new_ticket_requests_1'][-1]["ticket"], fd, indent=4, default=str)
+        with open("/tmp/serilaizer_data", "w") as fd:
+            import json
+            json.dump(serializer.data, fd, indent=4, default=str)
```
Benbb96, line 153: Were these lines useful for debugging? I don't see what they bring to the test

Benbb96 (approved):

helpdesk/tests/test_webhooks.py
timthelion: Oops. sorry about that, will remove. Yes I was trying to figure out what was going on...

uhurusurfa: Will merge when the extra code is removed

timthelion:
Oops, should be fixed now.

Benbb96 (approved):

PR #1153, opened 2024-01-22
Customization of Ticket PRIORITY_CHOICES in project settings

samsplunks:
I wish to have a different set of priorities for tickets.

This is done by adding HELPDESK_TICKET_PRIORITY_CHOICES to the project settings.

If the parameter is not set, it defaults to the official settings.

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -176,6 +176,21 @@ Options that change ticket updates
   **Default:** ``HELPDESK_STAFF_ONLY_TICKET_CC = False``
 
 
+Options that change ticket properties
+-------------------------------------
+
+- **HELPDESK_TICKET_PRIORITY_CHOICES** Customize the priority choices for all tickets.
+
+  The **default** is below::
+
+        HELPDESK_TICKET_PRIORITY_CHOICES = (
+            (1, _('1. Critical')),
+            (2, _('2. High')),
+            (3, _('3. Normal')),
+            (4, _('4. Low')),
+            (5, _('5. Very Low')),
+        )
+
 Staff Ticket Creation Settings
 ------------------------------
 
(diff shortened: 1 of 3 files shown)


helpdesk/settings.py
```diff
@@ -6,6 +6,7 @@
 from django import forms
 from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
+from django.utils.translation import gettext, gettext_lazy as _
```
Benbb96, line 9: You have an unused import which creates an error in the CI

```suggestion
from django.utils.translation import gettext_lazy as _
```

Benbb96 (commented):
Do you think you could update the documentation (the settings page) ? In order to give a small explanation and an example for this new setting

helpdesk/settings.py
samsplunks: Thanks for the quick feedback and good catch for the import.

I will update the docs as requested.

samsplunks:
I hope I comply with your request for the new commits.
Feel free to comment as it is my first contribution.

I have added an **Options that change ticket properties** section as I will propose more customization there.

docs/settings.rst
```diff
@@ -176,6 +176,32 @@ Options that change ticket updates
...
+  If you have a new instance, you may override those settings but if you want to keep previous tickets priorities and add new choices, you may increment integer values like this::
+
+        HELPDESK_TICKET_PRIORITY_CHOICES = (
+            (1, _('1. Critical')),
+            (2, _('2. High')),
+            (3, _('3. Normal')),
+            (4, _('4. Low')),
+            (5, _('5. Very Low')),
+            (6, _('1. Cold')),
+            (7, _('2. Hot')),
```
Benbb96, line 202:
```suggestion
            (6, _('6. Cold')),
            (7, _('7. Hot')),
```

Benbb96 (commented):
Looks good to me! Only one thing in the documentation to fix (with the labels not matching their ids), and I believe we can merge :wink:

samsplunks:
Thank you.

uhurusurfa (approved):
🚀

PR #1156, opened 2024-01-31
Implement Custom delete_queryset Method in QueueAdmin to Resolve Django 1062 Error

Ste-KNY:
This Pull Request implements a custom delete_queryset method within the QueueAdmin class to address the Django 1062 Duplicate Entry error. By individually deleting each Queue instance within a queryset, this method ensures that any custom deletion logic defined in the Queue model's delete method is executed. This approach effectively resolves issues related to the deletion of instances that have unique constraints, thus preventing the Django 1062 error.

### DIFF
--- a/helpdesk/admin.py
+++ b/helpdesk/admin.py
@@ -37,6 +37,10 @@ class QueueAdmin(admin.ModelAdmin):
         else:
             return "-"
 
+    def delete_queryset(self, request, queryset):
+        for queue in queryset:
+            queue.delete()
+
 
 @admin.register(Ticket)
 class TicketAdmin(admin.ModelAdmin):


Benbb96:
Hi @Ste-KNY, thanks for your contribution. I don't understand why deleting a queryset would result in an integrity error? Do you have an example of the error, in order to know from where it is coming?

I found  a discussion on https://github.com/jazzband/django-simple-history/issues/391 which might be related. Maybe we need to better handle db constraint to avoid this error

Ste-KNY:
Hi @Benbb96, thank you for your fast response.

The issue arises when a queue, once deleted, is recreated using the same name. The custom delete method in the Queue model class isn't triggered during deletion via the admin panel. This results in the following error:
`
(1062, "Duplicate entry '204-queue_access_test' for key 'auth_permission_content_type_id_codename_01ab375a_uniq'")
`

Benbb96:
Oh yes, good catch, I understand now. For reference, [here is the custom delete method](https://github.com/django-helpdesk/django-helpdesk/blob/147151a271da818c2a9fdd7c7eed960086bd4167/helpdesk/models.py#L432).

I guess we don't have other ways than triggering the delete method for each queue in the queryset like you're suggesting.

Benbb96 (approved):

timthelion (approved):

PR #1161, opened 2024-02-18
Add some convenient functions for protecting views in custom installations

timthelion:
Not sure if this needs to be merged into the main project but it sure would be convenient for me :) . Client needed to paywall the ticketing system and this was the easiest way ;)


Benbb96 (changes requested):
I think it makes sense to add these possible checks in decorators. Could you just add the related documentation in order to show how they might be used, please?

The setting `VALID_EXTENSIONS` is documented but isn't present in `settings.py` so I believe it should be removed in profit for your `HELPDESK_VALIDATE_ATTACHMENT_TYPES`.

The tests also need to be adapted for this new setting I think.

timthelion:
Oops, I didn't realize that `VALID_EXTENSIONS` was documented. I'll update the docs, but not this week ;) No time...

Benbb96 (dismissed):
All good for me!

timthelion:
@uhurusurfa Can you merge this please?

docs/settings.rst
```diff
@@ -85,18 +103,10 @@ These changes are visible throughout django-helpdesk
 
   **Default:** ``HELPDESK_MAX_EMAIL_ATTACHMENT_SIZE = 512000``
 
-- **VALID_EXTENSIONS** Valid extensions for file types that can be attached to tickets
-
-  **Default:** ``VALID_EXTENSIONS = ['.txt', '.asc', '.htm', '.html', '.pdf', '.doc', '.docx', '.odt', '.jpg', '.png', '.eml']
-
```
uhurusurfa, line 91: Mark this as deprecated but continue to support it until some futiure release to maintain backwards compatibility.

helpdesk/settings.py
```diff
@@ -329,6 +337,10 @@
 # Override it in your own Django settings.py
 HELPDESK_ATTACHMENT_DIR_PERMS = int(getattr(settings, 'HELPDESK_ATTACHMENT_DIR_PERMS', "755"), 8)
 
+HELPDESK_VALID_EXTENSIONS = getattr(settings, 'HELPDESK_VALID_EXTENSIONS', ['.txt', '.asc', '.htm', '.html', '.pdf', '.doc', '.docx', '.odt', '.jpg', '.png', '.eml'])
+
+HELPDESK_VALIDATE_ATTACHMENT_TYPES = getattr(settings, 'HELPDESK_VALIDATE_ATTACHMENT_TYPES', True)
+
```
uhurusurfa, line 343: Use `VALID_EXTENSIONS` if it exists and fall back to `HELPDESK_VALID_EXTENSIONS` if empty to maintain backwards compatibility.

Failing there will ned to be clear upgrade notes telling users to change the name when they upgrade.

uhurusurfa (changes requested):
See inline comments

helpdesk/views/api.py
```diff
@@ -34,6 +36,10 @@ def get_queryset(self):
 class TicketViewSet(viewsets.ModelViewSet):
     """
     A viewset that provides the standard actions to handle Ticket
+
+    You can filter the tickets by status using the `status` query parameter. For example:
+
+    `/api/tickets/?status=Open,Resolved` will return all the tickets that are Open or Resolved.
```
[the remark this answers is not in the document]
timthelion: Somehow the idea of passing arbitrary numbers just felt less nice than having the names but you're right about the localization issue so I'll change it.

uhurusurfa (approved):

PR #1165, opened 2024-03-12
Fix `make rundemo` and Documentation files

### LINKED ISSUE #1163: Failed to install demo using both 'make rundemo' and docker variants
**Describe the bug**
Once make command had done its job it throws the exception `ModuleNotFoundError` see screenshort 

**To Reproduce**
1. Download help-desk-package from github
2.  Extract the data into folder then go to it
3.  Execute `make rundemo` command  from CLI 

If I use docker variant I get the same error. It seems like it has lack of dependecies named demo.
By the way.
If use the command `docker build . -t demodesk docker run --rm -v "$PWD:/app" -p 8080:8080 demodesk` from the docs it doesn't work because of error
image

 -  Ubuntu 20.04 LTS via WSL2 of Windows 11

image

Benbb96:
Fix #1163


uhurusurfa (approved):
Looks good 👍

PR #1167, opened 2024-04-08
Allow to track queue change in follow-ups

samsplunks:
Because of Time calculation feature that allows for time in specific queues not to be counted in, we need to track queue changes in TicketChanges.

As ticket changes are related to follow-ups, I have added the ability to change ticket queue in the "Change Further Details" of the ticket view.

If this is approved, it will enable me to code correct calculation for the FOLLOWUP_TIME_SPENT_EXCLUDE_QUEUES setting.

In my use case, queues are used to pass tickets to different teams.

### DIFF
--- a/helpdesk/update_ticket.py
+++ b/helpdesk/update_ticket.py
@@ -306,13 +306,11 @@ def update_ticket(
         ticket.priority = priority
 
     if queue != ticket.queue.id:
-        c = TicketChange(
-            followup=f,
+        c = f.ticketchange_set.create(
             field=_('Queue'),
             old_value=ticket.queue.id,
             new_value=queue,
         )
-        c.save()
         ticket.queue_id = queue
 
     if due_date != ticket.due_date:
--- a/helpdesk/update_ticket.py
+++ b/helpdesk/update_ticket.py
@@ -12,6 +12,7 @@ from helpdesk.decorators import (
     is_helpdesk_staff,
 )
 from helpdesk.models import (
+    Queue,
     FollowUp,
     Ticket,
     TicketCC,
@@ -200,6 +201,7 @@ def update_ticket(
         owner=-1,
         ticket_title=None,
         priority=-1,
+        queue=-1,
         new_status=None,
         time_spent=None,
         due_date=None,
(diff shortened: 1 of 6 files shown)


uhurusurfa (commented):
This change seems reasonable to me.
Could you add a small unit test for this additional feature (maybe in helpdesk/tests/test_ticket_actions.py)?

samsplunks:
Thanks, yes I will and I will also update the time calculation code relative to each follow-up.

The idea is that if a helpdesk admin wants to introduce time_spent auto calculation or wish to modify the calculation settings, he can make a queryset with all FollowUps, remove time_spent and .save() to (re)calculate the time.

samsplunks:
I have refactored the time_spent_calculation so it doesn't make unnecessary queries when no exclusion is set.

It is now dealing with queue changes over time with the newly introduced TicketChange for field _('Queue').

A unit test was made to validate creation of the TicketChange together with the modification of the ticket queue.

Another one to test multiple queue exclusions over time with correct time_spent calculation.

uhurusurfa (approved):
Looks good - thanks for adding the unit tests

uhurusurfa:
@Benbb96 - if you do not see any issues here then please merge the PR

helpdesk/update_ticket.py
```diff
@@ -302,6 +305,16 @@ def update_ticket(
...
         ticket.priority = priority
 
+    if queue != ticket.queue.id:
+        c = TicketChange(
+            followup=f,
+            field=_('Queue'),
+            old_value=ticket.queue.id,
+            new_value=queue,
+        )
+        c.save()
```
Benbb96, line 315: Do you think you could take the opportunity here to refactor all the creation of TicketChange like this :

```suggestion
        c = f.ticketchange_set.create(
            field=_('Queue'),
            old_value=ticket.queue.id,
            new_value=queue,
        )
```

samsplunks:
Thank you @Benbb96, suggestions committed!

Benbb96 (approved):
Alright, no problem!

PR #1169, opened 2024-04-16
Emit a signal when a ticket is updated

samsplunks:
I have a use case where I need to follow the modifications of the tickets.

I unfortunately cannot hook to the post_save signal of the FollowUp model because at this time, the TicketChange and Ticket models are not updated yet with new data.

Someone added the `notify_followup_webhooks` function to `update_ticket` for a similar purpose.

In this PR, I propose to have the update_ticket function emit a signal when its process is done.
This way, anyone can hook post processing workflows outside of helpdesk.

Not to have duplicate code logic, I move the notify_followup_webhooks outside of update_ticket and have it listen to the ticket_update_done signal the same way someone would do it outside of helpdesk.

### DIFF
--- a/docs/webhooks.rst
+++ b/docs/webhooks.rst
@@ -17,13 +17,14 @@ Signals
 
 Webhooks are triggered through django signals.
 
-The two existing signals are:
+The two available signals are:
   - new_ticket_done
   - update_ticket_done
 
 You have the opportunity to listen to those in your project if you have post processing workflows outside of webhooks::
 
-
+  
+  from django.dispatch import receiver
   from helpdesk.signals import new_ticket_done, update_ticket_done
   
   @receiver(new_ticket_done)
--- a/docs/webhooks.rst
+++ b/docs/webhooks.rst
@@ -10,3 +10,29 @@ You can register webhooks to allow third party apps to be notified of helpdesk e
 3. You can optionally set ``HELPDESK_WEBHOOK_TIMEOUT`` which defaults to 3 seconds. Warning, however, webhook requests are sent out sychronously on ticket update. If your webhook handling server is too slow, you should fix this rather than causing helpdesk freezes by messing with this variable.
 
 Once these URLs are configured, a serialized copy of the ticket object will be posted to each of these URLs each time a ticket is created or followed up on respectively.
+
+
+Django signals
+--------------
+
+Webhooks are triggered through signals emitted for new tickets creation and ticket follow-up updates.
+
+The two existing signals are:
+  - new_ticket_done
+  - update_ticket_done
+
+You have the opportunity to listen to those in your project if you have post processing workflows outside of webhooks.
+
+```
+from helpdesk.signals import new_ticket_done, update_ticket_done
+
+@receiver(new_ticket_done)
+def process_new_ticket(sender, ticket, **kwargs):
+    "Triggers this code when a ticket is created."
+    pass
+    
+@receiver(update_ticket_done)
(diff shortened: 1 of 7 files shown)


samsplunks:
Also added the signal to email.py, this way `notify_followup_webhooks` is only triggered through signal.

uhurusurfa (commented):
Looks good - can you add  a basic unit test to f=verify the webhook feature?

timthelion:
I already included tests for webhooks 🙂https://github.com/django-helpdesk/django-helpdesk/blob/main/helpdesk/tests/test_webhooks.py

There is a second webhook that is triggered on ticket creation. It would be nice to go all the way and wire that through the signals as well.

samsplunks:
Yes, I used the test_webhooks to verify compliance, thanks Timothy!

I knew you were going to ask. ;)

samsplunks:
The new ticket webhook is activated with a specific role, where a signal should be emitted no matter what.
And I don't really understand the relation between new_ticket_cc and the webhook activation, could you explain it?

```
        if "new_ticket_cc" in roles:
            send_new_ticket_webhook(self)
```

Anyway, what I would do is place the signal close to the new ticket creation, which I think is the `forms.TicketForm.save()` method before self._send_messages:
https://github.com/django-helpdesk/django-helpdesk/blob/main/helpdesk/forms.py#L421

And use `ticket._state.adding` to check whether the ticket was created or modified.
The signal would be emitted at every ticket modification but the webhook would be activated at ticket creation.

Does that sound OK to you?

samsplunks:
Looks good to me.

Note that the send_new_ticket_webhook is now disconnected from the new_ticket_cc role.

timthelion:
I put it in the "new_ticket_cc" location because the tickets are not created with the form when they are processed in email.py if I recall correctly.

samsplunks:
Thank you.
I had the same problem with email.py and fixed it by adding the signal to it.

helpdesk/signals.py
```diff
@@ -0,0 +1,7 @@
+import django.dispatch
+
+# create a signal for *TicketForm
```
Benbb96, line 3: small typo here I think

docs/webhooks.rst
```diff
@@ -10,3 +10,28 @@ You can register webhooks to allow third party apps to be notified of helpdesk e
 3. You can optionally set ``HELPDESK_WEBHOOK_TIMEOUT`` which defaults to 3 seconds. Warning, however, webhook requests are sent out sychronously on ticket update. If your webhook handling server is too slow, you should fix this rather than causing helpdesk freezes by messing with this variable.
 
 Once these URLs are configured, a serialized copy of the ticket object will be posted to each of these URLs each time a ticket is created or followed up on respectively.
+
+
+Signals
+--------------
+
+Webhooks are triggered through django signals.
```
Benbb96, line 18: A link to the django documentation could be a nice thing:

```suggestion
Webhooks are triggered through `Django Signals <https://docs.djangoproject.com/en/stable/topics/signals/>_`.
```

helpdesk/signals.py
samsplunks: It was intended as TicketForm and PublicTicketForm are emitting the signal

Benbb96: Oh ok, I understand now. Yeah it was strange for a typo

Benbb96 (approved):

timthelion:
@uhurusurfa I think this can be merged now.

PR #1173, opened 2024-04-24
Update release number for package creation

uhurusurfa:
Release number updated so that we can make a new release.


Benbb96 (approved):

Benbb96:
Should we wait for #1161 to be merged before releasing?

PR #1176, opened 2024-05-02
Sidebar bookmarks, closes  #1175

### LINKED ISSUE #1175: Configurable list of bookmarks in sidebar menu

- When integrating django-helpdesk into an existing Django project one can 'enter' the helpdesk, but not 'leave' -
  navigate back to other apps in the project.
- In a helpdesk environment it is helpful provide the personnel with links to related applications or sites.

A configurable list of "bookmarks" in the sidebar menu, which is clearly separated from the django-helpdesk menu entries.
This bookmarks section must not show, when no bookmarks are configured.

None

**Additional context**

PR is following

jorge-leon:
This PR adds a configuration setting:

     HELPDESK_BOOKMARKS

When set to a list of 'bookmarks', an additional section is shown on the sidebar menu, rendering the bookmarks.

There are two types of bookmarks:

- internal: given as names to resolve by the `reverse()` method of the URL dispatcher
- external: given as URLs.

They are rendered differently so the user can distinguish if she will stay in the Django project or leave to another site.

A bookmark is specified in the following way:
- internal: `{ 'name': `_internal_name_`, 'label': `_label_ }`
- external: `{ 'URL': `_external_url_`, 'label': `_label_ }`

The `label` is the link text. It is optional, if not given the `name` or `URL` is used instead.

### DIFF
--- a/helpdesk/settings.py
+++ b/helpdesk/settings.py
@@ -71,6 +71,19 @@ HELPDESK_TICKETS_TIMELINE_ENABLED = getattr(
 HELPDESK_NAVIGATION_ENABLED = getattr(
     settings, 'HELPDESK_NAVIGATION_ENABLED', False)
 
+# Configure a list of Bookmarks/links shown on the bottom of the sidebar
+# menu. Disabled if no bookmark is configured.
+#
+# Set to a list of dictionary in one of the following forms:
+# - { 'URL': external_url, 'label': label } .. external link
+# - { 'name': internal_name, 'label': label } .. internal name is resolved with reverse()
+#
+# If the label is omitted, the URL or name respectively is used instead.
+
+HELPDESK_BOOKMARKS = getattr(
+    settings, 'HELPDESK_BOOKMARKS', [])
+
+
 # use public CDNs to serve jquery and other javascript by default?
 # otherwise, use built-in static copy
 HELPDESK_USE_CDN = getattr(settings, 'HELPDESK_USE_CDN', False)
--- a/helpdesk/templates/helpdesk/navigation-sidebar.html
+++ b/helpdesk/templates/helpdesk/navigation-sidebar.html
@@ -91,5 +91,23 @@
           </a>
         </li>
         {% endif %}
+      {% endif %}
+      {% if helpdesk_settings.HELPDESK_BOOKMARKS and user.is_authenticated %}
+      <li><hr class="dropdown-divider"></li>
+      {% for bookmark in helpdesk_settings.HELPDESK_BOOKMARKS %}
+      <li class="nav-item">
+        {% if bookmark.URL %}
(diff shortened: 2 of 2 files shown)


timthelion:
I personally would prefer to add sections to the django template if needed so that you can simply over-ride and reimport the the template than to have a new setting for this.

jorge-leon:
> I personally would prefer to add sections to the django template if needed so that you can simply over-ride and reimport the the template than to have a new setting for this.

Can you elaborate on this. I'll consider doing the respective changes when I know how to do it. Maybe there is an example you can point me to.

Benbb96:
Yes I think documentation could be better explained on [this page](https://django-helpdesk.readthedocs.io/en/v1.2.0/custom_templates.html), maybe with an example like overriding the sidebar.

Following [Django documentation](https://docs.djangoproject.com/en/stable/howto/overriding-templates/#extending-an-overridden-template), we could add custom blocks in _django-helpdesk_ templates in order to avoid too much copy-pasting when overriding a file.

timthelion:
@jorge-leon 

To do this you would rename https://github.com/django-helpdesk/django-helpdesk/blob/main/helpdesk/templates/helpdesk/navigation-sidebar.html to navigation-sidebar-base.html you can then add blocks like so https://github.com/django-helpdesk/django-helpdesk/blob/dce3c0c73f3a304118d62c9affdee1c73ba608e7/helpdesk/templates/helpdesk/base.html#L48 where you want to insert things into the sidebar.

Then you would add a new `navigation-sidebar.html` file which uses an extends line like this https://github.com/django-helpdesk/django-helpdesk/tree/main/helpdesk/templates/helpdesk

Then you would create a PR to this repo.

After that, you can add a template file to the appropriate directory to override `navigation-sidebar.html`. Here is an example path for where you would place a file to over-ride a template:

https://github.com/django-helpdesk/django-helpdesk/blob/dce3c0c73f3a304118d62c9affdee1c73ba608e7/standalone/docker-compose.yml#L21

uhurusurfa:
@jorge-leon Is this still a work in progress or ready for review?

jorge-leon:
> @jorge-leon Is this still a work in progress or ready for review?

Now that the other PR's are done, I will work on this one.
However it will be a completely different implementation.  So it would also be ok to just close this PR without merging until I work out the other approach.

uhurusurfa:
> > @jorge-leon Is this still a work in progress or ready for review?
> 
> Now that the other PR's are done, I will work on this one. However it will be a completely different implementation. So it would also be ok to just close this PR without merging until I work out the other approach.

Yes go ahead.
Note that now thatr you are a team collaborator you can raise PR's within this project instead of a cloned version.

uhurusurfa:
@jorge-leon Just to confirm that we should close this PR as you will raise a different one to accomplish a similar bit of functionality?

PR #1178, opened 2024-05-03
repair pre-set replies broken due to raw url parameter name mismatch

ilikerobots:
In https://github.com/django-helpdesk/django-helpdesk/commit/574395ee2814486e6bdca54dad83081279cb20b7#diff-206db4cdb6cf21b5ed528d9c5ebf6be95c221a1da93bbf1d2470521af208774eL1316, the raw_details parameter was changed from `type` to `type_` but the corresponding path regex parameter was not also updated, rendering pre-set replies unloadable.

### DIFF
--- a/helpdesk/urls.py
+++ b/helpdesk/urls.py
@@ -103,7 +103,7 @@ urlpatterns = [
         staff.delete_ticket_checklist,
         name="delete_ticket_checklist"
     ),
-    re_path(r"^raw/(?P<type>\w+)/$", staff.raw_details, name="raw"),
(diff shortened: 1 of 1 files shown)


Benbb96 (approved):
Good catch!

It shows preset replies are not used much. Maybe there are other bugs around this feature...

ilikerobots:
We're using it in production, all seems ok except for the issue above, but I've only upgraded to 1.0 so far.

uhurusurfa (approved):

PR #1181, opened 2024-05-07
Fix logout in Django 5 with POST action instead of GET

### LINKED ISSUE #1174: logout stops working with Django >4.0
**Describe the bug**

Since Django 4.1 the contributed `LogoutView` requires the POST method, see:

https://docs.djangoproject.com/en/4.1/releases/4.1/#log-out-via-get

Find attached a patch which replaces the logout link (`<a href=`...) with a form. In order to maintain formatting
a label with a hidden submit button is used. I do not know if this works for all browsers.
 
[django_logout_post.patch.txt](https://github.com/django-helpdesk/django-helpdesk/files/15129648/django_logout_post.patch.txt)

Note: in `urls.py`  the `next_page` parameter uses a plain `../` which might not do what is meant in all situations. It could be
replaced by `reverse_lazy("helpdesk:home")`, however I plan to integrate a more general solution in a future PR for improving
integration of django-helpdesk into a django-project with other apps.

**To Reproduce**
1. Install django-helpdesk with a Django version greater or equal to 4.1
2. Set up a user and login
3. Click on the "Logout" link in the user dropdown
4. Click on the back button of the browser and optionally reload
5. The user is still logged in.

**Expected behavior**

The user is logged out

Benbb96:
Fixes #1174

Uses technique from https://learndjango.com/tutorials/django-login-and-logout-tutorial#log-out-button

### DIFF
--- a/helpdesk/urls.py
+++ b/helpdesk/urls.py
@@ -9,7 +9,7 @@ urls.py - Mapping of URL's to our various views. Note we always used NAMED
 
 from django.contrib.auth import views as auth_views
 from django.contrib.auth.decorators import login_required
-from django.urls import include, path, re_path, reverse_lazy
+from django.urls import include, path, re_path
 from django.views.generic import TemplateView
 from helpdesk import settings as helpdesk_settings
 from helpdesk.decorators import helpdesk_staff_member_required, protect_view
--- a/helpdesk/urls.py
+++ b/helpdesk/urls.py
@@ -9,7 +9,7 @@ urls.py - Mapping of URL's to our various views. Note we always used NAMED
 
 from django.contrib.auth import views as auth_views
 from django.contrib.auth.decorators import login_required
-from django.urls import include, path, re_path
+from django.urls import include, path, re_path, reverse_lazy
 from django.views.generic import TemplateView
 from helpdesk import settings as helpdesk_settings
 from helpdesk.decorators import helpdesk_staff_member_required, protect_view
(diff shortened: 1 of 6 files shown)


helpdesk/urls.py
```diff
@@ -215,7 +215,7 @@ def get_context_data(self, **kwargs):
     path(
         "logout/",
         auth_views.LogoutView.as_view(
-            template_name="helpdesk/registration/login.html", next_page="../"
+            template_name="helpdesk/registration/login.html", next_page=reverse_lazy("helpdesk:home")
```
Benbb96, line 218: Note that the `template_name` does not seem to be useful when we provide `next_page`.  
Moreover, there is a template for the [confirmation of logging out](https://github.com/django-helpdesk/django-helpdesk/blob/main/helpdesk/templates/helpdesk/registration/logged_out.html), but it's not called anywhere. I believe it would make more sense to point to it instead of login page.

uhurusurfa: Agreed - go ahead and change the logout next page to the confirmation of logout page.

uhurusurfa: Ideally the logged out page needs a login button otherwise the user who is security concious and logs out regularly has to change the URL or use a bookmark.

Benbb96: Alright, here's what I have now for the logged_out page :

image

What do you think?

A login button is available in the header so it might not be needed to show it again.

uhurusurfa (approved):
🚀

uhurusurfa:
Awesome - much better flow

Issue #1184, opened 2024-05-17
Make parent tickets visible in ticket view, improve handling of dependencies


jorge-leon:
**Is your feature request related to a problem? Please describe.**

I want to use  "milestone tickets", where individual tickets to be resolved are added.

While I can see the dependent tickets on the milestone, there is no way to see the milestone(s) a ticket belongs to.

**Describe the solution you'd like**

The ticket description should have an additional field which contains a list of 'parent' tickets, i.e. tickets which depend
on the current one.

**Describe alternatives you've considered**

Filtering, but this is easily ruled out as very inefficient.

**Additional context**

A PR will be supplied.

lonewolve24:
Hello @uhurusurfa I will like to work on this bug if this issue can be assigned to me thanks

jorge-leon:
> Hello @uhurusurfa I will like to work on this bug if this issue can be assigned to me thanks

Sorry, that one is already resolved. will close this issue now.

PR #1185, opened 2024-05-17
Add list of 'parent' tickets to ticket description table, solves #1184

jorge-leon:
'parent' tickets are all tickets on which the current ticket depends.

The parent tickets can only be resolved if the current tickets is closed.

This patch splits the current |Dependencies|Total time spent| table row in two rows.  The table layout is now

    |Status |Total time spent|
    |Depends|Resolves        |

Status and Total time only have one line.  Status was chosen as filler field, because I believe it should be moved out of the title to reduce visual distraction.

Depends and Resolve will have a variable number of lines and are closely related.

Depends was chosen instead of "Dependencies", to indicate directly what the listed tickets do and because it is shorter.

The commit also simplifies the wording of the tooltips and harmonizes with the other tooltips in the same table: It indicates directly what happens when the button is clicked (or touched!).

Notes:

- The complete row could be omitted if now Depends or Resolves exists for a ticket.  This would reduce visual distraction but was not implemented because it could lead to misled expectations.
- No button for removing/adding parent tickets was added. If considered important this can be added any time.  Currently a user has to go to the parent ticket and delete the dependency (easy) or remember the ticket id/title, search for the parent and add the ticket as dependency (difficult)

### DIFF
--- a/helpdesk/forms.py
+++ b/helpdesk/forms.py
@@ -591,6 +591,14 @@ class TicketDependencyForm(forms.ModelForm):
         model = TicketDependency
         exclude = ('ticket',)
 
+class TicketResolvesForm(forms.ModelForm):
+    ''' Adds this ticket as a dependency for a different ticket '''
+
+    class Meta:
+        model = TicketDependency
+        #exclude = ('depends_on',)
+        fields = ('ticket',)
+
 
 class MultipleTicketSelectForm(forms.Form):
     tickets = forms.ModelMultipleChoiceField(
--- a/helpdesk/forms.py
+++ b/helpdesk/forms.py
@@ -591,6 +591,12 @@ class TicketDependencyForm(forms.ModelForm):
         model = TicketDependency
         exclude = ('ticket',)
 
+    def __init__(self, ticket, *args, **kwargs):
+        super(TicketDependencyForm,self).__init__(*args, **kwargs)
+
+        # Exclude duplicate tickets, myself, existing dependencies and parents
+        self.fields['depends_on'].queryset = Ticket.objects.exclude(status=Ticket.DUPLICATE_STATUS).exclude(id=ticket.id).exclude(depends_on__ticket=ticket).exclude(ticketdependency__depends_on=ticket)
+
 class TicketResolvesForm(forms.ModelForm):
     ''' Adds this ticket as a dependency for a different ticket '''
 
@@ -598,6 +604,12 @@ class TicketResolvesForm(forms.ModelForm):
         model = TicketDependency
         #exclude = ('depends_on',)
         fields = ('ticket',)
+
+    def __init__(self, ticket, *args, **kwargs):
+        super(TicketResolvesForm,self).__init__(*args, **kwargs)
+
+        # Exclude duplicate tickets, myself, existing dependencies and parents
+        self.fields['ticket'].queryset = Ticket.objects.exclude(status=Ticket.DUPLICATE_STATUS).exclude(id=ticket.id).exclude(depends_on__ticket=ticket).exclude(ticketdependency__depends_on=ticket)
 
 
 class MultipleTicketSelectForm(forms.Form):
--- a/helpdesk/forms.py
+++ b/helpdesk/forms.py
@@ -589,7 +589,7 @@ class TicketDependencyForm(forms.ModelForm):
 
     class Meta:
         model = TicketDependency
-        exclude = ('ticket',)
+        fields = ('depends_on',)
 
     def __init__(self, ticket, *args, **kwargs):
         super(TicketDependencyForm,self).__init__(*args, **kwargs)
@@ -602,7 +602,6 @@ class TicketResolvesForm(forms.ModelForm):
 
     class Meta:
         model = TicketDependency
-        #exclude = ('depends_on',)
         fields = ('ticket',)
 
     def __init__(self, ticket, *args, **kwargs):
--- a/helpdesk/forms.py
+++ b/helpdesk/forms.py
@@ -594,8 +594,8 @@ class TicketDependencyForm(forms.ModelForm):
     def __init__(self, ticket, *args, **kwargs):
         super(TicketDependencyForm,self).__init__(*args, **kwargs)
 
-        # Exclude duplicate tickets, myself, existing dependencies and parents
-        self.fields['depends_on'].queryset = Ticket.objects.exclude(status=Ticket.DUPLICATE_STATUS).exclude(id=ticket.id).exclude(depends_on__ticket=ticket).exclude(ticketdependency__depends_on=ticket)
+        # Only open tickets except myself, existing dependencies and parents
+        self.fields['depends_on'].queryset = Ticket.objects.filter(status__in=Ticket.OPEN_STATUSES).exclude(id=ticket.id).exclude(depends_on__ticket=ticket).exclude(ticketdependency__depends_on=ticket)
 
 class TicketResolvesForm(forms.ModelForm):
     ''' Adds this ticket as a dependency for a different ticket '''
@@ -608,8 +608,8 @@ class TicketResolvesForm(forms.ModelForm):
     def __init__(self, ticket, *args, **kwargs):
         super(TicketResolvesForm,self).__init__(*args, **kwargs)
 
-        # Exclude duplicate tickets, myself, existing dependencies and parents
-        self.fields['ticket'].queryset = Ticket.objects.exclude(status=Ticket.DUPLICATE_STATUS).exclude(id=ticket.id).exclude(depends_on__ticket=ticket).exclude(ticketdependency__depends_on=ticket)
+        # Only open tickets except myself, existing dependencies and parents
+        self.fields['ticket'].queryset = Ticket.objects.exclude(status__in=Ticket.OPEN_STATUSES).exclude(id=ticket.id).exclude(depends_on__ticket=ticket).exclude(ticketdependency__depends_on=ticket)
 
 
 class MultipleTicketSelectForm(forms.Form):
--- a/helpdesk/templates/helpdesk/ticket_resolves_add.html
+++ b/helpdesk/templates/helpdesk/ticket_resolves_add.html
@@ -1,0 +1,37 @@
+{% extends "helpdesk/base.html" %}{% load i18n %}
+
+{% block helpdesk_title %}{% trans "Add Ticket as Dependency" %}{% endblock %}
+
+{% block helpdesk_breadcrumb %}
+<li class="breadcrumb-item">
+    <a href="{% url 'helpdesk:list' %}">{% trans "Tickets" %}</a>
+</li>
+<li class="breadcrumb-item">
+    <a href="{% url 'helpdesk:list' %}{{ depends_on.id }}/">{{ depends_on.queue.slug }}-{{ depends_on.id }}</a>
+</li>
+<li class="breadcrumb-item active">{% trans "Add Ticket as Dependency" %}</li>
+{% endblock %}
+
+{% block helpdesk_body %}{% blocktrans %}
+<h2>Add Ticket as Dependency</h2>
+
+<p>Ticket: {{ depends_on }} </p>
+
+<p>Adding this ticket as dependency will stop you resolving the selected ticket until this ticket has been resolved or closed.</p>{% endblocktrans %}
+
+<form method='post' action='./'>
+
+<fieldset>
+    <dl>{% for field in form %}
+        <dt><label for='id_{{ field.name }}'>{{ field.label }}</label></dt>
+        <dd>{{ field }}</dd>
+        {% if field.errors %}<dd class='error'>{{ field.errors }}</dd>{% endif %}
+        {% if field.help_text %}<dd class='form_help_text'>{{ field.help_text }}</dd>{% endif %}
+    {% endfor %}</dl>
+</fieldset>
+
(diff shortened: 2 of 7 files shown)


Benbb96:
Hi @jorge-leon, thanks for your contribution!

Do you think you can provide screenshots to better see the rendering?

Buttons to add/remove dependencies could be a nice feature but it's up to you. And it could be done later in another PR.

Just a side note: when changing the wording in `trans` block, I believe it could be a good thing to recompile locale messages in order to easily translate them in`.po` files.

jorge-leon:
> Hi @jorge-leon, thanks for your contribution!
> 
> Do you think you can provide screenshots to better see the rendering?
> 
Sure!

With parent links:

image

With dependency link:

image

Benbb96 (approved):
Thanks for the screenshots. It looks good to me. There's only the status that get duplicated. Maybe we should use `ticket.get_status_display` (at line 15) instead of `get_status` because it adds `Open dependencies` which is a bit verbose in the title

jorge-leon:
> Hi @jorge-leon, thanks for your contribution!
...
> 
> Just a side note: when changing the wording in `trans` block, I believe it could be a good thing to recompile locale messages in order to easily translate them in`.po` files.

When I did the following, starting from the git checkout root directory:

1. activate the .venv
2. make the demo project
3. go to the `helpdesk` subdirectory
4. run `../demo/manage.py makemessages -a`

I got a `git diff` of over 4000 lines for `locale/en/LC_MESSAGES/django.po`. Almost every diff is only a changed line number of a string to translate.

Don't know if a) I did it the right way and b) if I should commit such big changes.

I prefer somebody more knowledgeable to create and commit the .po's. I'll do the English, German and Spanish translations then - as soon as I get access on Transifex.

jorge-leon:
> Thanks for the screenshots. It looks good to me. There's only the status that get duplicated. Maybe we should use `ticket.get_status_display` (at line 15) instead of `get_status` because it adds `Open dependencies` which is a bit verbose in the title

Will do that.

I even propose to remove the status completely from the title: a user clicking on a ticket link does this normally based on a decision which includes evaluating the status shown in the ticket list where she/he found the ticket in the first place.  The status is not an information that is primarily looked up by opening a ticket and looking at the title.  Instead, the ticket title is the main information. As you point out, it's already quite verbose, IMHO it need not be repeated at all in this place.

jorge-leon:
### Next iteration:

- Removed status from title.
- Added functionality to add/remove a "parent" ticket.
- Reformatted from two columns into two rows with full width
- Reformatted lists into (bootstrap) tables
- Ordered dependency list: first all open tickets then all others

### New screenshots:

#### Without dependencies:

image

#### With dependencies:

Note: it is possible to make circular dependencies. See the DH-4 ticket, which is
both a parent and a dependency. This might be addressed by another PR.

image

helpdesk/views/staff.py
```diff
@@ -422,8 +424,16 @@ def view_ticket(request, ticket_id):
 
         return redirect('helpdesk:edit_ticket_checklist', ticket.id, checklist.id)
 
+    open_dependencies = ticket.ticketdependency.filter(depends_on__status__in=Ticket.OPEN_STATUSES)
+    dependencies = [d for d in itertools.chain(
+        open_dependencies,
+        ticket.ticketdependency.all().difference(open_dependencies)
+    )]
```
Benbb96, line 431: I don't understand why you need to do this. Isn't it possible to simply do :

```suggestion
    dependencies = ticket.ticketdependency.all()
```

Maybe you wish to keep open statuses at the beginning?

helpdesk/templates/helpdesk/ticket_resolves_add.html
```diff
@@ -0,0 +1,37 @@
...
+        {% if field.errors %}<dd class='error'>{{ field.errors }}</dd>{% endif %}
+        {% if field.help_text %}<dd class='form_help_text'>{{ field.help_text }}</dd>{% endif %}
+    {% endfor %}</dl>
+</fieldset>
+
+<input class="btn btn-primary" type='submit' value='{% trans "Save Ticket Dependency" %}' />
+
+{% csrf_token %}</form>
+
+{% endblock %}
```
Benbb96, line 37: Can you show also show a screenshot of this new page, please?

helpdesk/views/staff.py
jorge-leon: Yes: It's about ordering. I use helpdesk (also) for tracking software development/maintenance. With the ticket dependencies I can model milestones, but I want to trickle resolved/closed tickets down the list, so I can pick always from top.

image

I thought about ordering by status first, just like you point it out. But  since ticket status can be configured / named freely ordering might not
in all cases put open tickets on top. In contrast, using the TICKET_OPEN_STATUSES settings is a stable approach.

The itertools trick allowed me to keep the template code tidy, since there is only one loop to go through - or an empty list.  The implementation is based on the assumption, that the number of dependent ticktes is not huge.  I would have preferred a QueryList, but could not came up with an order preserving expression.

Another side topic would be to sort the two lists (open/closed tickets) individually before creating the list, but then I did not have a specific preference and left it out.

helpdesk/templates/helpdesk/ticket_resolves_add.html
jorge-leon: Shure! It's almost a hardcopy of `ticket_depends_add.html`, For deleting a "parent" dependencies I could recicly `ticket_depends_del.html`, but needed a new view.

image

jorge-leon:
> ... I'll do the English, German and Spanish translations then - as soon as I get access on Transifex.

Any update on .po's and translations?

helpdesk/views/staff.py
uhurusurfa: Given that status values can be overridden and there is no directive to specify status values to specifically accommodate ordering so that the open statuses appear first, it may be better to enhance the operators available in the ORM to include "NOT IN" to retrieve the unopened dependencies like this:
`ticket.ticketdependency.filter(depends_on__status__not_in=Ticket.OPEN_STATUSES)`
Then you simplify the code to this:
```
    open_dependencies = ticket.ticketdependency.filter(depends_on__status__in=Ticket.OPEN_STATUSES)
    dependencies = open_dependencies + ticket.ticketdependency.filter(depends_on__status__not_in=Ticket.OPEN_STATUSES)
```

To do this we add this class to a module helpdesk/orm_operators.py:
```
from django.db.models.lookups import In

class NotIn(In):
    lookup_name = "not_in"

    def get_rhs_op(self, connection, rhs):
        return "NOT IN %s" % rhs
```

Then make the new operator available in the ORM by adding this to the ready method in helpdesk/apps.py:
```
        from django.db.models.fields import Field
        from helpdesk.orm_operators import NotIn
        Field.register_lookup(NotEqual)
```

uhurusurfa:
> > ... I'll do the English, German and Spanish translations then - as soon as I get access on Transifex.
> 
> Any update on .po's and Transifex access?

@jorge-leon  I am unaware of anyone with access to Trnsifex. I suggest you do the translations you can and we will see if we can get other users of the project who need the other supported languages to update the relevant language files via a PR.

jorge-leon:
> > > ... I'll do the English, German and Spanish translations then - as soon as I get access on Transifex.
> > 
> > 
> > Any update on .po's and Transifex access?
> 
> @jorge-leon I am unaware of anyone with access to Trnsifex. I suggest you do the translations you can and we will see if we can get other users of the project who need the other supported languages to update the relevant language files via a PR.

Oh! the `CONTRIBUTING.rst` file sent me to Transifex ...

helpdesk/views/staff.py
jorge-leon: Excelent, will do that.

uhurusurfa:
> Oh! the `CONTRIBUTING.rst` file sent me to Transifex ...

I was unaware of that... I have requested access to the Helpdesk project in Transifex and will provide you with access as soon as I get access (I ssume you have requested access after creating an open source user ID in Transifex)

helpdesk/views/staff.py
jorge-leon: The `+` does not work on querysets, but it seems to work with `|`

jorge-leon: `|` is not order preserving. The current implementation still uses NotIn, but concatenates via list()+list()

helpdesk/views/staff.py
```diff
@@ -1695,6 +1701,42 @@ def ticket_dependency_del(request, ticket_id, dependency_id):
...
+    depends_on = get_object_or_404(Ticket, id=ticket_id)
+    ticket_perm_check(request, depends_on)
+    if request.method == 'POST':
+        form = TicketResolvesForm(depends_on, request.POST)
+        if form.is_valid():
+            ticketdependency = form.save(commit=False)
+            ticketdependency.depends_on = depends_on
+            if ticketdependency.ticket != ticketdependency.depends_on:
+                ticketdependency.save()
+            return HttpResponseRedirect(reverse('helpdesk:view', args=[depends_on.id]))
```
Benbb96, line 1715: You use shortcut like this I believe:

```suggestion
            return redirect(depends_on.get_absolute_url())
```

helpdesk/views/staff.py
```diff
@@ -422,8 +423,13 @@ def view_ticket(request, ticket_id):
 
         return redirect('helpdesk:edit_ticket_checklist', ticket.id, checklist.id)
 
+    # Open tickets on top
+    dependencies = list(ticket.ticketdependency.filter(depends_on__status__in=Ticket.OPEN_STATUSES)) \
+        + list(ticket.ticketdependency.filter(depends_on__status__not_in=Ticket.OPEN_STATUSES))
```
Benbb96, line 428: `not_in` lookup is more readable so no need to change it, but I think `ticket.ticketdependency.exclude(depends_on__status__in=Ticket.OPEN_STATUSES)` would do the same

helpdesk/views/staff.py
jorge-leon: Not my code. Shall I change it anyway?

Benbb96: As you want, maybe this could be done for each occurrence in another MR. It's been a while since I'm thinking about refactoring/modernizing the code base (especially views and better formatting templates).

jorge-leon: Changed in two places, checked in, test are running, the first test already passed.

helpdesk/views/staff.py
jorge-leon: You are right. If you followed the related discussion you might see how we got here. The primary goal - a 'fast' way to concatentate two QuerySets in order - cannot be achieved anyway.  So I might as well remove the NotIn ORM operator, since this is the only place where it is used to remove code complexity.

There might be a way to do the chaining via annotations, as show [here](https://stackoverflow.com/questions/17216036/how-to-combine-two-select-queries-with-their-own-order) for SQL.

Benbb96: Yep, this could be a readable solution! Too bad there isn't an easy one for that king of thing

uhurusurfa: Please do not use EXCLUDE for this case.
It produces a very different set of underlying SQL that will be come a performance issue for deployments with large numbers of tickets.

jorge-leon:
```
from django.db.models import Value
open_deps = ticket.ticketdependency.filter(depends_on__status__in=Ticket.OPEN_STATUSES)).annotate(rank=Value(1))
not_open_deps = ticket.ticketdependency.filter(depends_on__status__not_in=Ticket.OPEN_STATUSES)).annotate(rank=Value(2))
dependencies=open_deps.union(not_open_deps).order_by('rank')
```

Better?

@uhurusurfa : Please elaborate or point to some information regarding `.exclude` performance penalty.
A quick lookup in the web did not bring up substantial related information for me.

jorge-leon: I think I have found now an optimal solution using conditional annotation, please review the last commit.

Benbb96: Yes this looks better!

helpdesk/forms.py
```diff
@@ -591,6 +591,26 @@ class Meta:
...
+
+        # Only open tickets except myself, existing dependencies and parents
+        self.fields['depends_on'].queryset = Ticket.objects.filter(status__in=Ticket.OPEN_STATUSES).exclude(id=ticket.id).exclude(depends_on__ticket=ticket).exclude(ticketdependency__depends_on=ticket)
+
+class TicketResolvesForm(forms.ModelForm):
+    ''' Adds this ticket as a dependency for a different ticket '''
+
+    class Meta:
+        model = TicketDependency
+        #exclude = ('depends_on',)
```
Benbb96, line 605: Maybe this can be removed

helpdesk/views/staff.py
```diff
@@ -422,8 +423,16 @@ def view_ticket(request, ticket_id):
 
         return redirect('helpdesk:edit_ticket_checklist', ticket.id, checklist.id)
 
+    # List open tickets on top
+    dependencies = ticket.ticketdependency.annotate(
+        rank=Case(
+                When(depends_on__status__in=Ticket.OPEN_STATUSES, then=Value('1')),
+                default=Value('2')
```
Benbb96, line 430: I believe you can use integer for the rank values

```suggestion
                When(depends_on__status__in=Ticket.OPEN_STATUSES, then=Value(1)),
                default=Value(2)
```

Benbb96 (approved):

PR #1189, opened 2024-06-06
Attachment enable setting

### LINKED ISSUE #848: Configuration Option Allow/Disallow Attachments
Attachments on tickets are a wonderful addition to the ticketing system, however, they also have maintenance overhead and security risks. Some companies do not allow attachments on emails or ticket systems. In a simpler installation of the ticketing system, having the option to disable attachments can be beneficial from both maintenance and security standpoints. An alternative to save a file in the database or in the ticket systems is to allow the capability to include a link to a file location on a network.

Create a configuration option HELPDESK_ALLOW_ATTACHMENTS that is set in the Django Project settings.py file.
If set to false it would remove the option from the menus and not allow the actions to occur. Cases where the option was allowed and files exist would need to be taken into account.

1) Monkey patch the Attachments class save method in the helpdesk models.py to just return. This is effective at disallowing attachments, but provides no lasting solution and can perplex users trying to attach files.
2) Have this option and/or alternative to use a link to a file location provided on a network.

jorge-leon:
Fixes #848. See also discussion  #1188

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -39,8 +39,8 @@ If you want to override the default settings for your users, create ``HELPDESK_D
     }
 
 
-Access controll & Security
----------------
+Access control & Security
+-------------------------
 These settings can be used to change who can access the helpdesk.
 
 - **HELPDESK_PUBLIC_VIEW_PROTECTOR** This is a function that takes a request and can either return `None` granting access to to a public view or a redirect denying access.
@@ -51,16 +51,24 @@ These settings can be used to change who can access the helpdesk.
 
   **Default:** ``HELPDESK_REDIRECT_TO_LOGIN_BY_DEFAULT = False``
 
-- **HELPDESK_VALID_EXTENSIONS** Valid extensions for file types that can be attached to tickets. Note: This used to be calle **VALID_EXTENSIONS** which is now deprecated.
-
-  **Default:** ``HELPDESK_VALID_EXTENSIONS = ['.txt', '.asc', '.htm', '.html', '.pdf', '.doc', '.docx', '.odt', '.jpg', '.png', '.eml']
-
-- **HELPDESK_VALIDATE_ATTACHMENT_TYPES** If you'd like to turn of filtering of helpdesk extension types you can set this to False.
-
 - **HELPDESK_ANON_ACCESS_RAISES_404** If True, redirects user to a 404 page when attempting to reach ticket pages while not logged in, rather than redirecting to a login screen.
 
   **Default:** ``HELPDESK_ANON_ACCESS_RAISES_404 = False``
 
+Settings related to attachments:
+
+- **HELPDESK_ENABLE_ATTACHMENTS** If set to ``True``, files can be
+  attached to tickets and followups, and emails are searched for
+  attachments which are then attached to the ticket.  Also enables the
+  ``HELPDESK_ALWAYS_SAVE_INCOMING_EMAIL_MESSAGE`` setting.
+  
+- **HELPDESK_VALID_EXTENSIONS** Valid extensions for file types that can be attached to tickets. Note: This used to be called **VALID_EXTENSIONS** which is now deprecated.
+
+  **Default:** ``HELPDESK_VALID_EXTENSIONS = ['.txt', '.asc', '.htm', '.html', '.pdf', '.doc', '.docx', '.odt', '.jpg', '.png', '.eml']``
+
+- **HELPDESK_VALIDATE_ATTACHMENT_TYPES** If you'd like to turn of filtering of helpdesk extension types you can set this to False.
+
+  
 Generic Options
 ---------------
 These changes are visible throughout django-helpdesk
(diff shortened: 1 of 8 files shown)


helpdesk/settings.py
```diff
@@ -56,6 +56,11 @@
                                          'HELPDESK_STAFF_VIEW_PROTECTOR',
                                          lambda _: None)
 
+# Enable ticket and Email attachments
+HELPDESK_ENABLE_ATTACHMENTS = getattr(settings,
+                                      'HELPDESK_ENABLE_ATTACHMENTS',
+                                      False)
+
```
uhurusurfa, line 63: Unless there is a strong argument to changing a default, I will always err on the side of remaining backwards compatible.
In this case I can see the argument that there is a security risk for new users who do not want attachments enabled but that is still an installation decision new users should make and as such, as long as we clearly document it in the install instructions, I do not see that it is necessary to make the default "False". 
This means that long time existing users will not be surprised by a change they did not expect when they upgrade and find they have lost a chunk of data for their tickets before they realise what the problem is.
I am open to discussion on this if there are strong opinions for making it default to False.

uhurusurfa (changes requested):

helpdesk/settings.py
jorge-leon: No strong argument at all.

I'm always happy to find, that makers of open source package 'X' have my back covered when I start dabbling with 'X', but I also understand, that breaking software with an update will always cause confusion and anger.

I propose to
- Expand the security topic in the documentation
- Set safety default in a future major upgrade with due announcements.

For now I will update the PR to default to enabled attachments.

jorge-leon:
Testing fails with `Queue processing failed: test2 -- Error Q2`, I guess when processing a (mocked?) Email queue.

Can you give pointers how to test/mock Email on the development system in order to look into this?  The pytest output does not give clues and I'm unfamiliar with pytesting.

uhurusurfa:
> Testing fails with `Queue processing failed: test2 -- Error Q2`, I guess when processing a (mocked?) Email queue.
> 
> Can you give pointers how to test/mock Email on the development system in order to look into this? The pytest output does not give clues and I'm unfamiliar with pytesting.

You PR is passing right now but generally you can do this from the Helpdesk project base folder in a shell(I am using *nix notation so use backslash if you are Windows):

1. Run all test by executing quicktest.py ./quicktest.py (or: python3 quicktest.py if python3 is not the default python for your system)
2. Run a specific test use the format: ./quicktest.py <fully qualified path to a test clas or method within a test class>
           eg. ./quicktest.py helpdesk.tests.test_get_email.GetEmailCommonTests.test_email_with_txt_as_attachment_with_simple_alternative_message

NOTE: You must be in a virtualenv shell when running the tests so you can do this to activate a virtualenv from the Helpdesk project base folder assuming you hve set up a virtualenv per the "Developer Environment" instructions in README.rst:
`.  ./.venv/bin/activate`

Alternatively you can use the debugger in whatever

uhurusurfa (approved):
🚀  Looks good to me. @Benbb96  do you have any firther input for this PR?

jorge-leon:
...
>     1. Run all test by executing quicktest.py ./quicktest.py (or: python3 quicktest.py if python3 is not the default python for your system)
> 
... 
> Alternatively you can use the debugger in whatever

Thank you for the instructions.  It looks like Q2 and Q4 are expected to throw an exception during test.

helpdesk/templates/helpdesk/ticket.html
```diff
@@ -197,7 +199,9 @@ <h5>{{ checklist }}</h5>
             </div>
         {% endif %}
 
+        {% if helpdesk_settings.HELPDESK_ENABLE_ATTACHMENTS %}
         <p id='ShowFileUploadPara'><button type="button" class="btn btn-warning btn-sm" id='ShowFileUpload'>{% trans "Attach File(s) &raquo;" %}</button></p>
```
Benbb96, line 203: There are also related parts below in JS (line 278, and 307 to the end) that can be conditionally displayed.

Benbb96 (commented):
One minor thing, either way it looks good to me too. Thanks for the contribution @jorge-leon !

helpdesk/templates/helpdesk/ticket.html
jorge-leon: I hoped to get away without touching JS. But here we are.

Can you confirm that the following lines are the ones related to file handling?
- 278-282: `$("#ShowFileUpload")`...
- 307-334: `$('.add_file_fields_wrap)?`... and `$(document).on('change', ':file'`...

Benbb96: Yes I confirm. These lines shouldn't trigger errors in console but it's better not to have them in DOM if attachments are disabled.

docs/settings.rst
```diff
@@ -51,16 +51,30 @@ These settings can be used to change who can access the helpdesk.
...
+Settings related to attachments:
 
-- **HELPDESK_ANON_ACCESS_RAISES_404** If True, redirects user to a 404 page when attempting to reach ticket pages while not logged in, rather than redirecting to a login screen.
+- **HELPDESK_ENABLE_ATTACHMENTS** If set to ``True``, files can be
+  attached to tickets and followups, and emails are searched for
+  attachments which are then attached to the ticket.  Also enables the
+  ``HELPDESK_ALWAYS_SAVE_INCOMING_EMAIL_MESSAGE`` setting.
 
-  **Default:** ``HELPDESK_ANON_ACCESS_RAISES_404 = False``
+  **Caution**: Set this to False, unless you have secured access to
```
timthelion, line 65: This is false FUD. Unless you have a non-standard Media setup that allows file listing, attachment names are protected by the ticket's secret key and cannot be guessed:

https://github.com/django-helpdesk/django-helpdesk/blob/c77bb30035bfc32d00dc6aab635b38f18e6538c4/helpdesk/models.py#L1218

timthelion:
https://github.com/django-helpdesk/django-helpdesk/commit/caaa0e515844f61b2c47cb29fb320dfcee0d45a5

uhurusurfa (approved):
👍 Looks good to me.

timthelion:
It still contains the mostly false warning:

> 
> **Caution**: Set this to False, unless you have secured access to the uploaded files. Otherwise anyone on the Internet will be able to download your ticket attachments.
> Attachments are enabled by default for backwards compatibility.
> 

Why include lies and FUD in our docs?

A better text would be:

> 
> **Caution**: With most Django Media back-ends attachments can be downloaded by anyone who knows the unique and unguessable URL. If you want greater security choose a Media back-end which is secured or disable attachments by setting this setting to False.

jorge-leon:
I like the wording provided by @timthelion, it makes things clear and shows how to act on.

If all agree and if not done before I'll be happy to 'sneak' it in with some future PR.

PR #1193, opened 2024-07-09
delete_query view should be catching a "pk" field

samsplunks:
Simple BUGFIX as staff.delete_saved_query function is expecting a pk field  (and not an id field).


samsplunks:
Looks like there is an error with ruff linting:
`error: `ruff <path>` has been removed. Use `ruff check <path>` instead.`

Benbb96 (approved):

PR #1194, opened 2024-07-10
Moving python time_spent calculation to database aggregation

samsplunks:
With an outstanding number of followups, the time_spent count in the Queue model is very slow.

By turning over the count at the database level, the calculation was way faster.

To normalize the behavior, I have also modified the time_spent property for the Ticket model.


Benbb96 (approved):
LGTM

uhurusurfa:
You will have to update .github/workflows/pythonpackage.yml call to ruff.

PR #1195, opened 2024-07-16
Add ci workflow for building standalone docker image

timthelion:
Before this is merged we need to set up secrets to do this I can share those secrets or someone can make me a maintainer ;)


timthelion:
Images can be found here https://hub.docker.com/r/djangohelpdesk/standalone/tags

Benbb96 (dismissed):
It seems good to me. And I believe it's ok to make you a maintainer since you contributed a lot to the project! But I let @uhurusurfa or @gwasser take the decision.

timthelion:
Realized this wasn't quite as polished as I had thought changing to draft.

docs/standalone.rst
```diff
@@ -1,7 +1,9 @@
-Django-Helpdesk Standalone Installation
+Standalone Installation
 =======================================
 
-Installation
+You can find standalone docker images at `djangohelpdesk/standalone:latest <https://hub.docker.com/r/djangohelpdesk/standalone/tags>`_.
+
```
Benbb96, line 5: Maybe add a text to explain how to install with remote images and also what does the `standalone-extras` image bring in addition to the initial image.

timthelion: I think that most people who would use this standalone image would be inserting it into docker-compose or kubernetes yaml manifest. Not really sure what I should put there, maybe a docker-compose example?

Benbb96: Alright I see, so yes just an exemple of commands for docker-compose and why would you be using the extra image.

timthelion:
@Benbb96 OK, I've updated the docs :)

Benbb96 (commented):
Alright thanks for the update. I think it's all good to me now.

There's only one thing in the extra requirements that i don't know what it's for: django-keycloak

uhurusurfa (changes requested):
In principal the PR looks fine from a technical perspective.
However, as a rule, open source applications should try not to force users into using technology that may not be their preferred tech stack if it is not necessary to do so and the installed packages in the container should be restricted to only the packages required for the particular use case to minimise memory usage and load time.
This docker script is becoming increasingly specific to a particular deployment and needs to be made more generic so that it is more usable by the broader community.

1. The script is already specific to using a Postgres database and ideally that should be driven by environment variables or command line options that can be used by the docker build script to use different clients for the desired database system the user wants for data persistence.
2. The KeyCloak package should only be included based on a configuration variable.

timthelion:
@uhurusurfa 

1. The database backend is configurable by using your own localsettings.py file as documented here: https://github.com/django-helpdesk/django-helpdesk/pull/1195/files#diff-d8d76a1c758ef3ff139bc3c432d12ce86ca69f6c8b60a8c59457c1176d82c32bR72

 I don't think it makes sense to create a whole new system for using ENV vars to specify various DB backends as Django is already configurable by default.

2. As for keycloak, no one is forcing you to use the `extras` image. The `extras` image is intended for extra stuff that not everyone would need.

You write "the installed packages in the container should be restricted to only the packages required for the particular use case to minimize memory usage and load time" and I think that we have adhered to this principle in the base `Dockerfile` which, by the way, is not part of this PR at all https://github.com/django-helpdesk/django-helpdesk/blob/main/standalone/Dockerfile Is there anything in that image that you consider to be unnecessary?

uhurusurfa:
The base Docker file you referenced installs the Postfgres client on lines 4 and 5 and those have a host of dependent libraries that then get added to the container as well so irrespective of what the user wants to use as a database backend, the container now contains the Postgres client and all its dependent libraries. There should be a way to exclude them without having to modify the Dockerfile or provide a way to install custom packages dependent on the database server the app user prefers.

Given that you are going to push the container to Docker hub since you asked for maintainer status on the project in order to add secrets, the container you are going to push up will contain libraries that are specific to your use case although as you point out, it could be used more generically but will have packages that are likely not going to be used. 

If we are going to be pushing Docker containers up that are for users of Django Helpdesk then the containers should provide a more generic deployment as a baseline and anyone wanting customised versions can then modify the appropriate scripts for generating a Docker container with a minimum of effort.

jorge-leon:
My 2c: If somebody is using Docker to run django-helpdesk than they would be interested in a solution which just works.  Selection of a specific stack might be secondary. So be it Postgres and be it Caddy.

The more you have to configure or specify to set up a standalone django-helpdesk from Docker, the higher the threshold to try it out or adopt it. 

Anybody wanting to selectively swap components of the tech stack will be knowledgeable enough to derive and optimize the Docker configuration files or to set it up from source.

timthelion:
@uhurusurfa actually it's already all documented (including setting up the env vars) here: https://github.com/django-helpdesk/django-helpdesk/blob/main/docs/standalone.rst

`setup.sh` will set up the env vars that you need for postgres.

timthelion:
The only thing this PR really does is that it makes it easier to install the standalone helpdesk by providing pre-built images. The standalone helpdesk is already merged.

uhurusurfa:
I must be missing something - can you point out where it documents which environment variables to set for using an external Postrges DB?
Basically if this is going save someone time then knowing how to use this image without have to hunt around in the code is key.

timthelion:
@uhurusurfa just added more docs for the env vars.

uhurusurfa (approved):
👍  thanks

PR #1198, opened 2024-07-26
Adding custom fields modifications in follow-ups

samsplunks:
In Helpdesk, we have the ability to create custom fields that are only editable in the ticket and are not traced in the follow-ups.

The idea of the PR is to have them when a staff user creates a follow-up.

I also want to introduce a setting to be able to chose the custom fields we want displayed.

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -202,6 +202,9 @@ Options that change ticket updates
 
   **Default:** ``HELPDESK_STAFF_ONLY_TICKET_CC = False``
 
+- **HELPDESK_SHOW_CUSTOM_FIELDS_FOLLOW_UP_LIST** Show configured custom fields in the follow-up form.
+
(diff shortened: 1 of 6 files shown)


timthelion:
Does this show up in the UI somehow?

uhurusurfa (approved):
Looks good to me

docs/settings.rst
```diff
@@ -202,6 +202,9 @@ Options that change ticket updates
 
   **Default:** ``HELPDESK_STAFF_ONLY_TICKET_CC = False``
 
+- **HELPDESK_SHOW_CUSTOM_FIELDS_FOLLOW_UP_LIST** Show configured custom fields in the follow-up form.
```
Benbb96, line 205: I would have add an example to show that fields don't need to be prefixed by `custom_`

PR #1204, opened 2024-09-10
Update create_ticket.html

vritika221:
this line showing the error because  some bootstrap4from is not support on the current version of django


jorge-leon:
As you can see from the checks, your change does not pass the checks.

django-helpdesk is not ready to work with the "current version of django" (5.x at the time of writing).

While I run a django 5.0 adapted django-helpdesk installation, I only use a very restricted set of features and have some compatibility code in a private repository.

A complete update is not trivial to do and might require a separate development branch.

If you are determined to tackle a Django 5.x update of django-helpdesk I'll happily share my current findings.

PR #1210, opened 2024-09-18
Pyproject setup

loydbanks:


uhurusurfa:
@loydbanks  Is this intended to be a PR against the original or was this done in error?

loydbanks:
> @loydbanks Is this intended to be a PR against the original or was this done in error?

It's not, sorry. I wanted to push to our fork but made the mistake of pushing here.

PR #1213, opened 2024-10-18
Refactor custom commands with modern syntax

### LINKED ISSUE #1208: Management commands that use option_list fail.
**Describe the bug**

BaseCommand.option_list is deprecated in Django 1.8 and used in management commands like:
helpdesk/management/commands/create_queue_permissions.py

**To Reproduce**
1. python manage.py create_queue_permissions
2. returns error: `AttributeError: 'Command' object has no attribute 'option_list'`

**Expected behavior**
create_queue_permissions

Benbb96:
Fixes #1208 

There's a breaking change because `create_escalation_exclusions` and `escalate_tickets` were able to be run without using Django manage script. But these custom commands are meant to be run in a Django environment.

### DIFF
--- a/helpdesk/management/commands/create_escalation_exclusions.py
+++ b/helpdesk/management/commands/create_escalation_exclusions.py
@@ -32,7 +32,7 @@ class Command(BaseCommand):
             nargs='*',
             choices=list(day_names.keys()),
             required=True,
-            help='Days of week (monday, tuesday, etc)'
+            help='Days of week (monday, tuesday, etc). Enter the days as space separated list.'
         )
         parser.add_argument(
             '-o',
@@ -46,7 +46,7 @@ class Command(BaseCommand):
             '--queues',
             nargs='*',
             choices=list(Queue.objects.values_list('slug', flat=True)),
-            help='Queues to include (default: all). Use queue slugs'
+            help='Queues to include (default: all). Enter the queues slug as space separated list.'
         )
         parser.add_argument(
             '-x',
--- a/helpdesk/management/commands/create_escalation_exclusions.py
+++ b/helpdesk/management/commands/create_escalation_exclusions.py
@@ -11,63 +11,7 @@ scripts/create_escalation_exclusion.py - Easy way to routinely add particular
 
 from datetime import date, timedelta
 from django.core.management.base import BaseCommand, CommandError
-import getopt
 from helpdesk.models import EscalationExclusion, Queue
-from optparse import make_option
-import sys
-
-
-class Command(BaseCommand):
-
-    def __init__(self):
-        BaseCommand.__init__(self)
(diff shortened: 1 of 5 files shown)


helpdesk/management/commands/create_escalation_exclusions.py
```diff
@@ -80,79 +24,70 @@ def handle(self, *args, **options):
...
-
+class Command(BaseCommand):
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '-d',
+            '--days',
+            nargs='*',
+            choices=list(day_names.keys()),
+            required=True,
+            help='Days of week (monday, tuesday, etc)'
```
uhurusurfa, line 35: Might be worth being a little more explicit about the format... something like:
'Days of week (monday, tuesday, etc). Enter the days as a space separted list.'

helpdesk/management/commands/create_escalation_exclusions.py
```diff
@@ -80,79 +24,70 @@ def handle(self, *args, **options):
...
+            default=1,
+            type=int,
+            help='Occurrences: How many weeks ahead to exclude this day'
+        )
+        parser.add_argument(
+            '-q',
+            '--queues',
+            nargs='*',
+            choices=list(Queue.objects.values_list('slug', flat=True)),
+            help='Queues to include (default: all). Use queue slugs'
```
uhurusurfa, line 49: Similar to the days args .... specify the list is space separated.

helpdesk/management/commands/create_queue_permissions.py
```diff
@@ -15,55 +15,54 @@
...
-            make_option(
-                '--queues', '-q',
-                help='Queues to include (default: all). Use queue slugs'),
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '-q',
+            '--queues',
+            nargs='*',
+            choices=list(Queue.objects.values_list('slug', flat=True)),
+            help='Queues to include (default: all). Use queue slugs'
```
uhurusurfa, line 32: Same as other help related messages above about space separated values.

uhurusurfa (dismissed):
Just the comments about the help text but if you feel the text is adequate then go ahead and merge this.

helpdesk/management/commands/create_escalation_exclusions.py
Benbb96: Yes, good point, I updated help texts.

uhurusurfa (approved):
🚀

Benbb96:
I'm adding the "_hacktoberfest-accepted_" label just to try if it works. I'm not sure if I'll have the time to make 4 PRs until the end of the month. But this event could be something to think of if we want to attract more people in the project 😃

PR #1227, opened 2024-12-21
Entrypoint adapted to be run multiple times and --noreload arg added to runserver to stop the container with CTRL-C

feroda:


uhurusurfa (approved):
👍 Thanks for the enhncement

PR #1234, opened 2025-01-21
WIP: Removing filtering_or pending clarification

### LINKED ISSUE #899: Date Range filter on ticket list doesn't work
When doing my PR #898, I tested filters and it appears that the date range was not respected under certain circonstances.

**To Reproduce**
1. Go to Ticket List
2. Add Date Range filter box
3. Add some values for From and To dates
4. Apply filter

**Expected behavior**
Tickets should be filtered by these dates.

**Screenshots**
image

 - OS: Windows
 - Browser: Google chrome
 - Version: 86.0.4240.111

**Additional context**
I think there is a conflict with other filters (like the status) during the SQL query. **Because if I remove the status filter, it then works as expected**.

DavidVadnais:
While trying to fix #899 I noticed that the `filtering_or` always creates a scenario were it ignores all the filters.   This commit does fix the behavior I was seeing however, it does not seem to be the intended fix.


DavidVadnais:
With the original code I added the following prints:

```python
print(f"filter_or = {filter_or}")
print(f"filter = {filter}")
```

When all other or any grouping of the filters on filtering tickets page is selected these return the same value. When date range is all there filter_or returns the filter without the date range. Example
```python
filter_or = {'assigned_to__id__in': [4], 'status__in': [1, 2, 3]}
filter = {'assigned_to__id__in': [4], 'status__in': [1, 2, 3], 'created__gte': '2024-10-01', 'created__lte': '2024-10-30'}
```

Because of this it is fair to conclude that the original code is working as intended. 
```python
queryset = queryset.filter((Q(**filter) | Q(**filter_or)) & self.get_search_filter_args())
```

However the problem still remains somewhere upstream of [query.py](https://github.com/django-helpdesk/django-helpdesk/blob/main/helpdesk/query.py).

uhurusurfa:
There are 2 problems with the filtering:

1. The  `__Query__` classes "**get_search_filter_args(self)**" _method_  in query.py invokes the "**get_search_filter_args(search)**" _function_ in query.py with an empty string if no search params are passed. This causes invalid filters to be applied. to fix this add this to the entry of the function : 
```
    if not search:
        return Q()
```
2. The filter_or is supposed to support searching for fields where they value is empty (NULL). For instance if you choose the "Owner" filter then one of the options is "Unassigned" which has a value of -1. However, the implementation is really messy and does not include the date filters in the OR filter so it does not work as expected and needs to be completely reworked so that it produces the correct query and also an efficient query.

DavidVadnais:
Will be fixed by #1241

PR #1237, opened 2025-02-03
Issue #1235 - Escalation only notify users

### LINKED ISSUE #1235: In ticket escalation "notify only" option may be desired
Escalations can clutter tickets. 

My team wants the escalation option to not clutter the follow ups field with entries.  I will implement an argument that removes the followup section of escalation but leaves the email section. 

`python manage.py  escalate_tickets --notify_only`

I could also make another command.

**Additional context**
I will be implementing this locally.  Do you want me to upstream this feature or do you believe other users would not desire this capability?

DavidVadnais:
Fixes #1235. Allows users to add a flag to the escalation command to only notify users and not make a followup entry or increase priority.

I tested this in my local environment but environment is slowly pulling away from the default one.


DavidVadnais:
I am not sure were to add this feature in the docs.  Is [configuration page](https://django-helpdesk.readthedocs.io/en/v1.3.0/configuration.html) the correct page?

Benbb96:
> I am not sure were to add this feature in the docs. Is [configuration page](https://django-helpdesk.readthedocs.io/en/v1.3.0/configuration.html) the correct page?

Yes that seems like the right place.

Something like  "Adding the flag `--notify-only` will only send the emails and no followup will be added to the ticket" will do I suppose

DavidVadnais:
Here is a screen grab of the docs after running `make html` in the docs directory. 

2025-03-08_helpdesk-doc-configuration

Please let me know if this was the correct way to edit the documentation, as I’m not very familiar with rst.

Benbb96:
Yes that's the correct way of doing it. I would just add backticks around the `--notify-only` flag. And IMO, there's no need to add another step. you could just put your explanation in the 4th point since it is about the same command.

uhurusurfa (dismissed):
👍  Looks good

DavidVadnais:
I moved the new field to bullet point four and added back-ticks around `--notify-only`. I removed the `--` from the sentence because when it rendered it appeared as one dash. I have attached a screenshot of the docs page after running `make html`. 

2025-03-10_helpdesk-doc-configuration

Let me know if anything else is needed!

Benbb96 (approved):
It looks good to me, thanks for this new feature!

PR #1232, opened 2025-01-18
Fix for my-tickets div not appending to DOM

DavidVadnais:
Fix for issue described in issue #1231 

edit: for linking Fixes #1231

PR #1239, opened 2025-03-03
BUG - Bug workflow ubuntu latest deprecated 

DavidVadnais:
Workflows are failing due to the pending deprecation of workflow v2 https://github.blog/changelog/2024-12-05-notice-of-upcoming-releases-and-breaking-changes-for-github-actions/#actions-cache-v1-v2-and-actions-toolkit-cache-package-closing-down

Merging this into the merge requests #1232 and #1234 should fix the respective workflow issues.

### DIFF
--- a/.github/workflows/build-standalone-docker-image.yml
+++ b/.github/workflows/build-standalone-docker-image.yml
@@ -7,11 +7,11 @@ on:
 
 jobs:
   build-and-push:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu 24
 
     steps:
     - name: Checkout repository
-      uses: actions/checkout@v2
+      uses: actions/checkout@v4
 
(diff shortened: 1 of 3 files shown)


uhurusurfa (commented):
Is there a specific reason you changed "ubuntu-latest" to "ubuntu 24" because "ubuntu-latest" is "ubuntu-24" ATM and the whole reason for using "ubuntu-latest" is to avoid having to update it every time a new version of ubuntu is deplyed.
This app should be independent of which version of Ubuntu it is deployed on.

DavidVadnais:
I just switched it back I thought I read that they were removing latest for specific tags but I misread the section were they are updating latest to 24.04. See runner images: https://github.com/actions/runner-images/

uhurusurfa (approved):
🚀 Thanks for fix

PR #1241, opened 2025-03-06
Ticket query fixes

uhurusurfa:
Fix multiple issues with filtering in the staff view.

### DIFF
--- a/helpdesk/query.py
+++ b/helpdesk/query.py
@@ -28,14 +28,18 @@ def query_from_base64(b64data):
 
 
 def get_search_filter_args(search):
+    if not search:
+        return Q()
     if search.startswith('queue:'):
         return Q(queue__title__icontains=search[len('queue:'):])
     if search.startswith('priority:'):
         return Q(priority__icontains=search[len('priority:'):])
-    filter = Q()
+    my_filter = Q()
     for subsearch in search.split("OR"):
         subsearch = subsearch.strip()
-        filter = (
+        if not subsearch:
+            continue
+        my_filter = (
             filter |
             Q(id__icontains=subsearch) |
             Q(title__icontains=subsearch) |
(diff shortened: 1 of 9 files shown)


uhurusurfa:
@DavidVadnais - this is the fix for the query filter issues

helpdesk/query.py
```diff
@@ -91,23 +95,49 @@ def get_search_filter_args(self):
...
-        Apply a dict-based set of filters & parameters to a queryset.
+        Apply a dict-based set of value_filters & parameters to a queryset.
 
         queryset is a Django queryset, eg MyModel.objects.all() or
             MyModel.objects.filter(user=request.user)
 
         params is a dictionary that contains the following:
-           filtering: A dict of Django ORM filters, eg:
+           filtering: A dict of Django ORM value_filters, eg:
             {'user__id__in': [1, 3, 103], 'title__contains': 'foo'}
```
Benbb96, line 105: Maybe add an example here with null values ?

helpdesk/query.py
```diff
@@ -91,23 +95,49 @@ def get_search_filter_args(self):
...
+                        # Remove the matching filters into a Q param
+                        matched_null_keys.append(null_key)
+                        # Create an OR query for the selected value(s) OR if the field is NULL
+                        v = {}
+                        v[val_key] = value_filters[val_key]
+                        n = {}
+                        n[null_key] = null_filters[null_key]
+                        q_args.append((Q(**v) | Q(**n)))
+                        del value_filters[matched_key]
+                # Now remove the matched mull keys
```
Benbb96, line 136:
```suggestion
                # Now remove the matched null keys
```

Benbb96 (commented):
The code in `query.py` is a bit hard to follow but it looks good to me and there are tests (maybe add another one for more complex filters ?)

uhurusurfa:
@Benbb96  Making mnore complex tests would be best.
Unfortunately it took long enough just to fix the mess and making more records would require reworking the existing tests to accommodate adding more records.
I will try to come back to this and add better unit tests when I have more time but for now this fixes a broken filtering system.

uhurusurfa:
@Benbb96 - can you please approve this PR when you have a moment.

Benbb96 (approved):

PR #1245, opened 2025-03-12
Enhancement sort tickets by last followup

### LINKED ISSUE #1243: Sort tickets by most recent follow-up in the ticket_list template
My users want to be able to sort tickets by which tickets have had a most recent follow-up. 

A new column visibility option on the [ticket list](https://github.com/django-helpdesk/django-helpdesk/blob/main/helpdesk/templates/helpdesk/ticket_list.html) page that allows for sorting by most recent follow-up. This would work like the github filter **Last updated** (see image).

Image

You could maybe do a follow-up feed that shows all most recent follow-ups and there corresponding tickets.  That might satisfy the same use case.

DavidVadnais:
Fixes #1243 "A new column visibility option on the [ticket list](https://github.com/django-helpdesk/django-helpdesk/blob/main/helpdesk/templates/helpdesk/ticket_list.html) page that allows for sorting by most recent follow-up. This would work like the github filter Last updated."


DavidVadnais:
# Questions
1. Do we want the column to show by default? Currently I have it set this way.
2. Do we want to do something with the found unused variable in views/staff.py?
3. Are there any documentation ideas for a feature like this? 

# Follow up
*    clean this up
pyflakes3 finds the below unused local variable but it is outside the scope of this pull request. 
```
pyflakes3 query.py serializers.py tests/test_query.py views/staff.py 
views/staff.py:600:5 local variable 'time_spent' is assigned to but never used
```

*    fix the unit tests
Done and this field is exercised by the updated tests.

*    run additional smoke tests
Hid and then re-displayed the column last followup using the column visibility button. 
Sorted in both directions the last follow up column and attempt to sort this column after sorting on several other fields. 
Tested that the default to blank works for tickets with no follow ups

*    documentation
I don't see a good location for documenting this type of feature.

uhurusurfa (changes requested):
Just a few comments inline.

helpdesk/serializers.py
```diff
@@ -70,8 +73,16 @@ def get_row_class(self, obj):
 
     def get_kbitem(self, obj):
         return obj.kbitem.title if obj.kbitem else ""
-
-
+        
+    def get_last_followup(self, obj):
+        try:
+            followup = obj.followup_set.latest('date')
```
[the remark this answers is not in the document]
DavidVadnais: I think I got the soul of what you're recommending but in case I missed something here is the path that leads me to a slightly different code snippet.

First I switched `object` to `obj` so that we could access followup_set. 

This fixed the error `AttributeError: type object 'object' has no attribute 'followup_set'`

Second I added `flat=True` to `.values_list("last_followup")` so I would get a datetime object.

This fixed the error output `AttributeError: 'tuple' object has no attribute 'utcoffset'`

DavidVadnais:
I am unable to recreate the three test failures locally so this may take a more commits then I would like.

```python
python --version
Python 3.11.10

python quicktest.py helpdesk.tests.test_query
Found 3 test(s).
Creating test database for alias 'default'...
System check identified no issues (0 silenced).
...
----------------------------------------------------------------------
Ran 3 tests in 1.671s

OK
Destroying test database for alias 'default'...
```

helpdesk/serializers.py
```diff
@@ -70,8 +73,19 @@ def get_row_class(self, obj):
...
+    def get_last_followup(self, obj):
+        followup = obj.followup_set.order_by().annotate(
+            last_followup=Window(
+                expression=Max("date"),
+                partition_by=[F("ticket_id"),],
+                order_by="-date"
+            )
+        ).values("last_followup").distinct().values_list("last_followup", flat=True)
+        # If there are no followups for the ticket then the result will be empty
+        return localtime(followup[0]).strftime('%Y-%m-%d %H:%M:%S') if followup else ""
```
Benbb96, line 86: Maybe you could format this like the `created` and the `due_date` fields, using `humanize.naturaltime` ?

Not really sure if that's the way to go, formatting might be better to be applied in the frontend (ticket_list.html) and not in the serializer...

uhurusurfa:
Try remoiving Django 3.x from the test runs - I have a feeling this is related to the order_by requiring a parameter:
https://docs.djangoproject.com/en/3.2/ref/models/querysets/#order-by

In truth, since Django 3.x is now conasidered insecure we should stop supporting it and support Django 4 and 5 in the tests.

Benbb96 (commented):
Below are some of my interrogations.

Concerning the error in the test. I think it's related to the Window function. There seems to be a a change in the API in [Django 4.1](https://docs.djangoproject.com/en/4.1/ref/models/expressions/#window-functions), now accepting `order_by `parameter to be a string with an optional "-" prefix.

In order to be backward compatible, I think we should use an expression (something like `F('date').desc()` I believe)

helpdesk/serializers.py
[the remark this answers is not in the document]
uhurusurfa: Looks good

helpdesk/views/staff.py
```diff
@@ -1025,12 +1025,14 @@ def ticket_list(request):
             ('assigned_to', 'assigned_to__id__in'),
             ('status', 'status__in'),
             ('kbitem', 'kbitem__in'),
+            ('last_followup', 'last_followup__in'),
```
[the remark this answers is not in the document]
uhurusurfa: Agreed - maybe it can be used to find open tickets that have not had a followup since a user selected date.

helpdesk/serializers.py
uhurusurfa: Agreed on where to format.
Personally I do not like the humanize format but can see that is is useful under certain circumastances depending on how you are trying to review the data.

helpdesk/views/staff.py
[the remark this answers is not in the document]
DavidVadnais: Good catch. I was initially going to add a filter option but I backed off of it.

helpdesk/serializers.py
DavidVadnais: Is this what you had in mind?

[example_attachment.txt](https://github.com/user-attachments/files/19329460/example_attachment.txt)

DavidVadnais:
django 5 is not compatible with python 3.8 and 3.9 https://docs.djangoproject.com/en/5.1/releases/5.0/

cgi is removed if you try and go to python 3.13 https://docs.python.org/3/library/cgi.html

DavidVadnais:
I got the tests to be able to keep python 3.8 and 3.9 with django 5 and I got python 3.12 running but I have to complete another task so I backed off the django 5 add until later or another pull request

helpdesk/serializers.py
uhurusurfa: Looks fine to me ... @Benbb96  any input?

helpdesk/query.py
```diff
@@ -169,13 +170,13 @@ def get_datatables_context(self, **kwargs):
...
         order = kwargs.get('order[0][dir]', ["asc"])[0]
-
+        
         order_column = DATATABLES_ORDER_COLUMN_CHOICES[order_column]
         # django orm '-' -> desc
         if order == 'desc':
             order_column = '-' + order_column
 
-        queryset = objects.all().order_by(order_by)
+        queryset = objects.annotate(last_followup=Max('followup__date')).order_by(order_by)
```
uhurusurfa, line 179: Thought I had added a further comment on this but it seems to have vanished.

The query it produces is going to be hugely inefficient for large datasets and will not work if the ordering is not on `followup__date`.

It will have to be something like this:
```
        queryset = objects.annotate(
            last_followup=Subquery(
                FollowUp.objects.order_by().annotate(
                    last_followup=Window(
                        expression=Max("date"),
                        partition_by=[F("ticket_id"),],
                        order_by="-date"
                    )
                ).filter(
                    ticket_id=OuterRef("id")
                ).values("last_followup").distinct()
            )
        ).order_by(order_by)
```
This produces a query  to render the "last_followup" field as a subquery on the final result set instead of forcing a GROUP BY across every field on every model in the query and is immune to changes in the resultset sorting order.

uhurusurfa (changes requested):
Query fix below.

helpdesk/serializers.py
Benbb96: Yes it looks good but I think the default locale should be taken into account instead of forcing 'en-US'. As well as for the "No followup found" that should be translatable

DavidVadnais: I think I have the locale handled.  Is our location choice variable LANGUAGE_CODE or do we use something else?

DavidVadnais:
also while here I stuck in this change.  Do you disagree with it? 

Before 
before_reports_2025-03-19

After
after_reports_2025-03-19

uhurusurfa:
Looks much better with just 2 decomal digits.

uhurusurfa (commented):
I will try to pull this down and give it a smoke test some time later today.

helpdesk/templates/helpdesk/ticket_list.html
```diff
@@ -415,6 +422,27 @@ <h5 class="mb-0">
                         }
                     },
                     {data: "submitter"},
+                    {
+                        data: "last_followup",
+                        render: function (data, type, row) {
+                            let locale = window.helpdesk_settings.LANGUAGE_CODE;
```
uhurusurfa, line 428: Change this to:
let locale = navigator.language || navigator.userLanguage || window.helpdesk_settings.LANGUAGE_CODE;

It will allow rendering of the timestamp in the users browser locale which will generally be wherever the user is located in the world.

helpdesk/templates/helpdesk/ticket_list.html
```diff
@@ -330,7 +331,13 @@ <h5 class="mb-0">
     {% else %}
     <script src="{% static 'helpdesk/vendor/timeline3/js/timeline.js' %}"></script>
     {% endif %}
+
     <script>
+        window.helpdesk_settings = {
+            LANGUAGE_CODE: "{{ helpdesk_settings.LANGUAGE_CODE|default:'en-US' }}",
+            NO_FOLLOWUP_TEXT: "{{ helpdesk_settings.TRANSLATIONS.No_followup_found|default:'No followup found' }}"
```
uhurusurfa, line 338: Change this to:
`NO_FOLLOWUP_TEXT: "{% trans helpdesk_settings.TRANSLATIONS.No_followup_found|default:'No followup found' %}"`
It will ensure that the string can be translated.
Possibly change the settings based alternative for the english string to use something other than TRANSLATIONS since it is not actually translating the string - just providing an alternative string to the default.... maybe ALTERNATIVE_UI_STRINGS or something along those lines that provides a clearer idea of what it is doing.

helpdesk/views/staff.py
```diff
@@ -1061,14 +1061,22 @@ def ticket_list(request):
...
             query_params['filtering']['created__lte'] = date_to
 
+        followup_from = request.GET.get('followup_from')
+        if followup_from:
+            query_params['filtering']['followup__gte'] = followup_from
+
+        followup_to = request.GET.get('followup_to')
+        if followup_to:
+            query_params['filtering']['followup__lte'] = followup_to
+
```
uhurusurfa, line 1071: Currently this is redundant since there is no implementation in the form - are you going to add an implementation?

helpdesk/serializers.py
uhurusurfa, line 86: Actually there is no need for this lookup at all since the primary query already does the lookup.
Just this will work and no wasting another DB lookup:
```
    def get_last_followup(self, obj):
        return obj.last_followup
```

uhurusurfa (changes requested):
Just the issues mentioned inline ... functionally it works well.

helpdesk/views/staff.py
DavidVadnais: I was planning to but Ive backed off the idea.  Ill remove this

DavidVadnais:
I think I have addressed all of the review improvements requested. I ran the code locally and it seemed to be working. Let me know if I missed anything.

helpdesk/templates/helpdesk/ticket_list.html
```diff
@@ -330,7 +331,13 @@ <h5 class="mb-0">
     {% else %}
     <script src="{% static 'helpdesk/vendor/timeline3/js/timeline.js' %}"></script>
     {% endif %}
+
     <script>
+        window.helpdesk_settings = {
+            LANGUAGE_CODE: "{{ helpdesk_settings.LANGUAGE_CODE|default:'en-US' }}",
+            NO_FOLLOWUP_TEXT: "{% trans helpdesk_settings.TRANSLATIONS.No_followup_found|default:'No followup found' %}"
```
uhurusurfa, line 338: Use something other than TRANSLATIONS since it is not actually translating the string - just providing an alternative string to the default.... maybe ALTERNATIVE_UI_STRINGS or something along those lines that provides a clearer idea of what it is doing.

uhurusurfa (changes requested):
Just the one minor change please. Otherwise looks great.

helpdesk/templates/helpdesk/ticket_list.html
DavidVadnais: I made the change but I am not able to test it easily right now so I did not run any form of smoke test on the commit. Ill be able to test it more thoroughly either Sunday or Monday.

uhurusurfa (approved):
Looks good and works for me

PR #1254, opened 2025-04-04
Fix hide due_date when HELPDESK_PUBLIC_TICKET_DUE_DATE set to None

### LINKED ISSUE #1228: HELPDESK_PUBLIC_TICKET_DUE_DATE = None doesn't work as described
**Describe the bug**
In the docs, it says the following:

> **HELPDESK_PUBLIC_TICKET_DUE_DATE** Sets the due date for tickets submitted through the public form. If defined, the matching form field will be hidden. Set to None if you want to hide the form field but do not want to define a value.

But I have this in my settings.py file:

`HELPDESK_PUBLIC_TICKET_PRIORITY = 3`
`HELPDESK_PUBLIC_TICKET_DUE_DATE = None  # hide the due date field on the ticket creation form`

But the form is still showing the due date field (see the screenshot below). If I put `datetime.now()` as the value rather than `None`, the field disappears as you would expect so that part of the description checks out.

Note that the ticket priority field is not showing which is working as described in the docs.

**To Reproduce**
Add `HELPDESK_PUBLIC_TICKET_DUE_DATE = None` to `settings.py` and then go to the public home page which shows the submit ticket form.

**Expected behavior**
The **Due on** field should not be displayed.

**Screenshots**

image

 - OS: macOS 14.6.1
 - Safari 
 - 17.6 (19618.3.11.11.5)

**Additional context**
This is a brand new install. Here are my settings thus far:

`HELPDESK_TEAMS_MODE_ENABLED = False`
`HELPDESK_PUBLIC_TICKET_PRIORITY = 3`
`HELPDESK_PUBLIC_TICKET_DUE_DATE = None  # hide the due date field on the ticket creation form`

uhurusurfa:
This fixes #1228

### DIFF
--- a/helpdesk/forms.py
+++ b/helpdesk/forms.py
@@ -545,7 +545,10 @@ class PublicTicketForm(AbstractTicketForm):
 
         for field_name, field_setting_key in field_deletion_table.items():
             has_settings_default_value = getattr(settings, field_setting_key, None)
-            if has_settings_default_value is not None:
+            if has_settings_default_value is not None or (
+                "HELPDESK_PUBLIC_TICKET_DUE_DATE" == field_setting_key
+                and hasattr(settings, field_setting_key)
+            ):
                 del self.fields[field_name]
 
         public_queues = Queue.objects.filter(allow_public_submission=True)


DavidVadnais (approved):
Great.

Issue #1259, opened 2025-04-18
ticket_list sorting behaviour is unexpected


finnertysea:
**Describe the bug**
On the /tickets/ page, the sorting filter (including reverse) does not change the table's data.

**To Reproduce**
1. Go to /tickets/
2. Select a value for "sorting"
3. Submit the filters form
4. See that the data table does not react as expected

**Expected behavior**
The table should be sorted according to the selected filter.

**Additional context**
In sorting.html, the name for the sorting select is "sortx", not matching "sort" as it's referred to in ticket_list(). Once changed to match, the ticket_list() view correctly assembles a query set for the chosen "sorting" value, but when get_datatables_context() is called, it overwrites any ordering not done by datatables. Pull request to follow with a proposed fix.

uhurusurfa:
@finnertysea - if you just change line 452 in helpdesk/templates/helpdesk/ticket_list.html
to this then it will ensure there is no ordering on initalisiation of the table and will therefore use the default ordering specified by the "Sorting" selector:
```
                    {% endif %}
                ],
                order: [  ]
            });
```

ie. Just add a comma after the closing square bracket and then the empty order key.
That way it will use the user defined sorting unless the user clicks on one of the column headings within the data table.

PR #1260, opened 2025-04-18
#1259 -  fix ticket_list sorting

finnertysea:
#### changes to sorting.html:
- make id and name of select element match name of param in ticket_list() view.

#### changes to get_datatables_context():
- adjust defaults to get information from existing query set if available while still changing with any new sorting done by datatables.
- remove first .orderby() since it is made useless by second use of .orderby() right before return.


DavidVadnais:
I just confirmed that sorting does not work in the latest commit 1f22f545b5cafc38ac2a2375be64e12ab72c478d. I am not going to test the fix.

DavidVadnais:
From what I can tell the issue persists. I likely need more context. I believed the goal was to fix the reverse button. I would guess that it would yield a different ticket ticket order but it did not for me.
tickets-no_reverse-2025-04-21
tickets-reverse-2025-04-21

I need to run to meeting but I will provide more context ASAP. I cloned your repo and built a docker image with `docker build --progress=plain -t djangohelpdesk/standalone:latest -f standalone/Dockerfile .` then I went into standalone and performed `docker compose up`

finnertysea:
It's working for me. To verify, I did a clean install from the latest revision in my helpdesk repo. I have some custom styles in my dev env which is why the filters are tabbed, but that shouldn't be impacting their functionality at all.

When /tickets/ is first loaded:
Screenshot 2025-04-24 at 10 11 40

When sorting="title" and the form is submitted:
Screenshot 2025-04-24 at 10 12 26

And when sorting="title" and reverse is checked, then the form submitted:
Screenshot 2025-04-24 at 10 12 39

DavidVadnais:
I still cant replicate the success. Can you tell me what database you're using and the python/django versions?

The standalone uses postgresql 17 and 
``` bash
python --version
Python 3.11.12

pip list 
...
Django                 5.2
django-appconf         1.1.0
django-bootstrap4-form 4.0.2
django-cleanup         9.0.0
django-helpdesk        1.4.0       /opt/django-helpdesk
django-model-utils     5.0.0
django-reversion       5.1.0
django-user-accounts   3.3.2
djangorestframework    3.16.0
...
```

Also I'm going to bring in the big guns as they might have a better idea of whats happening. 
@uhurusurfa, @timthelion, @Benbb96

uhurusurfa:
I will be completely offline for the next 12 days but will review as soon
as I can.

helpdesk/query.py
```diff
@@ -162,13 +165,18 @@ def get_datatables_context(self, **kwargs):
...
         search_value = kwargs.get("search[value]", [""])[0]
-        order_column = kwargs.get("order[0][column]", ["5"])[0]
-        order = kwargs.get("order[0][dir]", ["asc"])[0]
+
+        sorting = self.params.get("sorting", "created")
+        default_order_col = DATATABLES_COLUMN_NUM_LOOKUP.get(sorting, "5")
+        sortreverse = self.params.get("sortreverse", None)
+        default_order = "desc" if sortreverse else "asc"
+
+        order_column = kwargs.get("order[0][column]", [default_order_col])[0]
```
uhurusurfa, line 178: This will never use the `default_order_col` because the DataTable defaults to the `id` field.

uhurusurfa (changes requested):
The changes do address the issue where the selected sorting order in the "Sorting" dropdown is sel;ected but it does n ot solve the problem of applying that sorting selection to the data table.
The problem is that the data tablke query checks to see if the data table has passed back a sorting column in the query and uses that if it exists otherwise will try to use the column selected in the "Sorting" filter.
Since the DataTable defaults to ordering on the "id" field it will never use the "Sorting" value and the only way to change the sorting is to click on the column title within the DataTable --- this works as expected.

The use of the "sorting" field is perhaps only useful in setting an initial sort order for the DataTable but it would require pasing an extra query  parameter into the `datatables_ticket_list` URL callback toat populates the data table to tell it to use the default instead of the query param value passed back by the DataTable handler.

uhurusurfa (approved):
Thanks for the fix. I will merge this and raise a firther PR to fix the datatable initial sort to use the selected user sort.

PR #1264, opened 2025-05-09
Fix cc crash when CC email address is assocaited with multiple active User records

### LINKED ISSUE #1096: create_ticket_cc generates MultipleObjectsReturned Exception when multiple users have same email address
The create_ticket_cc function attempts to build a User instance based on the cc email address. However, if there happens to be more than one use registered in Django that has the same email address, this function fails and if this is being actioned from a scheduled task monitoring for email tickets there is the potential for it to block any following email tickets from being processed.

Error occurs at line 485 in email.py

```
        try:
            user = User.objects.get(email=cced_email)  # @UndefinedVariable
        except User.DoesNotExist:
            pass

```

The quick fix was to find the users with the same email address (one was a test user) and either delete them or adjust the email address to be unique. The Django user entry seems to allow duplicate email addresses.

Alternative fix is to monitor for multiple results and respond accordingly and to somehow park the inbound email so that it is no processed again until a unique email address can be sorted out.

uhurusurfa:
Fixes #1096


docs/configuration.rst
```diff
@@ -94,4 +94,12 @@ You may add your own site specific navigation header to be included inside the <
 
     helpdesk/custom_navigation_header.html
 
-2. Update the contents to display your custom navigation.
\ No newline at end of file
+2. Update the contents to display your custom navigation.
+
+Suppressable Log Messages
+-------------------------
+Some logging messages support being switched on or off accroding to deplpoyment preferences.
```
DavidVadnais, line 101: typo : "according to deployment"

docs/configuration.rst
```diff
@@ -94,4 +94,12 @@ You may add your own site specific navigation header to be included inside the <
...
+
+The following settings variables control emiting log messages for specific scenarios:
```
DavidVadnais, line 103: Typo: emiting --> emitting

helpdesk/tests/test_ticket_submission.py
```diff
@@ -480,6 +482,44 @@ def test_create_ticket_from_email_with_invalid_carbon_copy(self):
...
+            )
+        # Check that the CC duplicate was still sent a notification email
+        email_count = len(mail.outbox)
+        found = False
+        for i in range(0, email_count - 1):
+            if dup_email in mail.outbox[i].to:
+                found = True
+                break
+        self.assertTrue(
+            found, "The duplicated email across user ID's was not sent a n otification."
```
DavidVadnais, line 520: extra space after n in word `n otification`

helpdesk/tests/test_ticket_submission.py
```diff
@@ -480,6 +482,44 @@ def test_create_ticket_from_email_with_invalid_carbon_copy(self):
...
+                MULTIPLE_USERS_SAME_EMAIL_MSG in dup_warn.output[0],
+                "The duplicated email across user ID's was was not logged.",
+            )
+        # Check that the CC duplicate was still sent a notification email
+        email_count = len(mail.outbox)
+        found = False
+        for i in range(0, email_count - 1):
+            if dup_email in mail.outbox[i].to:
+                found = True
+                break
```
DavidVadnais, line 518: Would one of the options I present below be better? Or is this the preferred notation?

Now:
```python
        found = False
        for i in range(0, email_count - 1):
            if dup_email in mail.outbox[i].to:
                found = True
                break
```

Option 1: We don't need `i`
``` python
found = any(dup_email in email.to for email in mail.outbox[:email_count - 1])
```

Option 2: We need `i`
``` python
found = False
for i, email in enumerate(mail.outbox[:email_count - 1]):
    if dup_email in email.to:
        found = True
        break
```

docs/configuration.rst
```diff
@@ -94,4 +94,12 @@ You may add your own site specific navigation header to be included inside the <
 
     helpdesk/custom_navigation_header.html
 
-2. Update the contents to display your custom navigation.
\ No newline at end of file
+2. Update the contents to display your custom navigation.
+
+Suppressable Log Messages
```
DavidVadnais, line 99: typo : Suppressable --> Suppressible

DavidVadnais (changes requested):
This looks great. I have run into this issue a bunch and this will definitely help in those cases.

docs/configuration.rst
uhurusurfa: Fixed

docs/configuration.rst
uhurusurfa: Fixed

helpdesk/tests/test_ticket_submission.py
uhurusurfa: Fixed

helpdesk/tests/test_ticket_submission.py
uhurusurfa: Chanfged to option1 - nice and clean

docs/configuration.rst
uhurusurfa: Fixed

DavidVadnais (approved):
Great!

PR #1266, opened 2025-05-13
Django 5 upgrade

uhurusurfa:
Changes to support Django 5


helpdesk/tests/test_api.py
```diff
@@ -15,10 +14,15 @@
     HTTP_403_FORBIDDEN,
 )
 from rest_framework.test import APITestCase
+from _datetime import timedelta
+from helpdesk.lib import convert_value
+from django.utils import timezone
+
+frozen_date_time_str = (datetime.datetime.now() - timedelta(days=100)).isoformat()
```
Benbb96, line 21: I believe the error from the test comes from the `isoformat` function. It does not add the `Z` at the end.

uhurusurfa: It depends on which platform the test is run. I have tried a few dsifferent options and traced the execution through but not identified why sometomes the folowup_set.due_date is rendered into a string from the FakeDateTime object that freezegun creates.
For now the hack works but perhaps at some point the underlying reason will be identified.

Benbb96: Oh yes I see, that's a strange behaviour

helpdesk/models.py
```diff
@@ -2128,6 +2122,8 @@ def build_api_field(self):
             attributes["max_digits"] = self.max_length
         elif self.data_type == "list":
             attributes["choices"] = self.get_choices()
+        # elif self.data_type == "datetime":
+        #     attributes["default_timezone"] = getattr(settings, "DEFAULT_TIMEZONE", "UTC")
```
Benbb96, line 2126: Is it useful to keep that commented block?

uhurusurfa: Removed

uhurusurfa:
@Benbb96  or @DavidVadnais - review and if appropriate, merge at your convenience please.

Benbb96 (approved):
Apart from the hack, it looks good to me. Maybe we should create an issue to look deeper for a better way of handling these timezone aware frozen dates in tests.

PR #1267, opened 2025-05-14
Fix datatable default sort

uhurusurfa:
When the user clicks the "Apply Filters" button having selected a sorting column and sort order, the dataTable will render in the selected order. The ordering enacted by clicking on one of the column titles will then apply that columns ordering until the user clicks the "Apply Filters" button again at which point it will relaod the data using the selected ordering from the "Sorting" dropdown.


DavidVadnais (approved):
Great!

PR #1269, opened 2025-05-16
Fix markdown javascript injection prevention

### LINKED ISSUE #1102: get_markdown incorrectly renders text that contains both hyperlinks and parentheses
If the following sentence is entered into a ticket, it does not render correctly in the ticket.html template:

Here is some example text containing (parentheses) before a link also surrounded by parentheses (https://github.com/).

It renders as:

Here is some example text containing (//github.com/).

The missing text in bold:

Here is some example text containing (**parentheses) before a link also surrounded by parentheses (https:**//github.com/).

I've narrowed down the bug to the regex and text replacement in get_markdown(). The text renders correctly when the replacement code (lines 58-67) is commented out.
https://github.com/django-helpdesk/django-helpdesk/blob/67eb0974c7f163216ececc1d8a715a0144d0375c/helpdesk/models.py#L52-L77

uhurusurfa:
Fixes #1102 

There ended up being some fairly significant shrotcomings in the orioginal algorithm and have now added more detailed tests to reflect the capability.

### DIFF
--- a/helpdesk/models.py
+++ b/helpdesk/models.py
@@ -49,7 +49,9 @@ def get_markdown(text):
 
     # Search for markdown that creates a clickable link and remove the undesirable ones
     pattern = re.compile(r"(\[[\s\S]*?\])\(([\w]*?):([\s\S]*?)\)", flags=re.MULTILINE)
-    rerun_scheme_check = True  # Used to decide if a re-check should be done after last pass
+    rerun_scheme_check = (
+        True  # Used to decide if a re-check should be done after last pass
+    )
     while rerun_scheme_check:
         has_illegal_scheme = False
         for m in re.finditer(pattern, text):
--- a/helpdesk/models.py
+++ b/helpdesk/models.py
@@ -49,7 +49,7 @@ def get_markdown(text):
 
     # Search for markdown that creates a clickable link and remove the undesirable ones
     pattern = re.compile(r"(\[[\s\S]*?\])\(([\w]*?):([\s\S]*?)\)", flags=re.MULTILINE)
-    rerun_scheme_check = True # Used to decided to re-check the text after each parse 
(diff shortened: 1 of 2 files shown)


helpdesk/models.py
```diff
@@ -38,22 +38,29 @@ def extendMarkdown(self, md):
...
-
+    # Search for markdown that creates a clickable link and remove the undesirable ones
+    pattern = re.compile(r"(\[[\s\S]*?\])\(([\w]*?):([\s\S]*?)\)", flags=re.MULTILINE)
+    rerun_scheme_check = True  # Used to decided to re-check the text after each parse
+    while rerun_scheme_check:
+        has_illegal_scheme = False
+        for m in re.finditer(pattern, text):
+            # check if scheme is allowed
+            if m.group(2).lower() in helpdesk_settings.ALLOWED_URL_SCHEMES:
+                # Considered safe so dn't change it.
```
DavidVadnais, line 58: `dn't` --> `don't`

helpdesk/models.py
```diff
@@ -38,22 +38,29 @@ def extendMarkdown(self, md):
...
-        if scheme in helpdesk_settings.ALLOWED_URL_SCHEMES:
-            replacement = "\\1(\\2:\\3)"
-        else:
-            replacement = "\\1(\\3)"
-
-        text = re.sub(pattern, replacement, text, flags=re.IGNORECASE)
-
+    # Search for markdown that creates a clickable link and remove the undesirable ones
+    pattern = re.compile(r"(\[[\s\S]*?\])\(([\w]*?):([\s\S]*?)\)", flags=re.MULTILINE)
+    rerun_scheme_check = True  # Used to decided to re-check the text after each parse
```
DavidVadnais, line 52: `decided` --> `decide` or some better rewording

DavidVadnais (changes requested):
This makes sense to me.

PR #1270, opened 2025-05-28
Make title length in form consistent with title length in model.

kennedydane:


uhurusurfa:
> I think we need to change the ticket change model as well.
> 
>     * [class TicketChange line 1164](https://github.com/django-helpdesk/django-helpdesk/blob/main/helpdesk/models.py#L1164)

That is just the title of the field (eg Title, Description, status etc) and I cannot imagine anyone creating a title that is longer than 100 characters.

uhurusurfa (approved):
Thanks for the fix.

uhurusurfa (changes requested):
You will need to run this locally and commit the changes to get the tests working:
`make format`

uhurusurfa (approved):
👍  - thanks for the PR

PR #1272, opened 2025-05-30
CLEANUP breakup models.py 

### LINKED ISSUE #1271: Code Cleaning: Break up models.py
When reading through the models in [models.py](https://github.com/django-helpdesk/django-helpdesk/blob/main/helpdesk/models.py) it can be hard to find things. I want to break it up.

We create a models directory can have each model in its own file. 

Leave it how it is.

**Additional context**
Before doing this I just want to make sure no one prefers the mega-file solution.  Its not that important to me to do this.

DavidVadnais:
Fixes #1271


DavidVadnais:
I'm not sure this is worth it.

Benbb96 (approved):
I just comme across your PR, after writing my comment on #1271.

It seems you managed to get it right by keeping an __init__ file so it doesn't look like it's going to break something. Also, tests are passing so I think it's OK to merge it.

One drawback I just think of is that it'll be a bit harder to check git history on models (but still possible though)

PR #1274, opened 2025-06-07
feat: add saved queries list view page with run and delete support

bhargav1002:
This feature adds a new Manage Saved Queries page accessible at `helpdesk/saved-searches/`. The page provides users with a list view of their saved queries, enabling easy management of these queries.

**Features**
- Displays all saved queries created by the user or shared with them.
- Allows users to run any saved query directly from the list.
- Enables users to delete saved queries they no longer need.
- Intuitive UI for managing saved queries efficiently.


uhurusurfa (approved):
Seems to work perfectly for me - makes managing saved queries more intuitive as stated in the PR - noice work.

PR #1278, opened 2025-06-11
Fix the issue of the sidebar toggle not working on base.html

Leggendario12:
- Add <script src="{% static 'helpdesk/js/sb-admin.js' %}"></script> to the helpdesk base.html just before the {% block helpdesk_js %} to ensure the sidebar toggle logic runs for logged-in users.

> **Note:** Including `base_js.html` both in the `<head>` *and* at the bottom of the `<body>` causes jQuery to load twice, which clobbers any plugins (e.g. the Timeline script). Ensure `base_js.html` is only included once, ideally just before

### DIFF
--- a/helpdesk/templates/helpdesk/public_base.html
+++ b/helpdesk/templates/helpdesk/public_base.html
@@ -6,10 +6,8 @@
 <html lang="en">
 
 <head>
-
     {% include 'helpdesk/base-head.html' %}
     {% block helpdesk_head %}{% endblock %}
-    {% include 'helpdesk/base_js.html' %}
 
 </head>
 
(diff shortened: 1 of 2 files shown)


uhurusurfa (changes requested):
Thanks for the fix - can you go ahead and remove the duplicate as part of this PR please?

uhurusurfa:
@Leggendario12 - can you also increment the version number to 1.6.1 in setup.py for a relase pelase

Leggendario12:
@uhurusurfa Thank you for the feedback! I’ve removed the duplicate {% include 'helpdesk/base_js.html' %} from the <head> and bumped the version to 1.6.1. Please let me know if there’s anything else.

PR #1280, opened 2025-06-17
Fix ticket ID matching across multiple queues

rachitpoudel101:
The issue was that the indentation was incorrect in your code. The block for checking other queues wasn't properly indented under the if ticket_id is None: condition, causing it to be outside the conditional block. This meant the code wasn't executing as intended when processing emails.

With this fix, the code will now properly check other queues only when no ticket ID is found in the current queue, ensuring that emails are correctly associated with existing tickets in any queue before attempting to create new tickets

Screenshot from 2025-06-17 10-20-29

### DIFF
--- a/helpdesk/email.py
+++ b/helpdesk/email.py
@@ -1096,10 +1096,11 @@ def extract_email_metadata(
     )
 
     # If no ticket ID found in the current queue, check all other queues
+    original_queue = queue  # Store the original queue in case we need to revert
     if ticket_id is None:
-        # Get all enabled queues except the current one
+        # Get all enabled queues except the current one, regardless of allow_email_submission
         other_queues = Queue.objects.exclude(id=queue.id).filter(
-            email_box_type__isnull=False, allow_email_submission=True
+            email_box_type__isnull=False
         )
         
         for other_queue in other_queues:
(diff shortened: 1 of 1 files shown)


helpdesk/email.py
```diff
@@ -1090,9 +1090,31 @@ def extract_email_metadata(
...
+    # First try to get ticket ID from the current queue's slug
     ticket_id: typing.Optional[int] = get_ticket_id_from_subject_slug(
         queue.slug, subject, logger
     )
+
+    # If no ticket ID found in the current queue, check all other queues
+    if ticket_id is None:
+        # Get all enabled queues except the current one
+        other_queues = Queue.objects.exclude(id=queue.id).filter(
+            email_box_type__isnull=False, allow_email_submission=True
```
mdesco-crgl, line 1102: A queue may accept email replies from a mailbox but not submissions. 
```suggestion
            email_box_type__isnull=False
```

rachitpoudel101: The changes allow matching ticket IDs across all queues with email configured, regardless of their 'allow_email_submission' setting. This is critical because a queue may need to accept email replies to existing tickets even if it doesn't accept new ticket submissions via email.

By storing the original queue and reverting back to it when no existing ticket is found, we ensure new tickets are always created in the appropriate receiving queue while still allowing updates to existing tickets in any queue.

This addresses the common workflow where certain queues only accept updates to existing tickets but not new submissions.

uhurusurfa (commented):
If you have the time, would you mind adding a unit test for extracting the ticket Id from the subject?
If not let me know and I will approve this PR and add one when I have the time before releasing.

uhurusurfa:
Fixed in new PR #1284

PR #1282, opened 2025-06-19
Fix due_date being overwritten by ticket update shortcuts

### LINKED ISSUE #1281: Taking a ticket resets due date
**Describe the bug**
When a ticket is "taken" (assigned to shortcut), the update_ticket method only populates a few attributes from the template context. This causes the "Due date" to be reset to None.

**To Reproduce**
1. Go to a ticket with a due date
2. Click on the "Take" button
3. Scroll down to the resultant follow-up
4. See that due date was reset.

**Expected behavior**
All existing attributes should be preserved. Only the ticket owner should get updated.

Version: 1.3.0

**Additional context**

The route for the Take update is here: https://github.com/django-helpdesk/django-helpdesk/blob/main/helpdesk/views/staff.py#L406

The context for the ticket is pulled from here: https://github.com/django-helpdesk/django-helpdesk/blob/main/helpdesk/update_ticket.py#L207

Due date can only be `None` here as it it passed from the function params: https://github.com/django-helpdesk/django-helpdesk/blob/main/helpdesk/update_ticket.py#L312

mdesco-crgl:
resolves #1281

This implementation is safe because [updating the ticket from the form](https://github.com/django-helpdesk/django-helpdesk/blob/main/helpdesk/views/staff.py#L671) already cleans the due_date from the form. When due_date is None, it will be set to the existing ticket.due_date anyways.

### DIFF
--- a/helpdesk/update_ticket.py
+++ b/helpdesk/update_ticket.py
@@ -309,7 +309,7 @@ def update_ticket(
         )
         ticket.queue_id = queue
 
-    if due_date != ticket.due_date:
+    if due_date and due_date != ticket.due_date:
         f.ticketchange_set.create(
             field=_("Due on"),
             old_value=ticket.due_date,


uhurusurfa (approved):
Thanks for the fix 🚀

PR #1285, opened 2025-07-08
Add custom sorting in get_datatable_context

ezra-varady:
Because the columns are hard coded sorting the datatable on anything other than the hard coded columns doesn't work. This is a nominal patch implementing the feature I need. I'm not familiar enough with the surrounding code to tell if this is the correct call site to implement this feature so I'm making this a draft. I believe it maintains the API though.


DavidVadnais:
Hello,

You're not passing ruff.  Can you update this merge to pass lint?

uhurusurfa:
I will review when the PR is changed from WIP to ready for review. Fundamentally looks correct.

uhurusurfa (approved):
🚀 - thanks for the enhancment

Issue #1286, opened 2025-07-15
Cannot add comment to ticket with custom fields


tissieres:
**Describe the bug**
Following the PR #1256, we cannot add a comment to an opened ticket with custom fields. The request is POSTed to view `staff.update_ticket_view`, which redirects to CBV `UpdateTicketView`. This CBV uses `TicketForm` to validate the data submitted. The `form_valid` method is never triggered and adding a `form_invalid` method helped me to understand that the `TicketForm` expects all the custom fields and the `queue` fields while this update sends `comment` and other fields. 

**To Reproduce**
1. Click on an opened ticket with custom fields
2. Enter text in the 'comment' textarea at the bottom and submit the form.
3. Comment is not added and the ticket is displayed again with this comment.

**Expected behavior**
The comment and/or attachment is added to the ticket and the following expected actions are performed (e.g. sending an email to the customer). 

**Additional context**
Tested on version 1.5.1 and 1.6.1.

uhurusurfa:
The bug has actually existed since some of the earliest versions and relates specifically to custom fields that are marked as required.
Since the custom fields are hidden by default when the view renders, if someone adds a required custom field after the ticket has been created, they would have to click the "Change Further Details" button to see the error message and update the new required field.
To solve this problem ( PR #1249 ) , when the user clicks on the "Update This Ticket" button, javascript automaitcally clicks the "Change Further Details" button so that if an error is found in the "hidden" fields it will be visible to the user but if no errors are found then the screen will simply update as expected.

tissieres:
I tested your PR #1287 on my project, but the form is still invalid in the `UpdateTicketView` even if all custom fields were already filled in for this ticket. 

The fields submitted when I add a comment are: `comment`, `new_status`, `public`, `time_spent`, `title`, `owner`, `priority`, `due_date`, `attachment` and the CSRF token.

The form errors show that all custom fields set as mandatory as well as `queue` are still missing.

uhurusurfa:
You will have to provide more infirmation on your setup as I cannot replicate your issue.

This is the POST from the browser and you can see all the fields including "Queue" in green and custom fields "Foo Bar" (custom_foo_bar) and "Boo Baa" (custom_boo) highlighted in red along with all the other expected fields being submitted in the POST:

<img width="1304" height="741" alt="Image" src="https://github.com/user-attachments/assets/3e863379-c0c7-49f3-9a1b-b20bf03573bd" />

[...]

PR #1287, opened 2025-07-16
Fix ticket form update when there are custom required fields

uhurusurfa:
The ticket form hides some extended fields by default including all the custom fields.
If a custom field that is set as required is rendered then the field is not visible to the user by default and updating the tickets fields that are visible by default fails because there is a javascript error complaining about the lack of visibility of the required field it is trying to add an error message to which is not shown in the UI and the UI is unaware that there is anything wrong.

This fix forces the hidden fields to be displayed before the form validation runs so that the error is visible to the user.

The PR also adds a unit test that specifically deals with required fields but does not test the soultion since there are no UI tests in this app.

Additionally a minor enhancemnet is added to the make script to allow the rundemo script in the makefile to be run repeatedly without issues surroundiung the admin user.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -91,6 +91,8 @@ readme:
 	$(TOX) -e readme
 
 
+PY ?= from django.contrib.auth import get_user_model;print(1 if get_user_model().objects.filter(username="admin").exists() else 0)
+NO_USER ?= $(shell demodesk shell -c '${PY}' --no-imports)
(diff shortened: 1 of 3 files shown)


uhurusurfa:
Fixes #1286

DavidVadnais (approved):
This all looks sane to me

PR #1288, opened 2025-07-21
Turkish translation

X-croot:
This pull request updates the Turkish translation (locale/tr/LC_MESSAGES/django.po and .mo).
Translations were checked and compiled with msgfmt to ensure correctness.


uhurusurfa (approved):
Thanks for the contribution

Issue #516, opened 2017-06-06
default assigned_to


mrkiwi-nz:
## Feature Request
As far as i can tell, there are no settings to handle this situation;
* A mailbox is set up to receive bounced email messages from automated systems
* These bounces should **never** be replied to - the bounce is there because the original sender is not contactable and the reason is somewhere in the message (ie over quota, bad MX, no mailbox at that address, temporary routing failure)
* A queue in django-helpdesk is created and configured to watch this mailbox - the tickets created by the emails **should** be automatically assigned to one person and an email sent to them, and as above, should **not** generate the email acknowledging the creation of the ticket

We have the same situation for queues that monitor application errors, printer faults, system alerts etc etc - all have the same attributes in that the owner/submitter of the ticket is not a real person and should not be replied to.

This feature request can be split into two parts - a flag set on the queue that specifies that the owner should not be contacted, and a system to automatically assign a ticket to a person, but only when that ticket is generated by an email queue. AFAIK this 2nd feature is **not** the same as default_owner - default_owner only assigns an owner when an anonymous user creates a ticket via the web page.

Any comments?

gwasser:
I like the proposal offhand, I have a similar problem in my usage where we prefer to not reply to the original submitter.

I'll have to take a look at `default_owner`, I'm not sure if that's only anonymous user via public or not. I know.

Probably I see the setting being more of a drop-down list of options rather than boolean on/off, of "Always assign to default_owner", "Only assign to default_owner when receive email report", or "Never assign to default_owner (always Unassigned)".

mrkiwi-nz:
I find that ownership (meaning the creator of the ticket) is somewhat muddied by the verb "Take" on the button next to "assigned to" - I read it as "Take Ownership" when it doesn't mean that at all - it means "Take this ticket" or "Assign to me". Also, internally we have the mantra of owning a problem until it is solved, so by "Take"ing a ticket, you own it until you assign it to someone else, or resolve it.
Do we have a clear nomenclature here?
Maybe we could improve it by more clearly defining the terms.

gwasser:
We could change the text on the button to read "Assign to Me" if that's more clear. What other terms do you think might be unclear?

PR #1293, opened 2025-08-01
Added option for ticket reassign

DavidVadnais:
My users want to be able to take the ticket even after someone else has taken it. I thought about just removing this if but this will allow it to be used as it was before and in the new way. 

image

The desire is to have the hand:

image

In tickets with an owner. What do you think we should do here? I almost think there there should be a drop down for team member assignment instead of the hand.

### DIFF
--- a/helpdesk/views/staff.py
+++ b/helpdesk/views/staff.py
@@ -485,7 +485,7 @@ def view_ticket(request, ticket_id):
             "checklist_form": checklist_form,
             "customfields_form": customfields_form,
             "assignable_users": get_assignable_users(
-                settings.HELPDESK_STAFF_ONLY_TICKET_OWNERS
+                bool(getattr(settings, "HELPDESK_STAFF_ONLY_TICKET_OWNERS", False))
             ),
             **extra_context_kwargs,
         },


DavidVadnais:
Okay now if I was assigned it would be
image

And if no one assigned would be:

image

Also as far as ticket #516 the hover would in my case say `Assign this ticket to David Vadnais` so I think that part of the ticket is done. However this doesnt touch the defaulting of ownership

uhurusurfa (dismissed):
Happy with what you have done so far.

helpdesk/templates/helpdesk/ticket_desc_table.html
```diff
@@ -73,14 +73,37 @@ <h3>{{ ticket.queue.slug }}-{{ ticket.id }}. {{ ticket.title }}</h3>
...
                         <th class="table-active">{% trans "Assigned To" %}</th>
                         <td>
-                            {{ ticket.get_assigned_to }}
-                            {% if _('Unassigned') == ticket.get_assigned_to  %}
-                                <a class="btn btn-primary btn-sm float-right" data-toggle="tooltip" href='?take' title='{% trans "Assign this ticket to " %}{{ request.user.email }}'>
-                                    <i class="fas fa-hand-paper"></i>
-                                </a>
-                            {% endif %}
+                            <!-- assignment drop down -->
+                            <form method="post" action="{% url 'helpdesk:mass_update' %}">
```
Benbb96, line 80: Using the mass update endpoint is a good idea since we can reuse the logic, but it's important to note that the followup indicating the assignee change will be worded like "Assigned to %(username)s in bulk update".

helpdesk/templates/helpdesk/ticket_desc_table.html
```diff
@@ -73,14 +73,37 @@ <h3>{{ ticket.queue.slug }}-{{ ticket.id }}. {{ ticket.title }}</h3>
...
+                                        {% for user in assignable_users %}
+                                            <option value="assign_{{ user.id }}" {% if ticket.assigned_to and ticket.assigned_to.id == user.id %}selected{% endif %}>
+                                                {{ user.get_full_name|default:user.username }}
+                                            </option>
+                                        {% endfor %}
+                                    </select>
+
+                                    <button type="submit" class="btn btn-primary btn-sm" title="Save ticket assignment">
+                                        <i class="fas fa-user-check"></i>
+                                    </button>
```
Benbb96, line 97: I think it could be better looking by integrating the submit button inside the select field using Bootstrap's input group : https://getbootstrap.com/docs/5.0/forms/input-group/#button-addons

helpdesk/views/staff.py
```diff
@@ -473,6 +477,10 @@ def view_ticket(request, ticket_id):
     # add custom fields to further details panel
     customfields_form = EditTicketCustomFieldForm(None, instance=ticket)
 
+    # Define users that the ticket can be assigned to
+    assignable_users = User.objects.filter(is_active=True).order_by("username")
```
Benbb96, line 481: Shouldn't we also filter by staff member?

helpdesk/views/staff.py
```diff
@@ -473,6 +477,10 @@ def view_ticket(request, ticket_id):
...
+    logger.debug("Assignable users:", assignable_users)
```
Benbb96, line 482: Not sure if we want to keep the debug?

helpdesk/templates/helpdesk/ticket_desc_table.html
```diff
@@ -73,14 +73,37 @@ <h3>{{ ticket.queue.slug }}-{{ ticket.id }}. {{ ticket.title }}</h3>
...
-                            {% endif %}
+                            <!-- assignment drop down -->
+                            <form method="post" action="{% url 'helpdesk:mass_update' %}">
+                                {% csrf_token %}
+                                <input type="hidden" name="ticket_id" value="{{ ticket.id }}">
+                                <input type="hidden" name="next" value="{% url 'helpdesk:view' ticket.id %}">
+
+                                <div class="d-flex align-items-center gap-2">
+                                    <select name="action" class="form-select form-select-sm">
+                                        <option disabled selected>Unassigned</option>
```
Benbb96, line 87:
```suggestion
                                        <option value="unassign"{% if not ticket.assigned_to %} disabled selected{% endif %}>{% trans "Unassigned" %}</option>
```

helpdesk/templates/helpdesk/ticket_desc_table.html
```diff
@@ -73,14 +73,37 @@ <h3>{{ ticket.queue.slug }}-{{ ticket.id }}. {{ ticket.title }}</h3>
...
+                                        {% for user in assignable_users %}
+                                            <option value="assign_{{ user.id }}" {% if ticket.assigned_to and ticket.assigned_to.id == user.id %}selected{% endif %}>
+                                                {{ user.get_full_name|default:user.username }}
+                                            </option>
+                                        {% endfor %}
+                                    </select>
+
+                                    <button type="submit" class="btn btn-primary btn-sm" title="Save ticket assignment">
```
Benbb96, line 95:
```suggestion
                                    <button type="submit" class="btn btn-primary btn-sm" data-toggle="tooltip" title="{% trans "Save ticket assignment" %}">
```

Benbb96 (changes requested):

helpdesk/templates/helpdesk/ticket_desc_table.html
Benbb96: I also noticed that the mass update won't create a TicketChange for assign/unassign like in the update ticket function. I think it's worth fixing it either in this MR or in a new one.

helpdesk/views/staff.py
DavidVadnais: removed

helpdesk/templates/helpdesk/ticket_desc_table.html
DavidVadnais: I believe I have done this

helpdesk/views/staff.py
```diff
@@ -488,6 +497,7 @@ def view_ticket(request, ticket_id):
             "SHOW_SUBSCRIBE": show_subscribe,
             "checklist_form": checklist_form,
             "customfields_form": customfields_form,
+            "assignable_users": assignable_users,
```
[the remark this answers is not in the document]
uhurusurfa: The get_active_users() seems like a really unhelpful implementation to have it controlled by a sertting to be active users or active users that are staff members.
Currently it is only used in the staff.py module and really should be set up with a method parameter to add filtering by staff members since the method name implies that it will get all actie users and if it becomes more broadly used then it may be difficult to undo down the road.
The callee should be passing a parameter requesting to filter additionally by staff members and then the method can be more generically used.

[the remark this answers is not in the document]
Benbb96: Maybe renaming the function by something like `get_assignable_users` would be more precise. And thus, we could keep it controlled by the setting.

Or, I can add a parameter `filter_staff` which would take the setting's value as the default. But for now, I don't see why we would use that function for something else than getting the list of assignable users to display in a dropdown.

DavidVadnais:
Now thanks to @Benbb96 comments:
image

image

helpdesk/views/staff.py
```diff
@@ -473,6 +477,11 @@ def view_ticket(request, ticket_id):
     # add custom fields to further details panel
     customfields_form = EditTicketCustomFieldForm(None, instance=ticket)
 
+    # Define users that the ticket can be assigned to
+    assignable_users = User.objects.filter(is_active=True, is_staff=True).order_by(
+        "username"
+    )
+
```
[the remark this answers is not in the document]
DavidVadnais: Done.

DavidVadnais:
I believe I have made all the requested changes. Let me know if anything else is wrong or if I misunderstood any feedback.

Benbb96 (changes requested):
The screenshot is looking great. Just some minor changes and it'll be good!

helpdesk/views/staff.py
```diff
@@ -483,6 +490,7 @@ def view_ticket(request, ticket_id):
             "SHOW_SUBSCRIBE": show_subscribe,
             "checklist_form": checklist_form,
             "customfields_form": customfields_form,
+            "assignable_users": assignable_users,
```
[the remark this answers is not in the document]
DavidVadnais: Done.

uhurusurfa (approved):
LGTM

helpdesk/views/staff.py
```diff
@@ -483,6 +484,9 @@ def view_ticket(request, ticket_id):
             "SHOW_SUBSCRIBE": show_subscribe,
             "checklist_form": checklist_form,
             "customfields_form": customfields_form,
+            "assignable_users": get_assignable_users(
+                settings.HELPDESK_STAFF_ONLY_TICKET_OWNERS
```
uhurusurfa: This needs to be accessed via `helpdesk_settings` not `settings`.

uhurusurfa (changes requested):
Needs change of module to access HELPDESK_STAFF_ONLY_TICKET_OWNERS

helpdesk/views/staff.py
```diff
@@ -99,6 +100,9 @@
 if helpdesk_settings.HELPDESK_KB_ENABLED:
     from helpdesk.models import KBItem
 
+logger = logging.getLogger(__name__)
```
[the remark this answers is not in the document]
DavidVadnais: done

helpdesk/views/staff.py
DavidVadnais: Thank you I was going crazy

DavidVadnais:
I did a different work around Let me open another PR

PR #1295, opened 2025-08-05
Fix unassign action in update ticket + refactor the get_active_users

Benbb96:
There was a bug when changing ticket owner to unassigned. It was transforming id 0 to -1 when calling update_ticket function.

I think it could be great to remove the use of magic values like -1 since it is prone to bugs like this.

I also refactored get_active_users() into lib.py file in order to reuse it in forms.py.

### DIFF
--- a/helpdesk/templates/helpdesk/ticket.html
+++ b/helpdesk/templates/helpdesk/ticket.html
@@ -161,7 +161,7 @@
                 <dt><label for='id_owner'>{% trans "Owner" %}</label></dt>
                 <dd>
                     <select id='id_owner' name='owner'>
-                        <option value='0'>{% trans "Unassign" %}</option>
+                        <option value='0'>{% trans "Unassigned" %}</option>
                         {% for u in assignable_users %}
                             <option value='{{ u.id }}'{% if u.id == ticket.assigned_to.id %} selected{% endif %}>{{ u }}</option>
                         {% endfor %}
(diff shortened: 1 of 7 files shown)


helpdesk/templates/helpdesk/ticket.html
```diff
@@ -159,7 +159,14 @@ <h5 class="mb-1">{{ followup.title|escape|num_to_link }}</h5>
                 <dd><input type='text' name='title' value='{{ ticket.title|escape }}' /></dd>
 
                 <dt><label for='id_owner'>{% trans "Owner" %}</label></dt>
-                <dd><select id='id_owner' name='owner'><option value='0'>{% trans "Unassign" %}</option>{% for u in active_users %}{% if u.id == ticket.assigned_to.id %}<option value='{{ u.id }}' selected>{{ u }}</option>{% else %}<option value='{{ u.id }}'>{{ u }}</option>{% endif %}{% endfor %}</select></dd>
+                <dd>
+                    <select id='id_owner' name='owner'>
+                        <option value='0'>{% trans "Unassign" %}</option>
```
uhurusurfa, line 164: The use of the word "Unassign" indicates an action wheras "Unassigned" indicates a state so maybe as part of this PR rename it to "Unassigned".

This would make it aligned wit hthe nomenclature on the owner filter:
[https://github.com/django-helpdesk/django-helpdesk/blob/main/helpdesk/templates/helpdesk/filters/owner.html#L12](https://github.com/django-helpdesk/django-helpdesk/blob/main/helpdesk/templates/helpdesk/filters/owner.html#L12)

uhurusurfa (changes requested):
Apart from the proposed change of  "Unassign" to "Unassigned" thie refactoring is great and have tested it working on my local via demodesk.

PR #1301, opened 2025-08-19
Add HELPDESK_PRIVATE_FOLLOWUP_MEANS_NO_EMAILS setting

kennedydane:
It feels like cheating and like I might get sent to jail, but here we go.
__________

Add new setting to suppress all email notifications for private follow-ups, providing complete privacy for internal staff communications.

Features:
- Defaults to False (backward compatible)
- When True, blocks ALL emails for private follow-ups (public=False):
  - Submitters, assigned users, CC'd users, queue notifications
- Public follow-ups (public=True) continue working normally
- Overrides other notification settings like NOTIFY_SUBMITTER_FOR_ALL_TICKET_CHANGES

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/helpdesk/settings.py
+++ b/helpdesk/settings.py
@@ -357,6 +357,11 @@ HELPDESK_NOTIFY_SUBMITTER_FOR_ALL_TICKET_CHANGES = getattr(
     False,
 )
 
+# If True, private follow-ups (public=False) will not trigger any email notifications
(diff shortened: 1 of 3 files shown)


Benbb96 (commented):
It looks good to me. Maybe it is worth adding something about this new setting in the documentation too: https://github.com/django-helpdesk/django-helpdesk/blob/main/docs/settings.rst

PR #1305, opened 2025-08-22
BUG: Dockerfile Use explicit trailing slashes to copy directories correctly

DavidVadnais:
Now we cant find `/opt/django-helpdesk/standalone/requirements.txt` in Dockerfile step  `RUN uv pip install -r /opt/django-helpdesk/standalone/requirements.txt`

### DIFF
--- a/standalone/Dockerfile
+++ b/standalone/Dockerfile
@@ -8,7 +8,14 @@ RUN apt-get install -yqq \
    cron \
    git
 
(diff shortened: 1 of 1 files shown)


Benbb96 (approved):
Seems good to me but maybe we need to harmonize standalone as well, making it use uv and pyproject.toml ?

PR #1306, opened 2025-08-25
BUG: Standalone docker build works now

DavidVadnais:
Okay I finally pulled it down and ran `docker build --file standalone/Dockerfile -t djangohelpdesk/standalone:latest ..` until it worked.


uhurusurfa (approved):
thanks for sorting - I am actually gpoing to completely replace the entire "standalone" folder with something more functional and integerated into the Helpdesk eco system. instead of being a custom deployment that it is now for someones specific use case..

Issue #1307, opened 2025-08-30
Unable to use with unfold admin


jbaxleyiii:
**Describe the bug**
When using this with django-unfold the project cannot startup or run migrations because of a `Models aren't loaded yet` exception being thrown. From what I can tell this is because of unfold calling `.render` in an admin display which loads template tags and helpdesk imports settings from a forms file where models are used at the top level [here](https://github.com/django-helpdesk/django-helpdesk/blob/ed2f7aef30ef33f4fef198f30ea94f9bb9cb9a27/src/helpdesk/templatetags/helpdesk_util.py#L5-L9)

Changing this:
```python
from helpdesk.forms import (
    CUSTOMFIELD_DATE_FORMAT,
    CUSTOMFIELD_DATETIME_FORMAT,
    CUSTOMFIELD_TIME_FORMAT,
)
```

to this:
```python
from helpdesk.settings import (
    CUSTOMFIELD_DATE_FORMAT,
    CUSTOMFIELD_DATETIME_FORMAT,
    CUSTOMFIELD_TIME_FORMAT,
)
```

allows the project to be run as expected

uhurusurfa:
Try moving the import inside the function to see if indeed that is the cause.

jbaxleyiii:
@uhurusurfa thanks for the response! PR open :) https://github.com/django-helpdesk/django-helpdesk/pull/1308

PR #1308, opened 2025-08-30
Update helpdesk_util.py

jbaxleyiii:
fixes https://github.com/django-helpdesk/django-helpdesk/issues/1307


uhurusurfa (approved):
🚀 Thanks for the fix

PR #1309, opened 2025-08-30
Make the "Copies To" field in the ticket edit screen readonly when not email notifications are not enabled in Queue

### LINKED ISSUE #914: "Copies To" field is unobviously ignore for queues without "Notify contacts when email updates arrive" set
This behaviour is confusing because people still adding CC users and may be surprised why emails don't arrive to them.

A proposed solution is just to hide that field if the queue has it turned off anyway. I don't think it's worth adding much checks around it and just don't render it will be fine.

Another alternative is to make it readonly with a popover describing why it can't be changed.

Please say your thoughts about that; the functionality is pretty simple to implement.

uhurusurfa:
Checks the Queue.enable_notifications_on_email_events field setting and shows a disabled if the field is not checked
Fixes #914


Benbb96 (approved):
LGTM

PR #1314, opened 2025-09-14
API endpoints are now optional

### LINKED ISSUE #1290: Use only the API  without adding all routes to a project
In a website designed to have a frontend using a JavaScript framework, like React, the backend is just a set of RESTful APIs. Therefore, there is no need to have all routes added to the backend, only the ones related to the APIs.

A way to do it is to make the API URLs separated from other URLs, so one could do this:

```python
url_patterns = [
    path('/api/helpdesk', include('helpdesk.api.urls')
]
```

Therefore, only the API-related URLs would be included into the project, without exposing any other URL. For projects that requires all URLs, the way it is today, nothing changes.

As alternative, one can replicate the code in `urls.py` by including the view sets and set up only the required views. Although it is possible and works, it is kind of extra work, prone to error.

ronflima:
Some projects doesn't need an API for helpdesk enabled. So, I have added a new setting:

`HELPDESK_API_ENABLED` which defaults to True. So a project not willing the API to be exposed, should add this to its `settings.py` (example):

```python
HELPDESK_API_ENABLED = False
```

This resolves #1290 .


uhurusurfa:
My understandoing of #1290 is that @raidsnaldo who posted that issue is only using the API in  Helpdesk and is not interested in the other URLs provided for the default UI for Django Helpdesk.
This also makes sense to me.
I think to make it flexible,  the URL's need tp be split into 3 logical categories:

1. User management URL's - this is really the authentication and security related URL's (login, change password etc)
2. Helpdesk URL's for default UI provided for creating and managing tickets and related records
3. API urls providing creating and managing tickets and related records via a RESTful API

Group 1 URL's are always enabled.
Group 2 and 3 URL's can be independently switched on and off
So this PR so far supports switching group 3 on or off but leaves group 2 URL's enabl;ed even if they are not needed.

To take this a bit further, to support extending the Django Helpdesk app with custom API's, and avoid having to overwrite the default urls.py it would be useful to have an optional include of a file that could contain custom URL's using the mechanism described by @raidsnaldo with a settings switch:

```
if helpdesk_settings.HELPDESK_CUSTOM_URLS_ENABLED:
  url_patterns = [
      path('/api/helpdesk', include('helpdesk.custom.urls')
  ]
```

DavidVadnais:
@ronflima Just checking in, are you still planning to continue this PR or should we close it for now? No rush at all.

ronflima:
> @ronflima Just checking in, are you still planning to continue this PR or should we close it for now? No rush at all.

Hi, David.  I'm in debt with this project. I'm picking it up to complete it.

ronflima:
As per @uhurusurfa request, I'm adding a new setting, `HELPDESK_UI_ENABLED`, which defaults to `True` and can disable all UI routes, but the ones he classified as "Group 1": 

> Helpdesk URL's for default UI provided for creating and managing tickets and related records

About the `helpdesk.custom.urls` I'll need to think it over. It is an interesting idea. I'll figure out something for it as well.

DavidVadnais:
Is this still draft and can you run the `make checkformat`?  

Thank you!

ronflima:
> Is this still draft and can you run the `make checkformat`?
> 
> Thank you!

Calm down, dear friend! I just had this weekend to work on it. Although very, very trivial and simple, I needed to seat on my computer to do so, right? After this comment this PR will be ready for review. 

@uhurusurfa a question for you. You suggested the following:

```python
if helpdesk_settings.HELPDESK_CUSTOM_URLS_ENABLED:
  url_patterns = [
      path('/api/helpdesk', include('helpdesk.custom.urls')
  ]
```

I not sure what is the utility of this. I'll go on and deliver this PR without it for the time being, because I need to understand better what was your idea behind this - the "why", which is not clear to me. It feels to me that a help desk user can add routes prefixed with `/api/helpdesk` outside the application, making this questionable.

uhurusurfa:
You can just ignore this piece:
```
if helpdesk_settings.HELPDESK_CUSTOM_URLS_ENABLED:
  url_patterns = [
      path('/api/helpdesk', include('helpdesk.custom.urls')
  ]
```

We can implement it in a separate PR if there is a user requirement needing it.

The PR as it stands seems to implement the expected behaviour.

@DavidVadnais  - do you see any issues wit hthe implementation?

DavidVadnais (approved):
This looks good to me

PR #1316, opened 2025-09-16
Fix crash in UpdateTicketView when unauthenticated user logs in after redirect

### LINKED ISSUE #1315: An error after logging in to update a ticket from an unauthenticated session
**Describe the bug**
When an unauthenticated user attempts to add a "Comment / Resolution" to a ticket, they are correctly redirected to the login page. However, upon successful login, instead of being redirected to the ticket page, they are taken to a 500 Internal Server Error page.

The error occurs because the **UpdateTicketView** attempts to access form data **(kwargs["form"].data)** during the GET request that follows the login redirection, but this data is not present.

**To Reproduce**
1. Go to Homepage without logging in
2. Create a new ticket
3. Try to add a comment / resolution 
4. User will be redirected to login page
5. Log in
6. The error page will be displayed

**Expected behavior**
The Ticket Update page should be presented for the log in page, if possible with the updates that were submitted before signing in

**Screenshots/Video**
I have a screencast demonstrating the issue, which is available here:
[https://drive.google.com/file/d/1Jdt0aqajCnWdT5ACZfVJ7pyvZu_elFsT/view?usp=sharing](url)

Please let me know what do you think about this, I'd be happy to create a PR for it as well

Arbaji:
### Problem
When an unauthenticated user tries to add a comment/resolution, they are redirected to login. After login, the redirect caused a 500 error because UpdateTicketView accessed form.data on a GET request.

### Solution
Added a safe check in get_context_data to only access form.data if it exists. Otherwise, provide an empty dict.

### Tests
Added regression test to confirm:
- Unauthenticated users are redirected to login.
- After login, the update page loads successfully (status 200).

### Related Issue
Closes #1315

### Future Work
It might be possible to preserve the entered comment/attachments across login using session storage or temporary persistence. This PR does not address that, but I’d be happy to explore it in a follow-up if maintainers are interested.

### DIFF
--- a/src/helpdesk/views/staff.py
+++ b/src/helpdesk/views/staff.py
@@ -1322,15 +1322,19 @@ class UpdateTicketView(
         extra = get_form_extra_kwargs(self.request.user)
         kwargs.update(extra)
         # Copy all data submitted that is not in the forms defined fields
-        form_fields = kwargs["form"].base_fields
-        all_fields = kwargs["form"].data
-        self.extra_context = {
-            "xform": {
-                k: v
-                for k, v in all_fields.items()
-                if k != "csrfmiddlewaretoken" and k not in form_fields
+        form = kwargs.get("form")
+        if form is not None and hasattr(form, "data") and form.data:
+            form_fields = form.base_fields
+            all_fields = form.data
+            self.extra_context = {
+                "xform": {
+                    k: v
+                    for k, v in all_fields.items()
+                    if k != "csrfmiddlewaretoken" and k not in form_fields
+                }
             }
-        }
+        else:
+            self.extra_context = {"xform": {}}
         return super().get_context_data(**kwargs)
 
     def get_form_kwargs(self):
--- a/tests/test_update_ticket_view.py
+++ b/tests/test_update_ticket_view.py
@@ -1,0 +1,37 @@
+from django.test import TestCase
+from django.urls import reverse
+from django.contrib.auth import get_user_model
+from helpdesk.models import Queue, Ticket
+
+User = get_user_model()
+
+
+class UpdateTicketViewTests(TestCase):
+    def setUp(self):
+        self.queue = Queue.objects.create(title="Test Queue", slug="test")
+        self.user = User.objects.create_user(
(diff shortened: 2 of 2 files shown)


src/helpdesk/views/staff.py
```diff
@@ -1322,15 +1322,19 @@ def get_context_data(self, **kwargs):
...
-        form_fields = kwargs["form"].base_fields
-        all_fields = kwargs["form"].data
-        self.extra_context = {
-            "xform": {
-                k: v
-                for k, v in all_fields.items()
-                if k != "csrfmiddlewaretoken" and k not in form_fields
+        form = kwargs.get("form")
+        if form is not None and hasattr(form, "data") and form.data:
+            form_fields = form.base_fields
```
DavidVadnais, line 1327: It may be worth check if this works better with form.fields 

If the form has dynamic fields (e.g. modified in __init__), form.base_fields might not reflect the runtime reality but I'm not entirely sure.

src/helpdesk/views/staff.py
```diff
@@ -1322,15 +1322,19 @@ def get_context_data(self, **kwargs):
...
+            self.extra_context = {
+                "xform": {
+                    k: v
+                    for k, v in all_fields.items()
+                    if k != "csrfmiddlewaretoken" and k not in form_fields
+                }
             }
-        }
+        else:
+            self.extra_context = {"xform": {}}
```
DavidVadnais, line 1337: Might be worth trying :

```suggestion
            self.extra_context = {"xform": xform}
```

src/helpdesk/views/staff.py
Arbaji: Thanks for the suggestion! I checked and confirmed that `xform` is only used in the template to repopulate `comment` and `time_spent`. So even if `base_fields` doesn’t cover dynamically added fields, it wouldn’t cause issues in practice.
That said, I’m fine with switching to `fields` here for clarity and future-proofing. Also, I noticed that custom field display has a separate bug, I’ll open a dedicated issue/PR for that.

PR #1319, opened 2025-09-20
Fix missing custom fields in UpdateTicketView after validation errors

### LINKED ISSUE #1318: Custom fields disappear from Ticket Update form after validation errors
**Describe the bug**
When a ticket update form includes custom fields, they display correctly on the first load.
However, if the form is submitted with invalid input, the page reloads with validation errors but the custom fields no longer appear.

**To Reproduce**
1. Enable custom fields for tickets.
2. Navigate to a ticket update form (the custom fields are visible).
3. Enter invalid data in a required field and submit.
4. After the page reloads with errors, the custom fields are missing.

**Expected behavior**
The ticket update form should display all custom fields consistently, both on first load and after validation errors.

**Screenshots**
 I have a screencast demonstrating the issue, which is available here:
https://drive.google.com/file/d/1sr1KD-FiIzq3fLmm7u-MSpoALkSl6NVP/view?usp=sharing

**Additional context**
It seems the form is not reinjecting custom fields on POST re-render, even though they are added in `TicketForm.__init__`.
I’ll be working on a PR to fix this issue.

Arbaji:
### Problem
When a ticket has custom fields, they display correctly on the first load (`/tickets/<id>/` via `view_ticket`).  
However, if invalid data is submitted on the update form, the page reloads with errors but the custom fields disappear.  
This happens because `UpdateTicketView` did not provide `customfields_form` to the template, unlike `view_ticket`.

### Solution
- Updated `UpdateTicketView.get_context_data` to always pass `customfields_form` (using `EditTicketCustomFieldForm`).  
- This ensures that custom fields remain visible both on initial GET and after POST with validation errors.

### Tests
Added a regression test:
  - Creates a custom field.
  - Confirms it is visible on the initial view (`helpdesk:view`).
  - Submits invalid data to `helpdesk:update`.
  - Confirms the page shows validation errors *and* the custom field is still visible.

### Related Issue
Closes #1318

### DIFF
--- a/tests/test_update_ticket_view.py
+++ b/tests/test_update_ticket_view.py
@@ -1,7 +1,7 @@
 from django.test import TestCase
 from django.urls import reverse
 from django.contrib.auth import get_user_model
-from helpdesk.models import Queue, Ticket
+from helpdesk.models import Queue, Ticket, CustomField
 
 User = get_user_model()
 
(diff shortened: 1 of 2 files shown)


tests/test_update_ticket_view.py
```diff
@@ -35,3 +35,48 @@ def test_redirect_after_login_does_not_crash(self):
...
+            {
+                "queue": self.queue.id,
+                # "title" missing
+                "body": "Test body",
+                "priority": 3,
+            },
+        )
+        self.assertEqual(response.status_code, 200)
+        self.assertContains(response, "There are errors in the form")
+        self.assertContains(response, "Extra Info")
```
Benbb96, line 82: You could also check that submitted values for custom fields are also still shown on the page

Benbb96 (dismissed):
Apart from the test that could be enhanced, it looks good to me!

tests/test_update_ticket_view.py
Arbaji: > You could also check that submitted values for custom fields are also still shown on the page
Test is updated

Benbb96: Perfect, thank you :)

Issue #1294, opened 2025-08-02
Update bootstrap, font awesome, jquery


DavidVadnais:
# Bootstrap
We use bootstrap 4.6 we should use the latest https://endoflife.date/bootstrap

This was tried before in [Release 0.3.3](https://github.com/django-helpdesk/django-helpdesk/releases/tag/0.3.3) commit [4df831d](https://github.com/django-helpdesk/django-helpdesk/commit/4df831d) and reverted in commit [ad519a2](https://github.com/django-helpdesk/django-helpdesk/commit/ad519a2)

The static files can be found here: https://github.com/django-helpdesk/django-helpdesk/tree/main/helpdesk/static/helpdesk/vendor/bootstrap.

The html tags will likely need significant change to make this work.  I'll do this soon unless someone else wants to do this 

# Font awsome
https://endoflife.date/font-awesome

Last update [e0c0754](https://github.com/django-helpdesk/django-helpdesk/commit/e0c0754)

# jquery
https://endoflife.date/jquery

last update [6ea9009dda7a94b883eb5a4d2bfaaaf9ec8cdd4a](https://github.com/django-helpdesk/django-helpdesk/commit/6ea9009dda7a94b883eb5a4d2bfaaaf9ec8cdd4a)

# Reference
#945 #598

uhurusurfa:
Definitely need to update the libraries used in this project.
There is a utility that might help in the process:

[https://github.com/coliff/bootstrap-5-migrate-tool/ ](https://github.com/coliff/bootstrap-5-migrate-tool/ )- this does a string searhc and replace so would probably handle Django templates

I will upgrade the project packaging stuff in the coming months as setup.py is deprecated:
https://packaging.python.org/en/latest/discussions/setup-py-deprecated/

DavidVadnais:
Thank you for the tool. Ill try and work this once #1293 is done.

uhurusurfa:
It would be worth implementing yarn or similar package manager to manage the library dependencies and make it easier to update to newer versions.

DavidVadnais:
Do you prefer yarn to npm?

PR #1324, opened 2025-10-14
first pass at switching to yarn

DavidVadnais:
# Yarn
Relates to  #1294 

I'm getting a sanity check before going forward as I am new to yarn. Ive tested the app and this approach does seem to render correctly. 

I'm testing with :
```
# step 1 - download dependencies
yarn install

# step 2 - Move dependencies
yarn build:static
# or
make static-vendor

# step 3 - Run project
make rundemo
```

## Still todo in this pr:
- [x] Add to CI/CD -- all changes handled through the dockerfile
- [x] make it so the make commands are aware of the need to run `yarn install`
- [x] fix demodesk

# still todo in a separate pr 
* update the libraries. in #1294

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "django-helpdesk"
-version = "2.0.2"
+version = "2.1.0"
 description = "Django-powered ticket tracker for your helpdesk"
 authors = [
     {name = "Ross Poulton", email = "ross@rossp.org"},


DavidVadnais:
From my initial testing only all tickets has significant problems

edit: 
this is true when accessed through docker
```
docker build --file standalone/Dockerfile --no-cache -t djangohelpdesk/standalone:latest ..
cd standalone
docker compose up
```

but not through `make rundemo`

DavidVadnais:
Okay we can work the docker problem in another ticket and do a release once that is fixed.

DavidVadnais:
Okay it now work consistently with `docker` and `make rundemo`

uhurusurfa:
We will need to automate updating the CDN versions in the templates but that can be for another PR.

PR #1327, opened 2025-10-29
Add requirements for jq, yarn, and npm

DavidVadnais:
Added requirements section to README for necessary tools.

I'm not sure this is the way we want to handle these new build dependencies but I figured this PR is a good way to get the conversation started.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -81,7 +81,7 @@ Front-end Dependencies
 ~~~~~~~~~~~~
 Before you get started, make sure you have the following installed:
 
-- ``jq`` – `Command-line JSON processor <https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html>`_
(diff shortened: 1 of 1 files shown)


uhurusurfa (commented):
These tools are only relevant to developers and should be moved int othe "Developer Environment" section.
It might be worth providing a link to each tools documentation.

README.rst
```diff
@@ -86,6 +76,19 @@ Installation
 
 Developer Environment
 ---------------------
+
+Front-end Dependencies
+~~~~~~~~~~~~
+Before you get started, make sure you have the following installed:
+
+- ``jq`` – `Command-line JSON processor <https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html>`_
```
uhurusurfa: The link should be to jq:
https://jqlang.org/download/

PR #1338, opened 2026-04-08
Fix #1335 staff access permissions for ticket queue checks

### LINKED ISSUE #1335: Multiple missing queue permission checks on delete/merge/attachment endpoints
## Summary

Several endpoints bypass the `ticket_perm_check()` / `huser.can_access_queue()` permission system that enforces per-queue staff access restrictions (when `HELPDESK_ENABLE_PER_QUEUE_STAFF_PERMISSION` is enabled).

## Most Critical: Cross-Ticket Attachment Deletion IDOR

**`views/staff.py` line 2038:**
```python
attachment = get_object_or_404(FollowUpAttachment, id=attachment_id)
```

The `attachment_del` view loads a ticket and validates queue permissions on it, then loads the attachment by ID independently without verifying the attachment belongs to that ticket. Any staff user can delete any attachment by specifying their own accessible ticket's ID with any attachment ID.

**Fix:** `get_object_or_404(FollowUpAttachment, id=attachment_id, followup__ticket=ticket)`

## Missing Queue Permission Checks (3 delete endpoints)

These delete views skip `ticket_perm_check()` while their corresponding add/list siblings enforce it:

1. **`ticket_cc_del`** - Deletes CC entries without queue check (but `ticket_cc_add` checks)
2. **`ticket_dependency_del`** - Deletes dependencies without queue check (but `ticket_dependency_add` checks)  
3. **`ticket_resolves_del`** - Deletes resolves-links without queue check

## Destructive Operation: merge_tickets

**`views/staff.py` line 1009** - `merge_tickets` performs a destructive merge operation without any queue permission validation, while `mass_update` (line 747) correctly calls `huser.can_access_queue(t.queue)`.

## Broader Scoping Issues

- **RSS feeds** - Queue-specific RSS feeds don't validate the requesting user has access to that queue
- **API** - The DRF API doesn't integrate `can_access_queue()` checks, leaking tickets across queue boundaries

## Context

These all affect deployments with `HELPDESK_ENABLE_PER_QUEUE_STAFF_PERMISSION = True` where staff should only see tickets from their assigned queues.

---
*Found via automated security audit. CWE-862: Missing Authorization, CWE-639: Authorization Bypass Through User-Controlled Key.*

DavidVadnais:
The original method granted full access to any is_staff user, making can_access_queue() always return True for staff  which, rendered every ticket_perm_check() call effectively a no-op when HELPDESK_ENABLE_PER_QUEUE_STAFF_PERMISSION=True. Fixed to return False for non-superusers when per-queue permissions are enabled.

### DIFF
--- a/src/helpdesk/user.py
+++ b/src/helpdesk/user.py
@@ -54,10 +54,12 @@ class HelpdeskUser:
         return Ticket.objects.filter(queue__in=self.get_queues())
 
     def has_full_access(self):
-        return (
-            self.user.is_superuser
-            or self.user.is_staff
-            or helpdesk_settings.HELPDESK_ALLOW_NON_STAFF_TICKET_UPDATE
+        if self.user.is_superuser:
+            return True
+        if helpdesk_settings.HELPDESK_ENABLE_PER_QUEUE_STAFF_PERMISSION:
+            return False
+        return self.user.is_staff or bool(
+            helpdesk_settings.HELPDESK_ALLOW_NON_STAFF_TICKET_UPDATE
         )
 
     def can_access_queue(self, queue):
--- a/src/helpdesk/views/feeds.py
+++ b/src/helpdesk/views/feeds.py
@@ -9,11 +9,13 @@ views/feeds.py - A handful of staff-only RSS feeds to provide ticket details
 
 from django.contrib.auth import get_user_model
 from django.contrib.syndication.views import Feed
+from django.core.exceptions import PermissionDenied
 from django.db.models import Q
 from django.shortcuts import get_object_or_404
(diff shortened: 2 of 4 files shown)


DavidVadnais:
Fixes #1335 

Tests: 
without HELPDESK_ENABLE_PER_QUEUE_STAFF_PERMISSION

super users and regular staff can see all tickets.

with HELPDESK_ENABLE_PER_QUEUE_STAFF_PERMISSION=True 

super users can see all tickets and regular staff can only see tickets they have permission to access.

uhurusurfa (approved):
Seems a pretty robust implementation. Nice work.

Issue #745, opened 2019-03-19
User should be able to change the password


jachym:
Currently, users are created manually using Django's admin interface. Non-staff users do not have possibility to log-in to the admin interface and change their password.

We should enable the user to set his/her password in the /helpdesk/settings/ page

timthelion:
Since django-helpdesk can be integrated into existing django applications, this feature won't always mesh well with that. I also wonder if there isn't already a 3rd party django app that would fulfill this need, so that the forms and views wouldn't even need to be programmed.

timthelion:
Isn't this already possible anyway? https://github.com/django-helpdesk/django-helpdesk/commit/d42fc23e08bd49094d35aac71b77bf264cf17ffa

gwasser:
There's a password form using the Django built-in stuff but I think `django-helpdesk` itself only displays it to staff users. The request as I understand it is for non-staff to also use that form.

jachym:
yes, the request is for non-staff members of the helpdesk

PR #1341, opened 2026-05-30
Initial setup: installed dependencies, ran migrations, collected stat…

emmanuelangelo4199:
…ic files


uhurusurfa:
What is the purpose of this pull request?

DavidVadnais:
This looks like only new libraries being added? 

If we are adding libraries we should use yarn when possible (like select2 in this pr.) And we should use the libraries somewhere.  I do not see initially how these are used.

DavidVadnais:
Is this the final goal? https://github.com/django-helpdesk/django-helpdesk/pull/1342/changes/5ec3d78dbb00c6d06ccdcd2edaea9fbb9a0185fb ( found this in the other pr #1342)

emmanuelangelo4199:
These are setup/config commits that should never have been submitted as a PR, was targeting #745

PR #1343, opened 2026-06-02
fix: show password change link to all authenticated users

### LINKED ISSUE #745: User should be able to change the password

emmanuelangelo4199:
Fixes #745

## Problem
Non-staff users had no way to change their password because the 
"Change password" link in the navigation was gated behind the 
HELPDESK_SHOW_CHANGE_PASSWORD setting, which defaults to False.

## Solution
Replaced the setting condition with `user.is_authenticated` so all 
logged-in users with a usable password can see and access the 
password change page.

## Changes
- `src/helpdesk/templates/helpdesk/navigation-header.html`: updated 
  the condition on the change password link

### DIFF
--- a/src/helpdesk/templates/helpdesk/navigation-header.html
+++ b/src/helpdesk/templates/helpdesk/navigation-header.html
@@ -48,7 +48,7 @@
           <div class="dropdown-menu dropdown-menu-end" aria-labelledby="userDropdown">
             <a class="dropdown-item" href="{% url 'helpdesk:user_settings' %}"><i class="fas fa-fw fa-user-cog"></i> {% trans "User Settings" %}</a>
             <a class="dropdown-item" href='{% url 'helpdesk:rss_index' %}'><i class="fas fa-fw fa-rss-square"></i> {% trans "RSS Feeds" %}</a>
-            {% if helpdesk_settings.HELPDESK_SHOW_CHANGE_PASSWORD and user.has_usable_password %}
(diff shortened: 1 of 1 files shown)


uhurusurfa (approved):
That is a logical change.

Benbb96:
If we accept that change, then this setting is not used anymore.

To me, the solution to allow non staff users updating their password is setting HELPDESK_SHOW_CHANGE_PASSWORD to True.

uhurusurfa:
I am not sure what the point of providing a user Id and password is if the user cannot change their own password.
It is a security risk to NOT be able to change your own password if you know your password has been compromised.

If external users should not be able to change their password then they should not have a login account.

Benbb96:
> I am not sure what the point of providing a user Id and password is if the user cannot change their own password. It is a security risk to NOT be able to change your own password if you know your password has been compromised.
> 
> If external users should not be able to change their password then they should not have a login account.

I completely agree with you. I just wanted to point out that the setting is now useless so it could be removed. But that would be a small breaking change.

uhurusurfa:
@Benbb96 - ah!! - ok, I misunderstood what you had written.
I will remove the setting and increment the version for a release shortly.

PR #1344, opened 2026-06-04
Cleanup redundant documentation and make teams mode default to off

uhurusurfa:
The change password link was controlled by a setting named HELPDESK_SHOW_CHANGE_PASSWORD which was removed and the link is now always visible.
This PR removes all references to HELPDESK_SHOW_CHANGE_PASSWORD in the documentation.

Teams mode no longer enabled by default:

- Teams mode requires Pinax libraries which have not been maintained in a number of years and appear to be abandoned so the HELPDESK_TEAMS_MODE_ENABLED is defaulted to "false" now

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,14 +26,14 @@ help:
 .PHONY: develop
 develop:
 	$(UV) venv
-	$(UV) sync --all-extras --dev --group test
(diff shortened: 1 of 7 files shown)


Benbb96 (approved):
Looks good to me !

PR #1345, opened 2026-06-05
Add a kanban board to django-helpdesk

DavidVadnais:
Added a Kanban board view to django-helpdesk with sidebar navigation
drag-and-drop status updates, and a working due-date filter.


DavidVadnais:
image

You can drag into into new columns and it updates that status. 

1. Filtering by due date (or no fliter with clear)
2. side bar nav to kanban board

DavidVadnais:
I could add a setting similar to knowledge base that lets this feature be disabled.  Do you think we want that?

Benbb96 (commented):
I think that's a nice feature to have ! And yes a setting to opt-out is important for the customization.

The code looks clean to me. Does it render well with a lot of tickets ?  
Maybe the CSS should live in a separate static file ?

DavidVadnais:
Okay I added the feature toggle and docs.  I'm going to start on the separate css and test lots of tickets next.

DavidVadnais:
@Benbb96 It does render well on my database ~11,000 tickets however, I have received complaints on the column size being fixed and the initial query is slow (page load is about 2 seconds with 11,000 tickets). Do you think that's fast enough?

DavidVadnais:
Fixed the columns being fixed max width. Still ~2 seconds for 11,000 tickets.

src/helpdesk/views/staff.py
```diff
@@ -2243,17 +2254,23 @@ def kanban_board(request):
...
-                "label": status_label,
-                "tickets": tickets.filter(status=status_value).order_by(
-                    "due_date", "-modified"
-                ),
-            }
-        )
+    all_tickets = sorted(
+        tickets,
+        key=lambda t: (t.due_date is None, t.due_date, -t.modified.timestamp()),
+    )
```
Benbb96: Isn't it better to keep the `order_by` on the DB query to avoid a `sorted` ? I think it should reduce the process time.

src/helpdesk/views/staff.py
```diff
@@ -2220,3 +2220,79 @@ def delete_checklist_template(request, checklist_template_id):
...
+            "priority",
+            "status",
+            "due_date",
+            "modified",
+            "queue__title",
+            "assigned_to__username",
+        )
+        .filter(queue__in=huser.get_queues())
+    )
+
```
uhurusurfa, line 2238: This ORM statement produces this query (I have used an asterisk in place of the "only" fields that it specifically selects) using the default setting for HELPDESK_ENABLE_PER_QUEUE_STAFF_PERMISSION of False:
```
SELECT * FROM "helpdesk_ticket" 
       INNER JOIN "helpdesk_queue" ON ("helpdesk_ticket"."queue_id" = "helpdesk_queue"."id")
       LEFT OUTER JOIN "auth_user" ON ("helpdesk_ticket"."assigned_to_id" = "auth_user"."id")
  WHERE "helpdesk_ticket"."queue_id" IN (SELECT "helpdesk_queue"."id" FROM "helpdesk_queue") 
ORDER BY "helpdesk_ticket"."id" ASC
```

The get_queues method is probably not a good option to use because what it returns is dependent on the HELPDESK_ENABLE_PER_QUEUE_STAFF_PERMISSION.
So basically it gets every ticket in the database then makes sure each ticket has a queue_id set. I am not sure this is necessarily the right choice but if you only want to see tickets assigned to a queue (when HELPDESK_ENABLE_PER_QUEUE_STAFF_PERMISSION is False) then a much more efficient query would be:
`Ticket.objects.select_related("queue", "assigned_to").filter(queue_id__isnull=False)`

If the HELPDESK_ENABLE_PER_QUEUE_STAFF_PERMISSION is set to True then it filters based on queues assigned to the user. So the first optimisation can be to make the filter statement different based on whether HELPDESK_ENABLE_PER_QUEUE_STAFF_PERMISSION is True or False.

In the case where HELPDESK_ENABLE_PER_QUEUE_STAFF_PERMISSION is True, this will become an increasingly poor performing query as the database grows since it will always first select every ticket in the DB before applying queue filtering.

NOTE that allowing the user to NOT have a time interval selection means every ticket would be rendered to the UI and that would create a situation where over a long enough period of time the query will become increasingly slow as ticket numbers go up and will also slow rendering in the browser and may bring the browser to a crawl.

uhurusurfa (commented):
This is specifically about the query that retrieves the data for the Kanban board

src/helpdesk/views/staff.py
uhurusurfa: Agreed.

uhurusurfa:
> You can drag into into new columns and it updates that status.
> 
>     1. Filtering by due date (or no fliter with clear)
> 
>     2. side bar nav to kanban board

Should it not be also showing past due tickets that are not completed? Or perhaps have a filter that includes or excludes past due tickets.

src/helpdesk/views/staff.py
uhurusurfa, line 2238: This ORM statement produces this query (I have used an asterisk in place of the "only" fields that it specifically selects) using the default setting for HELPDESK_ENABLE_PER_QUEUE_STAFF_PERMISSION of False:
```
SELECT * FROM "helpdesk_ticket" 
       INNER JOIN "helpdesk_queue" ON ("helpdesk_ticket"."queue_id" = "helpdesk_queue"."id")
       LEFT OUTER JOIN "auth_user" ON ("helpdesk_ticket"."assigned_to_id" = "auth_user"."id")
  WHERE "helpdesk_ticket"."queue_id" IN (SELECT "helpdesk_queue"."id" FROM "helpdesk_queue") 
ORDER BY "helpdesk_ticket"."id" ASC
```

The get_queues method is probably not a good option to use because what it returns is dependent on the HELPDESK_ENABLE_PER_QUEUE_STAFF_PERMISSION.
So basically it gets every ticket in the database then makes sure each ticket has a queue_id set. I am not sure this is necessarily the right choice but if you only want to see tickets assigned to a queue (when HELPDESK_ENABLE_PER_QUEUE_STAFF_PERMISSION is False) then a much more efficient query would be:
`Ticket.objects.select_related("queue", "assigned_to").filter(queue_id__isnull=False)`

If the HELPDESK_ENABLE_PER_QUEUE_STAFF_PERMISSION is set to True then it filters based on queues assigned to the user. So the first optimisation can be to make the filter statement different based on whether HELPDESK_ENABLE_PER_QUEUE_STAFF_PERMISSION is True or False.

In the case where HELPDESK_ENABLE_PER_QUEUE_STAFF_PERMISSION is True, this will become an increasingly poor performing query as the database grows since it will always first select every ticket in the DB before applying queue filtering.

DavidVadnais:
Its now taking about 75% as long. ~1.5 seconds for 11,000 tickets.

src/helpdesk/views/staff.py
DavidVadnais: I think ive got this working.

src/helpdesk/views/staff.py
DavidVadnais: pushed this to the DB.

DavidVadnais:
## Unchecked exclude past due
image


## checked exclude past due
image

DavidVadnais:
Okay now I want to set a variable somewhere that defaults to 2 weeks. That way if you have a massive database of tickets the view is still quick to get into

src/helpdesk/views/staff.py
```diff
@@ -2220,3 +2220,99 @@ def delete_checklist_template(request, checklist_template_id):
...
+        "assigned_to__username",
+    )
+    if (
+        helpdesk_settings.HELPDESK_ENABLE_PER_QUEUE_STAFF_PERMISSION
+        and not request.user.is_superuser
+    ):
+        queue_ids = list(huser.get_queues().values_list("pk", flat=True))
+        tickets = base_qs.filter(queue_id__in=queue_ids)
+    else:  # super users
+        tickets = base_qs.filter(queue_id__isnull=False)
```
Benbb96: This else part is useless I think because a ticket is always attached to a queue.

demodesk/config/settings.py
```diff
@@ -118,6 +118,10 @@
 # Should the Knowledgebase be enabled?
 HELPDESK_KB_ENABLED = True
 
+# Should the Kanban board be enabled?
+HELPDESK_KANBAN_ENABLED = True
+HELPDESK_KANBAN_DEFAULT_DUE_WEEKS = None
```
Benbb96, line 123: Shouldn't it defaults to 2 ?

src/helpdesk/urls.py
```diff
@@ -139,6 +139,18 @@ def get_context_data(self, **kwargs):
...
+                path("kanban/", staff.kanban_board, name="kanban"),
+                path(
+                    "tickets/<int:ticket_id>/kanban-update/",
+                    staff.kanban_update_ticket,
+                    name="kanban_update",
+                ),
+            ]
+            if helpdesk_settings.HELPDESK_KANBAN_ENABLED
+            else []
+        ),
```
Benbb96: For a better readability, I would use the same pattern as on line 184, appending to `urlpatterns`

src/helpdesk/views/staff.py
```diff
@@ -2220,3 +2220,99 @@ def delete_checklist_template(request, checklist_template_id):
...
+
+    now = timezone.now()
+    default_due_weeks = helpdesk_settings.HELPDESK_KANBAN_DEFAULT_DUE_WEEKS or None
```
Benbb96: If the defaults is 2, it has to be here too

src/helpdesk/views/staff.py
```diff
@@ -2220,3 +2220,99 @@ def delete_checklist_template(request, checklist_template_id):
...
+        cutoff = now + timedelta(weeks=due_weeks)
+        upcoming_q = Q(due_date__isnull=False, due_date__gte=now, due_date__lte=cutoff)
+        overdue_q = Q(
+            due_date__isnull=False, due_date__lt=now, status__in=Ticket.OPEN_STATUSES
+        )
+        tickets = tickets.filter(
+            upcoming_q if exclude_overdue else upcoming_q | overdue_q
+        )
+
+    tickets = tickets.order_by(F("due_date").asc(nulls_last=True), "-modified")
```
Benbb96, line 2277: Looks good to me now. Maybe it needs some unit tests to find the edge cases. But since you already test it against your real DB, I think it could be enough.

demodesk/config/settings.py
uhurusurfa: Definitely needs a default and should not be allowed to be empty IMO.

uhurusurfa: Possibly should have a MAX limit too as a Kanban board makes no sense for stuff that is over a year old and would ensure nobody can bring the server to its knees by choosing a very large number as the ticket count gets ever larger over time.
Keep in mind that there is a query per person and if you have a few people working simultaneously the time to run the query will go up.

uhurusurfa: At very least, it should exclude DONE tickets that are older than a certain time as it makes no sense to render tens, hundreds or thousands of tickets that were long since completed and no longer of interest.

DavidVadnais: > Shouldn't it defaults to 2 ?

I defaulted to None in the demodesk because I figured grabbing everything would make it most likely to pick up the few example tickets. If the variable is unset I'm defaulting to two week in `settings.py`.  Do we want this controlled somewhere else?

``` python
# default number of weeks ahead to show tickets on the kanban board (0 or None = no filter)
HELPDESK_KANBAN_DEFAULT_DUE_WEEKS = getattr(
    settings, "HELPDESK_KANBAN_DEFAULT_DUE_WEEKS", 2
)
```

> Possibly should have a MAX limit too as a Kanban board makes no sense for stuff that is over a year old and would ensure nobody can bring the server to its knees by choosing a very large number as the ticket count gets ever larger over time.
> Keep in mind that there is a query per person and if you have a few people working simultaneously the time to run the query will go up.

Good point this will be my next commit. I think I will add a variable that defaults to 6 weeks that sets the max back closed date to pull up

DavidVadnais: https://github.com/django-helpdesk/django-helpdesk/pull/1345/commits/0ae94b420c4628ab20598abc02a25960f44f9b25 should handle this but do we want a control on the page to allow users to change this? It would potentially allow some one to beat the server up if they used this without thinking.

src/helpdesk/views/staff.py
DavidVadnais: https://github.com/django-helpdesk/django-helpdesk/pull/1345/commits/6427628c6df717b709eb3bb4b062d9bf3196f99a

src/helpdesk/urls.py
DavidVadnais: https://github.com/django-helpdesk/django-helpdesk/pull/1345/commits/3cdcfbb730800fe740eae9684edb67f357b3d371 good point

src/helpdesk/views/staff.py
DavidVadnais: https://github.com/django-helpdesk/django-helpdesk/pull/1345/commits/8e1dad940b0015b2e796b563e070715a4d3e4917  I put some tests in. Probably not enough.

demodesk/config/settings.py
uhurusurfa: In my view the date limit on what you would see is going to be different per type of usage for the ticket system, typically affected by the expected worst case life time of a ticket.
I would make it a standard setting defined as part of the deployment and not selectable in the UI (or at worst only by a superuser)

DavidVadnais: Sounds good to me.

uhurusurfa (dismissed):
I have run it through a few scenarios on my laptop and seems to work really nicely - great job 🚀

uhurusurfa:
Since this is a significant new feature, perhaps bump the version to 3.0.0 instead of 2.3.0 in pyproject.toml and we can release it.

demodesk/config/settings.py
Benbb96: Alright, I didn't see it was the demodesk settings. Make sense to force None here

Benbb96 (dismissed):
It's all good for me. Thanks for the tests.

And in my mind, there's no need to bump a major version since there's no breaking change. It's juste a new feature (which can be opt-out). So only the minor is enough.

DavidVadnais:
The only change is the version bump.  I need re-approval. 

Thank you both for all the feedback, the page is working much better then before.

DavidVadnais:
I also added tags to the repo for [agile](https://github.com/topics/agile), [scrum](https://github.com/topics/scrum), and [kanban](https://github.com/topics/kanban) since we can now support that on top of ticketing. 

Feel free to add or remove these as desired.

uhurusurfa (approved):
This is a really nice improvement to the app.
Thanks for all the work.

PR #1346, opened 2026-06-10
codecoverag and docker push

DavidVadnais:
Tried to fix the code coverage but I cant see our existing codecov access token.  I'm also fine with removing this but our token hasn't updated in 5 years so we should do something. 

Also tried to fix docker push.  I can't test either of these so its a bit of a guess. 

Edit: 
`uv run --with django~="5.0 "coverage run quicktest.py` works with warnings.


uhurusurfa:
This was set up by @timthelion  and he has the login credentials for docker but seems to have gone AWOL.

There is the secret key in the repo that we may be able to expose in some way by logging it to a workflow log file but not sure it is possible as the workflow engine may prevent rendering it in a string context.

The docker image was constructed specifically for his own particular use case and IMO is not generic enough as it is configured currently.

I suggest we remove the docker workflow for now but retain the docker files allowing people to generate a docker image if they choose.

We will probably have to create a new Docker user ID and securely share the login credentials at some point unless they p[rovide some form of team based management of docker images in the future.

DavidVadnais:
Do we want to also remove the coverage icon?

uhurusurfa:
It is just the docker workflow I would remove for now (or disable it by setting tags to "x*" instead of "v*").

For code coverage I think coveralls.io will work better than codecov.io since it seems to allow anyone admin access to an open source project to configure the project using their personal login to coveralls. Verify this by logging in to coveralls.io using your Github account and you should see the last coveralls upload for Django-Helpdesk was in 2016.

DavidVadnais:
https://coveralls.io/github/django-helpdesk/django-helpdesk

somehow coverage has gone up since we stopped looking at 🤔

uhurusurfa (approved):
Nice work

PR #1350, opened 2026-07-21
Fix demo queue locale mismatch and enable CC notifications

DavidVadnais:
Queue locale was 'en-US' but EmailTemplate fixtures are keyed to 'en', so send_templated_mail() always failed to find a template and silently dropped every outgoing email. Also enable enable_notifications_on_email_events and add a TicketCC record so the 'Copies to' feature is exercisable out of the box.


DavidVadnais:
Pros: You can test the email functionality after `make rundemo`,  

Cons: it drops the emails into the cli were `make rundemo` was run which may confuse new users.

Should we make this change? @Benbb96 @uhurusurfa

uhurusurfa (approved):
I am fine with this change or adding a small script to enable email notifications which directly updates the DB as a separate step and documenting it. 

Leave it up to you to decide what is best since you are the active user and neither Ben nor I are currently using the app.

PR #1349, opened 2026-07-18
Enhancement/1093 redesign ticket description table

gurupratap-matharu:
This PR is for Issue #1093 to improve the layout of the ticket detail page viewed by staff.
I'm attaching some screenshots for you to review.

I'm not a UI designer but worked on something similar recently. The layout is a mix of inspiration from the the issue screenshot and tabler ui kit components. 

Action buttons sit nicely hidden in top right dropdowns
Removed a lot of borders and nested tables to bring space and clarity

Ticket Description 
ticket-description

This feature brings Nav Tabs interface. The complex template has one giant form which is in the Respond tab. Breaking this into four tabs - Followups / Extra Details / Attachments / Checklists would required considerable changes in the backend code. So I've kept it simple with two tabs: Followup & Respond

Followup Pane
ticket-followups

Respond Pane
ticket-respond

Some basic unit tests are added to test the view, url and template.

PR #1351, opened 2026-07-23
Ruff 0.16.0 updates

DavidVadnais:
Both `ruff check` and `ruff format --check` are fully clean, and `./quicktest.py` reports all 191 tests passing (OK, 2 skipped).

Summary of what was fixed (86 remaining ruff errors after the earlier auto-fix pass, now 0):

- RUF012 (17×): added ClassVar annotations to Django convention list/dict class attrs (fixtures, permission_classes, inlines, MIDDLEWARE, etc.)
- G201/LOG015/TRY401/BLE001 (email.py): converted .error(..., exc_info=True) --> .exception(...), replaced root-logger calls with the module logger, added justified noqa on two intentionally-broad STARTTLS fallback excepts
- UP031 (31×): remaining %-format strings converted to f-strings
- INT002 (4×): moved .format() calls outside _() so gettext translates the template, not the pre-interpolated string
- DTZ00x: swapped naive datetime.now()/date.today() for django.utils.timezone equivalents in views/commands/tests; forms.py's custom-field date parsing now chains .replace(tzinfo=...) directly on the strptime() result
- SIM102 (5×): collapsed nested ifs with and
- EXE001: removed vestigial shebangs from 4 non-executable management commands
- SIM113: converted a manual counter to enumerate() in staff.py
- TRY002: replaced bare raise Exception(...) with NotImplementedError (tests/utils.py) and ImproperlyConfigured (standalone settings.py)
- Ran ruff format . to normalize the 13 files the reformatted rules touched

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,7 +1,7 @@
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
     # Ruff version.
-    rev: v0.12.9
+    rev: v0.16.0
     hooks:
       # Run the linter.
       - id: ruff-check
--- a/demodesk/config/settings.py
+++ b/demodesk/config/settings.py
@@ -5,7 +5,6 @@ Django settings for django-helpdesk demodesk project.
 
 import os
 
-
 # Build paths inside the project like this: os.path.join(BASE_DIR, ...)
 BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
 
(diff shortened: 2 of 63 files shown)


DavidVadnais:
Once this is merged we can pull it into #1349 and that will allow that to pass. This will also check the fixes to coverall done in https://github.com/lemurheavy/coveralls-public/issues/1808

DavidVadnais:
I ran through the GUI and everything seems to be working. 

Unfortunately the coverage reports is still not passing. I'm working with the coveralls team now.

uhurusurfa (approved):
Went through all changes and they seem to be as stated- mostly formatiing.

PR #1358, opened 2026-07-29
Fix submitter email template always using "updated" regardless of ticket status

buoncri:
When a ticket is resolved or closed, the submitter email subject showed the wrong status because the template name was hardcoded to "updated_submitter". Use the same template_prefix logic already applied to owner/cc notifications.

### DIFF
--- a/src/helpdesk/update_ticket.py
+++ b/src/helpdesk/update_ticket.py
@@ -145,12 +145,10 @@ def process_email_notifications_for_ticket_update(
             or (follow_up.new_status in (Ticket.RESOLVED_STATUS, Ticket.CLOSED_STATUS))
         )
     ):
-        # Use hard coded prefix for submitter updates on  tickets for backwards compatibility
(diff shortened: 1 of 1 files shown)


buoncri:
Note: reopening a ticket (REOPENED_STATUS) is not covered by this change.  It falls through to the else branch and uses "updated_". Supporting it would require adding a "reopened_" prefix with corresponding templates (reopened_submitter, reopened_owner, reopened_cc) in the database. 
Same pattern, out of scope for this PR.

DavidVadnais:
I looked to see what backwards compatibility the original author was mentioning. 

Removing the hardcoded template lets one year dormant `resolved_submitter/closed_submitter` templates go live to real submitters for the first time in a year, so any site that deleted or miscustomized them (assuming they were unused) will now silently drop (due to translations missing) or send wrong emails.  

I think this concern is marginal so we should be good to go.

DavidVadnais (approved):

Issue #1361, opened 2026-07-30
Status badge colors break on non-English locales in ticket list


buoncri:
**Describe the bug**
Problem with the badge colors: they're mapped by English status strings in the client-side JS, so they default to gray on any locale other than en.

**To Reproduce**
1. use LANGUAGE_CODE = "it" in settings

**Expected behavior**
All status badges are gray

**Screenshots**

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/9be9ea6e-4045-4dfc-baf9-7b4e85eda38d" />

**Desktop (please complete the following information):**
 - OS: Debian Sid
 - Browser: Firefox
 - Version: 140.13

**Additional context**

buoncri:
**Root cause**
In ticket_list.html, the status column renderer uses a hardcoded contextMap keyed by display string:
```
const contextMap = {
    "Open": "danger",
    "Reopened": "warning",
    "Resolved": "success",
    "Closed": "success",
    "Duplicate": "secondary"
};
```
But DatatablesTicketSerializer.get_status() returns obj.get_status, which calls get_status_display() — a localized string. In Italian (just for example), that's "Risolto" instead of "Resolved", so the key never matches and every badge falls back to:
```
const context = contextMap[data] ?? "secondary";  // always gray
```
**Proposed fix**
Follow the same pattern already used for priority colors (get_priority_css_class → row_class in the JSON). Move the color logic server-side, immune to translations. A new setting (helpdesk/settings.py):
```
DEFAULT_TICKET_STATUS_CSS_CLASSES = {
    OPEN_STATUS: "danger",
    REOPENED_STATUS: "warning",
    RESOLVED_STATUS: "success",
    CLOSED_STATUS: "success",
    DUPLICATE_STATUS: "secondary",
}
TICKET_STATUS_CSS_CLASSES = getattr(
    settings, "HELPDESK_TICKET_STATUS_CSS_CLASSES", DEFAULT_TICKET_STATUS_CSS_CLASSES
)
```

The rest follows naturally: a property on Ticket that reads from this setting; a field on DatatablesTicketSerializer that exposes it; and a one-line change in the JS renderer to use row.status_css_class instead of the hardcoded map.

[...]

PR #1362, opened 2026-07-31
Fix ticket status rendering without translations

buoncri:
Updated ticket status rendering to use translation for status labels.
Resolve issue #1361.

### DIFF
--- a/src/helpdesk/templates/helpdesk/ticket_list.html
+++ b/src/helpdesk/templates/helpdesk/ticket_list.html
@@ -421,15 +421,19 @@
                            visible: false,
                       }, {
                            data: "status",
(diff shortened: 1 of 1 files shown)


DavidVadnais:
## Testing

```
cat demodesk/config/local_settings.py
LANGUAGE_CODE = "de"

python manage.py runserver
```

image

DavidVadnais (approved):

uhurusurfa:
Relying on the visually displayed text as a constant is not a good idea. If someone prefers to use "Completed" instead of "Closed" for the English translation then the code will break again.

It should be using the underlying status values that do not change no matter what language implementation is used to provide the correct substring to identify the desired badge.

PR #1365, opened 2026-07-31
Make all_tickets ticket name column wrap

### LINKED ISSUE #1364: In 'all tickets' view ticket title could better utilize space
> I'm not too thrilled with the "ticket description is in one line" because we start every ticket with the same formula, so with ticketing right now I can't see what the actual tickets are.  see attached.  I have a 14" laptop, not a 32"OLED monitor so I can't make it any wider.   Is there some way to remedy this?
> 
> Mouseover works, but that's tedious.

source: @jhohum 

Image

At least wrapping since there seems to be space.  Maybe adjustable column widths.

Do nothing and use deselecting columns.

DavidVadnais:
fixes #1364 

image

### DIFF
--- a/src/helpdesk/static/helpdesk/helpdesk-extend.css
+++ b/src/helpdesk/static/helpdesk/helpdesk-extend.css
@@ -113,11 +113,18 @@ pre {
 	border: 1pt solid white;
 }
 
+/* Ticket titles wrap onto several lines rather than being cut off with an
+   ellipsis, so the whole title is readable without hovering over it. */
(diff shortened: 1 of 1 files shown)


DavidVadnais:
I tried adding a column size adjuster and I dont think it adds much but ill leave this image here if we really want to push for it.

image

I used this https://www.npmjs.com/package/datatables.net-colresize-unofficial

uhurusurfa (approved):
Makes sense to me.

Issue #1369, opened 2026-08-05
Sorting dropdown on /tickets/ by "Last Followup" (and other sortable columns)


buoncri:
**Is your feature request related to a problem? Please describe.**

The "Sorting" dropdown in `src/helpdesk/templates/helpdesk/filters/sorting.html` only offers six fields (Created, Title, Queue, Status, Priority, Owner), while `ALLOWED_SORTS` in `src/helpdesk/views/staff.py` and `DATATABLES_ORDER_COLUMN_CHOICES` in `src/helpdesk/query.py` already accept `id`, `due_date`, `submitter_email`, `last_followup`, `kbitem`.

`last_followup` is an annotation, not a model field: `Query.__run__()` in `query.py` orders the queryset before the annotation exists, so it cannot be used as a sort field today.

**Describe the solution you'd like**

1. Add `last_followup`, `id`, `due_date`, `submitter_email`, `kbitem` to the sorting dropdown (`kbitem` conditional on `HELPDESK_KB_ENABLED`).
2. Add `id`, `due_date`, `submitter_email` to `ALLOWED_SORTS`.
3. Make `last_followup` sortable: extract its `Subquery` into a helper and attach it to the queryset **before** `order_by()` in `__run__()` (including reverse `-last_followup`), reusing the helper in `get_datatables_context()`.

**Additional context**

- Use case: we maintain a custom view listing tickets by last followup to see live activity; sorting in `/tickets/` would let us replace it with a saved query.
- Other fields are real model fields and already pass `_validate_sorting()` — UI + `ALLOWED_SORTS` only.
- Saved-query flow already persists `sorting`, no changes needed.

uhurusurfa:
Sounds reasonable to me.

PR #1370, opened 2026-08-05
feat: expose all sortable columns in ticket list sorting filter

buoncri:
Add id, last_followup, due_date, submitter_email and kbitem to the Sorting dropdown in the /tickets/ filters, matching the columns already supported by ALLOWED_SORTS and DATATABLES_ORDER_COLUMN_CHOICES.

last_followup is an annotation rather than a model field, so ordering by it previously raised a FieldError on both the datatables and timeline endpoints. Extract the followup Subquery into a reusable helper and annotate the queryset before order_by() in __run__().

(issue #1369)

### DIFF
--- a/src/helpdesk/query.py
+++ b/src/helpdesk/query.py
@@ -117,6 +117,29 @@ def _validate_sorting(sorting):
         raise SuspiciousOperation(f"Disallowed sort field: {sorting}")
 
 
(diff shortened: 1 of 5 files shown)


uhurusurfa (approved):
Nice work 👍

Issue #1363, opened 2026-07-31
if someone changes the default status names then the colloration on the ticket list screen will fail.


DavidVadnais:
**Describe the bug**
> Relying on the visually displayed text as a constant is not a good idea. If someone prefers to use "Completed" instead of "Closed" for the English translation then the code will break again.
> 
> It should be using the underlying status values that do not change no matter what language implementation is used to provide the correct substring to identify the desired badge.

 -- @uhurusurfa 

**To Reproduce**
1. Change the underlying ticket status names
2. Go to the ticket list view
3. see the colors no longer working

**Expected behavior**
If a users changes from default tickets status the color scheme still works.

**Additional context**
Previous context for how we got here/ 

https://github.com/django-helpdesk/django-helpdesk/pull/1362
https://github.com/django-helpdesk/django-helpdesk/issues/1361

buoncri:
You're right — the root cause is that the renderer keys off the **displayed text** instead of a stable identifier. That breaks on two axes: localized display strings (any non-English locale), and admins renaming/replacing labels (e.g. "Completed" instead of "Closed" in English).

The fix is to stop keying the JS by display text entirely.

**Design:**

1. **New setting** `HELPDESK_TICKET_STATUS_CSS_CLASSES` — `{status_id: bootstrap_class}`, default matching today's scheme. Plus `HELPDESK_TICKET_PRIORITY_CSS_CLASSES` for the parallel priority issue — the current JS priority map and the model's `_get_priority_css_class` disagree today: priorities 3/4 are colored in the UI but `""` in the model.

2. **Model** — add `Ticket.get_status_css_class` reading the setting; refactor `_get_priority_css_class` to read its setting. One source of truth.

3. **Serializer** (`DatatablesTicketSerializer`) — expose `status_css_class` (computed server-side from the stable status id, never from the localized label).

4. **JS** (`ticket_list.html`) — the renderer uses `row.status_css_class` directly; the hardcoded `contextMap` is removed for both status and priority.

This also fixes the deeper gap: a custom status (e.g. `HELPDESK_TICKET_FORKED_STATUS = 6`) gets a badge color by adding one entry to the setting — today it requires editing the template.

**Tests to include:**
- Renaming the English label (e.g. Closed → Completed) keeps colors
- Non-English locale keeps colors (key regression)
- Custom status with a configured color renders correctly
- Unknown status falls back to the default gray

**Question:** where should the color mapping live?

[...]

PR #1371, opened 2026-08-06
fix: key status/priority badge color by stable id, not localized label

buoncri:
The ticket list badge colors were keyed by the displayed status label (translated via {% translate %}), so they broke on non-English locales and when admins renamed/replaced status labels. Priority badges used a hardcoded client-side map.

Move both to a single server-side source of truth:

- New settings HELPDESK_TICKET_STATUS_CSS_CLASSES and HELPDESK_TICKET_PRIORITY_CSS_CLASSES (dicts keyed by the stable status/priority id) with defaults matching the previous UI scheme.

- Ticket.get_status_badge_class / get_priority_badge_class read from the settings; the legacy get_priority_css_class (row_class + non- DataTables templates) is kept unchanged to preserve the API.

- DatatablesTicketSerializer exposes status_badge_class and priority_badge_class per row.

- ticket_list.html renders from row.*_badge_class with a 'secondary' fallback; the hardcoded contextMaps and the {% translate %} keys are removed, so custom statuses get a color by editing settings only.

Ref #1363


Benbb96 (approved):
LGTM

uhurusurfa:
Nice work.

PR #1377, opened 2026-08-09
feat: render forms with our own Bootstrap 5 templates

### LINKED ISSUE #1375: Form rendering uses a Bootstrap 3-era package while the UI is Bootstrap 5
## Summary

The UI ships Bootstrap 5, but form fields are rendered by `django-bootstrap4-form`, whose markup is largely Bootstrap **3**. Three user-facing forms are getting CSS classes that do not exist in the stylesheet we load.

## Evidence

`package.json` pins the UI to Bootstrap 5:

```json
"bootstrap": "5.3.8",
"datatables.net-bs5": "^2.3.4",
```

The `field.html` template inside `django-bootstrap4-form` 4.0.2 emits:

| Class emitted | Reality |
|---|---|
| `form-group` | removed in Bootstrap 5 |
| `has-error` | Bootstrap 3, gone in 4 |
| `control-label` | Bootstrap 3, gone in 4 |
| `<div class="checkbox">` / `<div class="radio">` | Bootstrap 3, replaced by `form-check` in 4 |
| `text-muted` | deprecated in 5.3 in favour of `text-body-secondary` |

In practice the affected fields lose their bottom margin and checkboxes and radios are unstyled, because the browser is handed class names that appear nowhere in the loaded CSS.

## Scope is smaller than it looks

Seven templates run `{% load bootstrap4form %}`, but only three actually use the filter:

- `helpdesk/public_create_ticket_base.html`
- `helpdesk/edit_ticket.html`
- `helpdesk/user_settings.html`

The other four load the tag library and never call it: `create_ticket.html`, `public_homepage.html`, `ticket.html`, `registration/login.html`. Those four `{% load %}` lines can be deleted whatever we decide here.

The upstream package is also small: three templates and a 3 KB templatetag module. Its last two releases were September 2018 and October 2024, and it declares only `django` while importing `packaging`, which is why #1374 has to declare `packaging` explicitly.

## Why `django-helpdesk/django-bootstrap5-form` is not a drop-in

This org already publishes a Bootstrap 5 version, which is the obvious candidate. It cannot currently be installed alongside django-helpdesk:

```
Because all versions of django-bootstrap5-form depend on django>=3,<4
and you require django-bootstrap5-form, we can conclude that you require
django>=3,<4. And because you require django>=4.2, we can conclude that
your requirements are unsatisfiable.
```

It pins `django<4.0`, while we require `django>=4.2`. Last pushed October 2024.

## Options

**1. Fix and release `django-bootstrap5-form`.** The repo is in this org, so lifting the Django pin and cutting a release is within reach. Cost: we take on maintaining a second published package for three filter calls.

**2. Drop the dependency.** Django 4.1+ renders forms through templates, so `Form.template_name` can point at a small template of our own with correct Bootstrap 5 classes. Three call sites to convert, roughly forty lines of template to own, no external dependency, and the markup becomes correct.

**3. Adopt a maintained third-party library**, such as `django-crispy-forms` with `crispy-bootstrap5`, or zostera's `django-bootstrap5`. Well maintained, but it trades one dependency for a larger one right after #1372 and #1374 reduced the dependency surface.

**4. Do nothing** and accept the degraded rendering.

## Suggested direction

Option 2. The dependency is barely maintained, emits markup two generations behind the UI, and is used three times. Owning a small template seems cheaper than maintaining a published package or adopting a bigger one.

Happy to be argued into option 1 if we would rather keep a reusable package in the org, or into option 3 if anyone expects the form rendering needs to grow.

## Questions

1. Which option do we want?
2. Is anyone already touching these templates? I noticed `enhancement/improve-kb-list-detail-ui` in flight and would rather not collide with it.
3. Would before/after screenshots of the three forms help settle it? I can produce them.

Benbb96:
Closes #1375, following @uhurusurfa's call for option 2.

## The problem

The UI ships Bootstrap 5.3.8, but form fields were rendered by `django-bootstrap4-form`, whose markup is largely Bootstrap **3**: `form-group`, `control-label`, `has-error`, `<div class="checkbox">`. None of those classes exists in the stylesheet we load, so the affected fields lost their bottom margin and checkboxes rendered unstyled.

The org's own `django-bootstrap5-form` looked like the obvious replacement, but it pins `django<4.0` while we require `django>=4.2`, so it cannot be installed alongside us. With the filter used in exactly three templates, owning the rendering is cheaper than maintaining a published package or adopting a larger one.

## What this does

Adds `{% bootstrap5_form form %}` and two templates under `helpdesk/include/`.

Widget classes are assigned in Python rather than in the template, because a widget rendered through `{{ field }}` has already resolved its attributes and cannot be given a class afterwards. Selects get `form-select`, checkboxes and radios `form-check-input`, everything else `form-control`, and hidden inputs are left alone. Classes already declared in `forms.py` are preserved when they are unrelated to the Bootstrap widget family, and the obsolete `form-control-file` is dropped.

The three family classes are treated as mutually exclusive. `form-control` sets `width: 100%` while `form-check-input` sizes a square box, so a widget carrying both renders wrong. This is not hypothetical: `forms.py` declared `form-control` on the checkbox and the select it builds for custom fields, which is why a boolean custom field used to reach the browser as `form-check-input form-control` and render 26px wide instead of 16px. Those two declarations are corrected at the source as well.

Errors use `invalid-feedback d-block` rather than plain `invalid-feedback`. The latter is only revealed by an adjacent `.is-invalid`, and that relationship does not hold for the choice groups, where the inputs are nested one level deeper.

## Also in here

- The four `{% load bootstrap4form %}` lines in templates that never called the filter are gone: `create_ticket.html`, `public_homepage.html`, `ticket.html`, `registration/login.html`.
- `bootstrap4form` is removed from `INSTALLED_APPS` in `demodesk`, `standalone` and `quicktest.py`, and from the example in `docs/install.rst`.
- `packaging` is dropped. It was only ever declared because `django-bootstrap4-form` imports it without declaring it, and nothing else in a default install needs it.
- `docs/upgrade.rst` gains a note telling operators to drop `'bootstrap4form'` from their own `INSTALLED_APPS`, and what to do if they have overridden one of the three templates.
- The submit button on the user settings form was a bare `<input type="submit">` with no class and no `gettext` call. It now follows `edit_ticket.html`.

## Verification

Full suite passes: 253 tests, 2 skipped, with `bootstrap4form` and `packaging` genuinely uninstalled rather than merely undeclared. Thirteen of those tests are new, covering the class assignment per widget family, the exclusivity between families, idempotence, preservation of unrelated classes, the error markup, and that the three previously affected pages still render without any legacy class.

## Before and after

Captured on the demo project at the same viewport, with the same fixtures on both sides. Two custom fields were added to the demo database to exercise the widgets the default fixtures do not cover: a `boolean` one and a `list` one.

### Public ticket form

Fields have no bottom margin before, so each label sits against the previous field's help text. Selects are indistinguishable from text inputs. After, spacing is even and selects carry their chevron.

before-public-submit

after-public-submit

### Public ticket form, submitted empty

Before, the error text is red but no field is marked, and on Description and E-Mail the message runs into the help text on the same line. After, invalid fields get a red border and an alert icon, and each message sits on its own line.

before-public-submit-errors

after-public-submit-errors

### User settings

The clearest case. Before, the checkboxes are unstyled browser defaults with the label and help text running together on one overflowing line, and the submit button is unstyled. After, the checkboxes are Bootstrap 5 controls, help text sits under its label, and the button is styled.

before-user-settings

after-user-settings

### Edit ticket

Note the `Needs a callback?` custom field. Before, it is a full width empty rectangle with no visible checkbox at all, which is the `form-control` bug described above. After, it is a 16px checkbox with its label and help text, and `Business impact` is a proper `form-select`.

before-ticket-edit

after-ticket-edit

## Note for reviewers

`FormControlDeleteFormSet.deletion_widget` in `forms.py` still declares `form-control` on a checkbox. It is rendered by the checklist templates rather than through `bootstrap5_form`, so this branch does not reach it and I have left it for a separate change.

The codebase contains no `RadioSelect`, so radio groups never render on any of these pages today. The tag handles them, and the tests cover them, but nobody should go looking for a visual difference there.

### DIFF
--- a/demodesk/config/settings.py
+++ b/demodesk/config/settings.py
@@ -47,7 +47,6 @@ INSTALLED_APPS = [
     "django.contrib.staticfiles",
     "django.contrib.sites",
     "django.contrib.humanize",
-    "bootstrap4form",
     "helpdesk",  # This is us!
     "rest_framework",  # required for the API
 ]
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -74,7 +74,6 @@ errors with trying to create User settings.
             'django.contrib.sites',  # Required for determining domain url for use in emails
             'django.contrib.admin',  # Required for helpdesk admin/maintenance
             'django.contrib.humanize',  # Required for elapsed time formatting
-            'bootstrap4form', # Required for nicer formatting of forms with the default templates
             'rest_framework',  # required for the API
             'django_cleanup.apps.CleanupConfig',  # Remove this if you do NOT want to delete files on the file system when the associated record is deleted in the database
             'helpdesk',  # This is us!
--- a/docs/upgrade.rst
+++ b/docs/upgrade.rst
@@ -10,6 +10,26 @@ Please consult the Installation instructions for general instructions and tips.
 The tips below are based on modifications of the original installation instructions.
 
 
+Form rendering no longer needs django-bootstrap4-form
+-----------------------------------------------------
+
+``django-bootstrap4-form`` has been removed. It emitted Bootstrap 3 markup
+(``form-group``, ``control-label``, ``has-error``, ``<div class="checkbox">``),
+none of which exists in the Bootstrap 5 stylesheet this project ships, so the
+affected fields had lost their spacing and their checkboxes and radios were
+unstyled. Forms are now rendered by a template inside django-helpdesk itself.
+
+Remove ``'bootstrap4form'`` from your ``INSTALLED_APPS``. Nothing else is
+required, and the package can be uninstalled unless something else in your
+project uses it.
+
+If you have overridden ``user_settings.html``, ``edit_ticket.html`` or
+``public_create_ticket_base.html``, replace ``{{ form|bootstrap4form }}`` with::
+
+    {% load bootstrap5_form %}
+    {% bootstrap5_form form %}
+
+
 Optional integrations are no longer installed by default
 --------------------------------------------------------
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,16 +19,10 @@ packages = [
 ]
 dependencies = [
     "django>=4.2",
-    "django-bootstrap4-form",
     # Kept as a hard dependency on purpose: it activates automatically when listed
     # in INSTALLED_APPS, as docs/install.rst suggests, so a copy that follows our
     # own instructions would refuse to start without it.
     "django-cleanup",
-    # django-bootstrap4-form imports `packaging` in its meta.py but only
-    # declares `django`, so it cannot load without this. Until then it reached
-    # us transitively through celery's dependency chain, which stopped being
-    # installed by default here. Declared explicitly rather than left to luck.
-    "packaging",
     "email-reply-parser",
     "markdown",
     "beautifulsoup4",
--- a/quicktest.py
+++ b/quicktest.py
@@ -40,7 +40,6 @@ class QuickDjangoTest:
         "django.contrib.sessions",
         "django.contrib.sites",
         "django.contrib.staticfiles",
-        "bootstrap4form",
         #  The following commented apps are optional,
         #  related to teams functionalities
         # 'account',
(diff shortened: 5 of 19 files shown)


uhurusurfa (approved):
Seems to work as expected on my local.

PR #1378, opened 2026-08-10
docs: point reporters at private vulnerability reporting

Benbb96:
Follow-up to #1376, now that @uhurusurfa has enabled private vulnerability reporting on the repository.

## Reporting channel

The policy now sends reporters to [private vulnerability reporting](https://github.com/django-helpdesk/django-helpdesk/security/advisories/new) rather than to a personal email address. Both recent reports arrived by email, which meant they depended on one person reading their inbox, and the exchange lived in a mail thread while the advisory lived on GitHub.

The form removes both problems. It opens a draft advisory only maintainers can see, with the reporter already a collaborator on it, which is exactly what step 1 of the handling section previously told us to set up by hand. `<uhurusurfa@gmail.com>` stays as an explicit fallback for anyone without a GitHub account.

Step 1 is reworded to match: a report that came through the form has already created the draft, so there is nothing to open, only to take over and credit.

## Formatting

Also settles the file on **one sentence per line**, which @uhurusurfa agreed to in #1376. Before this the file was mixed: the section #1376 added was hard wrapped near 80 columns, while the older paragraphs each ran as one long line.

@DavidVadnais, you were asked for your view on the convention and had not answered yet, so say so if you would rather we did not. It is easy to revert in isolation, the formatting is the only thing in this pull request that touches lines that were already there.

Rendered output is unchanged: Markdown joins consecutive lines into a single paragraph.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -6,62 +6,58 @@ We only support the latest django-helpdesk version.
 
 ## Reporting a Vulnerability
 
-If you believe you have discovered a bug impacting security in a supported version, please DO NOT file a Issue / Bug Report for it publicly.
+If you believe you have discovered a bug impacting security in a supported version, please DO NOT file an Issue or Bug Report for it publicly.
 
-Instead, please send details to <uhurusurfa@gmail.com>. Please be sure to include "django-helpdesk security issue" in the subject line for fastest response.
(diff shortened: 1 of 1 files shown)


DavidVadnais (approved):
I checked the markdown and it renders how I would expect.  I also read the contents and the order/logic makes sense to me.

uhurusurfa (approved):
Perfect

PR #1379, opened 2026-08-11
feat: color code follow-ups by visibility and direction

### LINKED ISSUE #204: Followup list: different visual weight for private/public followups
Private followups often get used to add notes about a ticket that aren't shown to the submitter. 

For staff members it can be difficult to tell at a glance which items are private or public. By highlighting this (eg background shading, text colour, or otherwise) confusion could be avoided. 

Similar concept: show staff / submitter replies like a modern IM client, with left / right justification to help visualise the conversation.

DavidVadnais:
Closes #204.

A ticket's follow-up list mixes three kinds of entry: private staff notes, replies sent out to the submitter, and messages received from the submitter, and rendered all of them identically apart from a small 'Private' caption. Staff scanning a long ticket had to read each entry to work out who could see it, and the failure mode runs in the expensive direction: a note written as if it were internal that the submitter can in fact read.


DavidVadnais:
image

I also list who received each followup on my use case but that is pretty heavy over the long run.

Benbb96 (commented):
Nice work, and doubling the color with a text badge is the right call. One thing to sort out before merge.

## Public ticket view follow-ups are classified as outbound

The docs say `followup-item-inbound` covers a follow-up received "either by e-mail or through the public ticket view", but the second half is not true. `is_from_submitter` falls back to `self.user`, and neither `PublicTicketForm.save()` nor `update_ticket()` records a user when the author is not staff, so both anonymous paths land on `user=None` with no `message_id`. Running the real views:

```
title='Ticket Opened Via Web'      user=None message_id=None public=True -> direction='outbound'
title='Resolved ticket and Closed' user=None message_id=None public=True -> direction='outbound'
```

The first is the opening follow-up of **every anonymously submitted ticket**, so it is the common case. Nothing leaks, private stays internal, but the green "To submitter" badge is wrong there.

Two ways out, your call:

1. **Narrow the docs**: drop the public ticket view mention and treat e-mail as the only inbound signal. No migration, mergeable as is.
2. **Store provenance** explicitly at the creation sites. Cleaner, but needs a migration and is probably a separate PR.

Either way the `is_from_submitter` docstring contradicts itself: it says a follow-up with no user "was either received from the submitter or generated by the helpdesk itself", then returns `False` right below.

## Suggestion: make it visible in the demo

The four follow-ups in `demodesk/fixtures/demo.json` date back to 2017, all public and all attributed to the admin, so `make rundemo` shows four green rows and never the other two states. Turning follow-up 4 into a received e-mail and adding an internal note covers all three:

```json
{"model": "helpdesk.followup", "pk": 4, "fields": {"ticket": 3, "date": "2017-03-20T05:28:28.458Z", "title": "E-Mail Received from customer@example.com", "comment": "An image attachment goes here!", "public": true, "user": null, "message_id": "<demo-1@example.com>", "new_status": null}}
{"model": "helpdesk.followup", "pk": 5, "fields": {"ticket": 3, "date": "2017-03-20T05:31:00.000Z", "title": "Internal note", "comment": "Checked with the warehouse, the parcel never left.", "public": false, "user": 1, "new_status": null}}
```

## Not yours to fix

Your change surfaced two existing bugs, both coming as separate PRs: the edit view rebuilds a follow-up from scratch and drops its `message_id` (flipping inbound to outbound, and deleting its `TicketChange` rows on the way), and the three dots menu renders even when it holds no entries.

DavidVadnais:
Okay I updated the docs and here is an image of ticket 3 after running `make rundemo`

image

DavidVadnais:
We can go back through and do the bigger change with the migration later.

uhurusurfa (approved):
Works for me

PR #1381, opened 2026-08-11
WIP: docs: auto pull in liscense to docs/license.rst

### LINKED ISSUE #1380: 3rd Party Liscenses need updating and perhaps automation
**Describe the bug**
Looking through our docs I found this page https://django-helpdesk.readthedocs.io/en/v2.3.3/license.html in pr #1379 I made our license auto update and I fixed the footer.  Looking at this made me think of our yarn implementation a few years back. 

### Out of date vs. what yarn installed

all but Raphael

### Listed but no longer shipped

- jqPlot - no jqPlot file exists anywhere in the repo
- akismet.py - nothing vendored. Akismet is now an optional pip extra, so it isn't distributed with the package and doesn't belong in a "distributed with django-helpdesk" list.

### Shipped but missing from the page

- Timeline 3 - present at vendor/timeline3 (101 git-tracked files) and listed as entry 13 in `LICENSE.3RDPARTY`, but absent from the docs page entirely.

Not yarn-managed, so not checkable this way

SB Admin, Morris.js, and Flot are git-tracked under vendor/ (or static/helpdesk/{js,css,scss}/ for SB Admin) rather than pulled by yarn

Worth noting the docs page says SB Admin v2.0, 2013-2016 while `LICENSE.3RDPARTY` says v5.0.2, 2013-2018 - they disagree with each other.

---

`LICENSE.3RDPARTY` is stale too, so switching the docs page to .. include:: it would fix the drift but inherit the wrong data. Its version labels are jQuery 3.3.1 (actual 3.7.1), DataTables 1.10.18 (actual 2.3.4), Bootstrap 4.2.1 (actual 5.3.8), Font Awesome 5.6.3 (actual 6.7.2). So the include is still worth doing - one wrong copy beats two - but `LICENSE.3RDPARTY` needs its own correction pass first.

**To Reproduce**
1. update a yarn package that happens to inherit a new license
2. our documentation is now out of date

**Expected behavior**
Our documentation to align with the licenses of the libraries we use

DavidVadnais:
Fixes #1380


DavidVadnais:
https://github.com/jquery/jquery/blob/main/LICENSE.txt
https://github.com/jquery/jquery-ui/blob/main/LICENSE.txt
https://datatables.net/license/mit
https://www.jsdelivr.com/package/npm/timelinejs3

DavidVadnais:
I think we can drop  drop akismet.py and jqPlot.

uhurusurfa (approved):
Looks good.

PR #1382, opened 2026-08-11
fix: hide the follow-up menu when it holds no actions

Benbb96:
The three dots menu on each follow-up in the staff ticket view renders unconditionally, but both of its entries are gated. Edit needs the follow-up to have an author and the viewer to be that author or a superuser. Delete needs a superuser plus `HELPDESK_SHOW_DELETE_BUTTON_SUPERUSER_FOLLOW_UP`, which is off by default.

On a default install that leaves an empty menu open in two common situations: any staff member looking at a colleague's follow-up, and anyone at all looking at a follow-up with no author. The second covers more than it sounds, since `email.py` never records a user on a received reply, the public form records none for an anonymous submitter, and `update_ticket` drops the user when they are not staff. The opening follow-up of every publicly submitted ticket falls in that group.

Both permissions now live in template filters, so the decision to show the menu and the decision to show each entry cannot drift apart, and the menu is skipped when neither applies. Behaviour is otherwise unchanged: nobody gains or loses an action, the empty shell just stops being offered.

Worth noting separately, `followup_edit` itself has no ownership check beyond `helpdesk_staff_member_required` and `ticket_perm_check`, so hiding the Edit entry never protected anything. Whether an author-less follow-up should be editable is a policy question I have left alone here.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


src/helpdesk/templatetags/followup_permissions.py
```diff
@@ -0,0 +1,38 @@
...
+
+    A follow-up with no author was not written by a staff member -- it was
+    received by e-mail, submitted from the public form or raised by the
+    helpdesk itself -- and editing those is not offered to anyone.
+    """
+    if not helpdesk_settings.HELPDESK_SHOW_EDIT_BUTTON_FOLLOW_UP:
+        return False
+    if not followup.user:
+        return False
+    return user == followup.user or user.is_superuser
```
DavidVadnais, line 29: I think `or user.is_superuser` is never reached in this code

DavidVadnais (changes requested):

src/helpdesk/templatetags/followup_permissions.py
Benbb96: It is reached, and removing it changes behavior. The clause fires whenever the viewer is not the author, which is exactly the superuser looking at a colleague's follow-up. `test_superuser_is_offered_the_menu_on_a_colleagues_followup` covers that case, and dropping the clause turns it red:

```
FAIL: test_superuser_is_offered_the_menu_on_a_colleagues_followup
Ran 6 tests in 3.600s
FAILED (failures=1)
```

The condition is carried over verbatim from the template this filter replaces, `{% if request.user == followup.user or user.is_superuser %}`, so removing it would cost superusers the Edit entry on every follow-up but their own.

If it reads as dead code at a glance I am happy to make it more obvious, either by putting the broader test first as `user.is_superuser or user == followup.user`, or with a one line comment. Let me know which you prefer.

uhurusurfa: Ben is right.
If there is no followup user assigned to the followup then it does not get to that line.
If the assigned followup user is not the user currently viewing that record then it is only true if the user viewing the record is a super user.

uhurusurfa (approved):
LGTM

uhurusurfa (approved):

DavidVadnais:
I just tested this as well. @Benbb96 is right.  LGTM.

DavidVadnais (approved):

PR #1384, opened 2026-08-12
Enhancement/speed up test suite

gurupratap-matharu:
This PR brings in massive speed improvements in running the test suite for helpdesk.

The main one line change was to use `PASSWORD_HASHERS = (django.contrib.auth.hashers.MD5PasswordHasher,)` in test settings.

I learnt this from [this implementation in another project](https://github.com/healthchecks/healthchecks/blob/49653c350cddc47fc00a471bd1b08b5771a7967c/hc/api/tests/__init__.py#L12).

I also moved test specific settings from `quicktest.py` -> `tests/test_settings.py` for better readability and took inspiration from [this example in the official documentation](https://docs.djangoproject.com/en/6.1/topics/testing/advanced/#id3) on how to test reusable apps.

I use an old desktop for coding where tests ran in ~130s which as painful and after these changes its ~9s.
But please do verify this from you side as well in every way possible.

User can still pass specific test to run, say for example
`python quicktest.py tests.test_my_view -v 2` That remains untouched.

Thank you


tests/test_settings.py
```diff
@@ -0,0 +1,76 @@
...
+INSTALLED_APPS = (
+    "django.contrib.admin",
+    "django.contrib.auth",
+    "django.contrib.contenttypes",
+    "django.contrib.humanize",
+    "django.contrib.messages",
+    "django.contrib.sessions",
+    "django.contrib.sites",
+    "django.contrib.staticfiles",
+    "bootstrap4form",
```
DavidVadnais:
```suggestion
```

One of the recent PRs made this no longer installed.

quicktest.py
```diff
@@ -1,145 +1,36 @@
...
-    Example usage:
-
-        $ python quicktest.py test1 test2
-
+    Parse which tests the user wants to run with what verbosity
+    and run them via a custom test runner.
     """
     parser = argparse.ArgumentParser(usage="[args]", description="Run Django tests.")
-    parser.add_argument("tests", nargs="*", type=str)
+    parser.add_argument("tests", nargs="*", type=str, default=".")
```
DavidVadnais:
```suggestion
    parser.add_argument("tests", nargs="*", type=str, default=["."])
```

I think with `nargs="*"`, argparse assigns the default **unwrapped** - `args.tests` ends up as the string `"."`, not a list.

tests/test_settings.py
```diff
@@ -0,0 +1,75 @@
+import os
+from pathlib import Path
+
+BASE_DIR = Path(__file__).resolve().parent.parent
+
+DEBUG = True
+DATABASES = {
+    "default": {
+        "ENGINE": "django.db.backends.sqlite3",
+        "NAME": Path(BASE_DIR / "test.db"),
```
DavidVadnais:
```suggestion
        "NAME": BASE_DIR / "test.db",
```

already a path so i think this noops

DavidVadnais (changes requested):
`.github/workflows/test.yml:76` still sets

```
    env:
      DJANGO_SETTINGS_MODULE: helpdesk.settings
```

This can be removed

gurupratap-matharu:
All done and ready. ci/cd checks are running faster I suppose.

DavidVadnais (approved):
I’m happy with this. I’ll give everyone else a day or two to review.

Thank you for all your work!

PR #1387, opened 2026-08-14
hide the queue picker when only one queue exists, across create/edit ticket forms

### LINKED ISSUE #289: Hide queue selection when there's only a single queue

DavidVadnais:
fix #289

### DIFF
--- a/src/helpdesk/forms.py
+++ b/src/helpdesk/forms.py
@@ -147,6 +147,14 @@ class EditTicketForm(CustomFieldMixin, forms.ModelForm):
             "This ticket is merged into the selected ticket."
         )
 
+        # If there is only a single queue configured, there is nothing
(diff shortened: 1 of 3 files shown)


DavidVadnais:
## with 2 queues
image

image


## with 1 queue
image

image

DavidVadnais:
Uggh I see queue is still in ticket details.

image

Benbb96 (approved):
LGTM

Benbb96:
> Uggh I see queue is still in ticket details.

IMO, that should stay because the name of the queue could bring information. But that's debatable!

DavidVadnais:
I’ll leave it both because of what you mentioned and I don’t see a way to remove it that would look good.

PR #1388, opened 2026-08-15
Update supported key software versions of Django and Python

uhurusurfa:
Removes no longer supported versions of Python and Django from the documentation and test matrices.

Python 3.10 is 2 months from EOL but left it in for now.
Python 3.14 is about to be released within the next few months so have included it in advance.

Also updates the Django-Helpdesk version allowing us to release now or whenever the team feel is a good point in time given the flurry of development we have seen in the last few weeks.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -25,7 +25,7 @@ jobs:
           - python-version: "3.11"
             django-version: "5.2"
         # Define the general matrix for Python with Django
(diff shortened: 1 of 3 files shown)


DavidVadnais (approved):
LGTM

uhurusurfa:
@DavidVadnais - if you have no more PR's that you are currently working on then feel free to make a release.

PR #1389, opened 2026-08-19
DevOps: Setup Dependabot for dependency management

DavidVadnais:
I started using Dependabot in some of my other repos and it seems pretty good. Let me know if you don't like this technology, and separately, flag any gotchas I might be blind to as a new user."

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,78 @@
+version: 2
+
+updates:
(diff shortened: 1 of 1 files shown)


uhurusurfa (approved):
I am fine with a more detailed definition of the use of Dependabot.
Note that it is enabled in this repo and we do get PR's for security issues with dependent libraries.

Benbb96 (approved):

PR #1390, opened 2026-08-19
Feature: Followups now record who was emailed with the folowup

### LINKED ISSUE #1385: Followups could list the recipeints of the followup
Sometimes you want to know who was contacted for a followup

A line with who recieved which followups

Image

**Additional context**
I think this will take a migration.

DavidVadnais:
Fixes: #1385

Adds a list of recipients to each followups.  Old followup dont generate this field and followups with no recipients generate a "No one" message. This is only visible to staff to prevent unwanted emails getting out to external submitters/

image


DavidVadnais:
Do you think I should add a feature toggle to this? It could get messy with large numbers of recipients.

uhurusurfa (approved):
Makes sese. In  regards your question about it getting messy, I doubt that there should ever be a large number of recipients for a given followup - just would not make sense to get too many people involved and if there is a group of people that  are supposed to be involved the sensible way to do it is to use group emails.
If this does become an issue we can later make the it a configurable option to either do what you have programmed or display a link in the followup to pop a list of recipients instead of showing names directly.

PR #1406, opened 2026-08-20
fix submit ticket failure with single queue

### LINKED ISSUE #1405: with a single queue you can no longer create a ticket.
**Describe the bug**
with a single queue you can no longer create a ticket.

**To Reproduce**
1. make rundemo
2. Delete one queue
3. try and create a ticket

**Additional context**
almost certainly from 
https://github.com/django-helpdesk/django-helpdesk/pull/1387

DavidVadnais:
fixes #1405

### DIFF
--- a/src/helpdesk/forms.py
+++ b/src/helpdesk/forms.py
@@ -492,7 +492,6 @@ class TicketForm(AbstractTicketForm):
             # nothing meaningful to choose, so hide the field and default
             # to it.
             if len(queue_choices) == 1:
-                self.fields["queue"].initial = queue_choices[0][0]
                 self.initial["queue"] = queue_choices[0][0]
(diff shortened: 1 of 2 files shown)


src/helpdesk/forms.py
```diff
@@ -493,6 +493,7 @@ def __init__(
             # to it.
             if len(queue_choices) == 1:
                 self.fields["queue"].initial = queue_choices[0][0]
+                self.initial["queue"] = queue_choices[0][0]
```
Benbb96, line 495: So, is the line above still useful?

DavidVadnais: your right

Benbb96 (approved):

PR #1407, opened 2026-08-21
Enhancement/improve template structure

### LINKED ISSUE #598: Template structure
At present, the template structure seems very un-Django-like. There are two base templates base.html and public_base.html, both of which {% include navigation.html %}. The Django template language doesn't allow one to define blocks in an include file. (see Stack Overflow 3983872) Pasting a copy of navigation.html into both base files is not attractive either.

I've rough-and-ready verified that it is possible to refactor this as a single base file basebase.html (much like public_base.html, with added, mostly empty, blocks), and a navigation.html that extends basebase.html, and then new versions of public_base.html and base.html which extend navigation.html and inject differences by defining or extending basebase- and navigation-defined blocks. My intent is to make it easier for the end users to customize appearance of the helpdesk and local content of the navigation bar, without having to make changes to the underlying base files that will be awkward to propagate across upgrades.

These changes will of course be awkward to propagate if there are already other plans afoot for these templates in future releases. Are there?  Is there any interest in this for the project?

gurupratap-matharu:
Partially fixes #598

The issue talks about removing multiple base templates. Unknowingly partial progress was made on this in another PR on KB list detail one. 
While diving in another `help_base.html` was found which I'm not sure why it was given its own base or formatting?

This PR removes it and reformats the UI of the contextual reference template. I've kept it simple and edited the wordings a bit. Do proof read it. Here is a screenshot

contextual-reference-revamped

Simple unit tests are added.
The issue talk about many other things like jquery removal or vue.js implementation. That's why a partial fix ;-)

Thank you


Benbb96 (commented):
This looks good to me. Is it possible to take the opportunity to add translation blocks where it's missing on this page?

gurupratap-matharu:
Done. Does this look good to you?

src/helpdesk/templates/helpdesk/help_context.html
```diff
@@ -31,24 +33,24 @@ <h4 class="alert-heading" id="usage">{% translate "Inserting Fields" %}</h4>
...
           <ul class="list-group list-group-flush">
             <li class="list-group-item" id="queue_title">
               <p class="fw-medium mb-0">queue.title</p>
-              The display title of the queue
+              {% translate "The display title of the queue" %}
             </li>
             <li class="list-group-item" id="queue_slug">
               <p class="fw-medium mb-0">queue.slug</p>
-              The "slug" of the queue, a single-word lowercase identifier used in URL"s and e-mail subjects.
+              {% translate "The 'slug' of the queue, a single-word lowercase identifier used in urls and e-mail subjects." %}
```
Benbb96: Maybe you can preserve `"slug"` wrapped by double quotes?

`{% translate "The \"slug\" of the queue, a single-word lowercase identifier used in urls and e-mail subjects." %}`

src/helpdesk/templates/helpdesk/help_context.html
```diff
@@ -57,84 +59,85 @@ <h5 class="card-title" id="queue">{% translate "Queue Fields" %}</h5>
...
             </li>
             <li class="list-group-item" id="ticket_get_status_display">
               <p class="fw-medium mb-0">ticket.get_status_display</p>
-              Status of this ticket (Text-based)
+              {% translate "Status of this ticket (Text-based)" %}
             </li>
             <li class="list-group-item" id="ticket_get_status">
               <p class="fw-medium mb-0">ticket.get_status</p>
-              Status of this ticket (Text-based, includes "On Hold" if required)
+              {% translate "Status of this ticket (Text-based, includes 'On Hold' if required)" %}
```
Benbb96: Here too

src/helpdesk/templates/helpdesk/help_context.html
```diff
@@ -57,84 +59,85 @@ <h5 class="card-title" id="queue">{% translate "Queue Fields" %}</h5>
...
             </li>
             <li class="list-group-item" id="ticket_ticket">
               <p class="fw-medium mb-0">ticket.ticket</p>
-              The ticket ID in form "[queue_slug-id]", used in e-mail subjects.
+              {% translate "The ticket ID in form '[queue_slug-id]', used in e-mail subjects." %}
             </li>
             <li class="list-group-item" id="ticket_ticket_for_url">
               <p class="fw-medium mb-0">ticket.ticket_for_url</p>
-              The ticket ID in form "queue_slug-id", used in URLs
+              {% translate "The ticket ID in form 'queue_slug-id', used in URLs" %}
```
Benbb96: And also here. I was thinking that it could be wrapped by a `<code>` tag but it would need to be out of the transaltion, or to use the `|safe` filter.
Maybe we should keep it simple.

Benbb96 (changes requested):
Yes thanks for the change. Just a small tweaks to keep the double quotes like before.

gurupratap-matharu:
These are some nice suggestions.

Double quotes are now being preserved. 
Rather than escaping them I'm using single quotes on the outer string for simplicity.

Thank you

src/helpdesk/templates/helpdesk/help_context.html
gurupratap-matharu: Simply added a code tag outside the translate tag

Benbb96 (approved):
Perfect, thank you 🙂️

DavidVadnais:
Hmm I like rebase and merge with depndabot but I don’t like how at first glance this looks like i coauthored these commits when I didn’t. 

Two engineers I really trust think rebase and merge is always better so I’ll dive deeper into this later.

PR #1411, opened 2026-08-28
Tests/test ticket model

gurupratap-matharu:
This PR brings in unit test improvements for the central Ticket model of the project.

Intent has been made to introduce basic type hints on the Ticket model and its methods.
There is slight refactor on model methods to newer syntax or for readability
One new migration is produced due to changes on created, modified field
For property field declarations I tend towards the modern @property decorator for brevity.

There is effort to check for edge cases but it would be great if you can review and suggest 
other scenarios which can be tested.

Thank you.


src/helpdesk/models.py
```diff
@@ -798,175 +796,112 @@ def _get_allowed_status_flow(self):
...
+
+        if ticketcc:
+            ticketcc.ticket = self
+            ticketcc.save(update_fields=["ticket"])
+        elif user:
+            ticketcc, _ = self.ticketcc_set.get_or_create(user=user)
+        else:
+            ticketcc, _ = self.ticketcc_set.get_or_create(email=new_email)
+
+        return ticketcc
```
uhurusurfa, line 892: You could just remove this since nothing expects there to be a return value from this method.

uhurusurfa (dismissed):
The cleanup does make it more readable.
Works for me.

src/helpdesk/models.py
gurupratap-matharu: Removing this breaks an older test. [This one](https://github.com/django-helpdesk/django-helpdesk/blob/95c153bf23ac92a1cdf3c1abf3ce72a6898ea4fc/tests/test_ticket_lookup.py#L73)

If its no harm I'll prefer to keep a return value.

gurupratap-matharu:
@uhurusurfa @DavidVadnais 

Need you recommendation here...

I've slightly tweaked and moved the coverage configuration from .coveragerc to pyproject.toml just to keep all configs in one place. `.coveragerc` is deleted

Coverage version has also been bumped up in the test group using UV

As per the coverage py docs

 Coverage.py will read from “pyproject.toml” if TOML support is available, either because you are running on Python 3.11 or later, or because you installed with the toml extra (pip install coverage[toml]).

In the [github workflow on this line](https://github.com/django-helpdesk/django-helpdesk/blob/95c153bf23ac92a1cdf3c1abf3ce72a6898ea4fc/.github/workflows/test.yml#L71) 

Should we do `pip install coverage[toml]`?

uhurusurfa:
With EOL for Python 3.10 just weeks away I think we just drop support for it and that will resolve the issue.

DavidVadnais:
Im fine with dropping python 3.10

gurupratap-matharu:
Great. 

The python 3.10, Django 5.2 combination has been removed from the tests workflow.
Although the tests pass this has resulted in coveralls alerting about coverage decline.

What's the best work around to this?

uhurusurfa (changes requested):
In order to get past the issue with coveralls failing, I think setting `"time_spent"` to a non empty value in the `tests/test_followup_edit.py ` test set should trigger an additional test path that will fully exercise the format_time_spent(time_spent) method in `helpdesk/lib.py `and will satisfy the minimum code coverage setting.

src/helpdesk/models.py
```diff
@@ -738,46 +754,28 @@ def _get_priority_css_class(self):
...
-        """
-        held_msg = ""
-        if self.on_hold:
-            held_msg = _(" - On Hold")
-        dep_msg = ""
-        if not self.can_be_resolved:
-            dep_msg = _(" - Open dependencies")
-        return f"{self.get_status_display()}{held_msg}{dep_msg}"
+        held_msg = _(" - On Hold") if self.on_hold else ""
+        dep_msg = "" if self.can_be_resolved else _(" - Open dependencies")
```
Benbb96: IMO formatting shouldn't be in the translation string

```suggestion
        held_msg = f" - {_('On Hold')}" if self.on_hold else ""
        dep_msg = "" if self.can_be_resolved else f" - {_('Open dependencies')}"
```
But that means locale files need to be updated.

src/helpdesk/models.py
```diff
@@ -798,175 +796,112 @@ def _get_allowed_status_flow(self):
...
-            protocol,
-            site.domain,
-            reverse("helpdesk:public_view"),
+        return "{}{}?ticket={}&email={}&key={}".format(
+            base_url,
+            reverse_lazy("helpdesk:public_view"),
             self.ticket_for_url,
             self.submitter_email,
             self.secret_key,
         )
```
Benbb96: why not replacing with an f-string here ?

src/helpdesk/models.py
```diff
@@ -631,6 +629,27 @@ class Ticket(models.Model):
...
+
+    def __str__(self) -> str:
+        return f"{self.id} {self.title}"
+
+    def get_absolute_url(self) -> str:
+        return reverse_lazy("helpdesk:view", args=(self.id,))
+
+    def save(self, *args, **kwargs) -> None:
+        if not self.priority:
+            self.priority = 3
```
Benbb96: Do we need to keep this condition? Priority already has a default set to 3.

Benbb96 (commented):
Great modernization on models.py!

I think you need to add a test that produces an `ImproperlyConfigured` on `Site.objects.get_current()` (line 304 of lib.py) in order to cover that branch and have it to pass the CI.

src/helpdesk/models.py
gurupratap-matharu: Its not clear to me why you use the f-string when there is no variable to parse.

src/helpdesk/models.py
gurupratap-matharu: Had my doubts. Now removed it,

gurupratap-matharu:
Well all of this is great feedback. 

I think the Ticket model and tests are in much better shape now.
Coveralls seems to be happy too.

Handing off to you guys.

src/helpdesk/models.py
uhurusurfa: I agree that the formatting should not be part of the translation but since this is simply a case of concatenating 2 strings there is no need to use the more expensive string interpolation mechanism which can be useful for concatenating different data types to a string and the plus operator would suffice in this use case.

gurupratap-matharu: @Benbb96  I now understand your concern. I'm sending this change

```   
@property
def get_status(self) -> str:
    """Displays the ticket status, with an "On Hold" message if needed."""

    held_msg, dep_msg = _("On Hold"), _("Open dependencies")
    status = self.get_status_display()

    if self.on_hold:
        status = f"{status} - {held_msg}"

    if not self.can_be_resolved:
        status = f"{status} - {dep_msg}"

    return status
```

Hope this works for you

Benbb96: Yes that's far more readable, thanks

Benbb96 (approved):
All good to me

DavidVadnais (approved):
Looks good to me

uhurusurfa (approved):

PR #1415, opened 2026-08-31
docs: give settings and command options their own directives

Benbb96:
Stacked on #1414, so the diff to review is the second commit only. The base needs to move to `main` once #1414 lands.

The reference pages were plain definition lists, so no individual setting had an anchor. There was no way to link anyone to `HELPDESK_ENABLE_ATTACHMENTS`, and a setting name mentioned in prose was just literal text that nothing checked.

## A `setting` object type

`docs/conf.py` registers one through `add_object_type`, which is the built-in shortcut for what Django's own documentation does with a full custom domain:

```python
def setup(app):
    app.add_object_type(
        "setting",
        "setting",
        objname="django-helpdesk setting",
        indextemplate="pair: %s; setting",
    )
```

The Python domain's `py:data` would have worked too, and was the first thing I tried. I moved away from it because these are not attributes of any module we ship. They live in the deployer's own `settings.py`, and putting them in the `py` inventory would mix them into the Python index the day we add autodoc for the models.

The 81 entries in `docs/settings.rst` become `setting` directives, each with an anchor, a permalink and an index entry, and `` :setting:`NAME` `` now resolves to them. Five mentions in prose became cross-references. The default value moves out of the term and into the body, so it no longer has to be squeezed into the heading:

```rst
.. setting:: HELPDESK_KANBAN_ENABLED

   *Default:* ``True``

   Show the Kanban board?
```

You can see the result here : https://django-helpdesk--1415.org.readthedocs.build/en/1415/settings.html

The tables in `docs/standalone.rst` were deliberately left as they are. They document the Docker image's environment variables, which are a separate namespace, and two of the names listed there are not settings at all.

## Management commands

The three commands the configuration walkthrough tells you to run are now documented with `program` and `option` directives, their flags taken from each command's `add_arguments` rather than from the surrounding prose. `--debug_to_stdout` and `--notify-only` are now cross-references from the walkthrough.

The `--debug_to_stdout` example was wrapped in `**` inside a literal block, so the asterisks were rendering verbatim. Fixed.

## One setting was in the wrong section

`HELPDESK_ALWAYS_SAVE_INCOMING_EMAIL_MESSAGE` was listed under Discontinued Settings, but `src/helpdesk/settings.py` still reads it and `src/helpdesk/email.py` still acts on it, gated on `HELPDESK_ENABLE_ATTACHMENTS` exactly as its description claims. It moves back up with the other attachment settings. It is the only one of the seventeen names in that section that still exists in the code, the other sixteen are genuinely gone.

## Also

Single backticks that were rendering as italics rather than as literals, in `api.rst`, `install.rst`, `standalone.rst` and `settings.rst`. I chose this over setting `default_role = "literal"`, to avoid changing a Sphinx default for six occurrences.

The `epub_exclude_files` change is carried here as well: the EPUB builder never references the logo files, and with `fail_on_warning: true` the two unknown-MIME-type warnings were failing the Read the Docs build.

## Verification

`sphinx-build -W` clean for `html`, `epub` and `latex`, since `.readthedocs.yaml` sets `formats: all` with `fail_on_warning: true`.

Draft for now: I would like to see it rendered on Read the Docs before calling it done.
## Deliberately not here

`HELPDESK_PUBLIC_ENABLED` is set by the standalone and demo configs and promised by `docs/standalone.rst`, but nothing in django-helpdesk reads it. Documenting it in `docs/settings.rst` would promise the dead switch a second time, so the question is in #1416 instead.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -84,7 +84,7 @@ Here is an example with form-data (curl default) ::
 
 Accessing the endpoint ``/api/followups/`` with a **POST** request will let you create a new followup on a ticket.
 
-This time, you can attach multiple files thanks to the `attachments` field. Here is an example ::
+This time, you can attach multiple files thanks to the ``attachments`` field. Here is an example ::
 
     curl --location --request POST 'http://127.0.0.1:8000/api/followups/' \
     --header 'Authorization: Basic YWRtaW46YWRtaW4=' \
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -193,6 +193,17 @@ html_split_index = False
 htmlhelp_basename = "django-helpdeskdoc"
 
 
+# -- Options for EPUB output ---------------------------------------------
+
+# The EPUB builder has no sidebar and no favicon, so it never references the
+# logo files. Keep them out of the archive: their formats are not part of the
+# EPUB media types and Sphinx warns about every file it cannot map.
+epub_exclude_files = [
+    "_static/logo.avif",
+    "_static/logo.ico",
+]
+
+
 # -- Options for LaTeX output --------------------------------------------
 
 # The paper size ('letter' or 'a4').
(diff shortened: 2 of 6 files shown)


uhurusurfa (approved):
LGTM 👍

uhurusurfa (approved):
👍

uhurusurfa:
Works fine:
  https://django-helpdesk--1415.org.readthedocs.build/en/1415/

PR #1420, opened 2026-09-03
fix: remove HELPDESK_PUBLIC_ENABLED, which nothing reads

### LINKED ISSUE #1416: HELPDESK_PUBLIC_ENABLED is set and documented, but nothing reads it
Found while converting the settings reference to Sphinx directives in #1415. I went looking for `HELPDESK_PUBLIC_ENABLED` to document it and could not find anything that consumes it.

## What exists

* `standalone/config/settings.py:123` sets it from the `HELPDESK_PUBLIC_ENABLED` environment variable, under the comment `# Should the public web portal be enabled?`
* `demodesk/config/settings.py:114` sets it to `True`
* `docs/standalone.rst` lists it in the Public Portal Settings table, described as "Enable public web portal"

## What does not

Nothing reads it. It does not appear in `src/helpdesk/settings.py`, so it is not exposed as `helpdesk_settings.HELPDESK_PUBLIC_ENABLED` to the templates either, and grepping the tree turns up no other occurrence. The public portal is actually governed by `HELPDESK_VIEW_A_TICKET_PUBLIC` and `HELPDESK_SUBMIT_A_TICKET_PUBLIC`, which are both read and both documented.

So someone running the standalone image who sets `HELPDESK_PUBLIC_ENABLED=False` gets a fully working public portal, while the documentation tells them the portal is off. That is the part that makes this worth fixing rather than ignoring.

## The question

I did not document it in #1415, because documenting it in `docs/settings.rst` would be promising the same dead switch a second time. Three ways out, and I do not know which matches the original intent:

1. **Implement it in the core** as a master switch, so `HELPDESK_PUBLIC_ENABLED = False` implies both of the other two. It then gets documented in `docs/settings.rst` alongside them.
2. **Derive the other two from it in the standalone layer only**, so `standalone/config/settings.py` computes `HELPDESK_VIEW_A_TICKET_PUBLIC` and `HELPDESK_SUBMIT_A_TICKET_PUBLIC` from it. No core change, and the standalone documentation becomes true.
3. **Drop it**: remove the row from `docs/standalone.rst` and the two assignments, since they do nothing.

My preference is 1 if a master switch was the intent, 2 if it was only ever meant for the Docker image. Happy to open a PR for whichever.

Benbb96:
Closes #1416, following @uhurusurfa's answer there.

`HELPDESK_PUBLIC_ENABLED` was set by `standalone/config/settings.py` and `demodesk/config/settings.py`, and listed by `docs/standalone.rst` in the Public Portal Settings table as "Enable public web portal". Nothing in django-helpdesk ever read it. It is absent from `src/helpdesk/settings.py`, so it was not exposed to the templates either.

Someone running the standalone image with `HELPDESK_PUBLIC_ENABLED=False` got a fully working public portal while the documentation told them the portal was off. That is what made it worth removing rather than leaving alone.

The portal is governed by `HELPDESK_VIEW_A_TICKET_PUBLIC` and `HELPDESK_SUBMIT_A_TICKET_PUBLIC`, both read and both documented, so the switch had nothing left to do. Three lines removed, one per place it lived. The `# Should the public web portal be enabled?` comment stays, since it still describes the two remaining assignments accurately.

Nothing else referenced it: not `docker.env`, not `docker.env.template`, not the standalone README.

## Verification

* `ruff check` and `ruff format --check` clean on both settings files
* `sphinx-build -W` clean
* `manage.py check` clean against the demo settings

Found while converting the settings reference to Sphinx directives in #1415.


DavidVadnais (approved):
confirmed no live code ingests this

```
rg "HELPDESK_PUBLIC_ENABLED"
docs/standalone.rst
170:   ``HELPDESK_PUBLIC_ENABLED``, ``True``, "Enable public web portal"

standalone/config/settings.py
123:HELPDESK_PUBLIC_ENABLED = os.environ.get("HELPDESK_PUBLIC_ENABLED", "True") == "True"

demodesk/config/settings.py
114:HELPDESK_PUBLIC_ENABLED = True
```

PR #1422, opened 2026-09-04
fix: close the settings drift found in #1416

Benbb96:
The four quick fixes @uhurusurfa green-lit in #1416, one commit each.

## `HELPDESK_KANBAN_ENABLED` in the standalone image

`docs/standalone.rst` lists it under Feature Toggles, but `standalone/config/settings.py` never read the environment variable, so setting it in `docker.env` did nothing. The setting itself works: `src/helpdesk/settings.py:449` reads it, `urls.py:180` drops the kanban route, `navigation-sidebar.html:57` drops the sidebar link. The two neighbouring rows of the same table are wired, so this was an omission.

I had proposed removing the documented row instead. @DavidVadnais pointed out the setting is live, which turns the fix around: add the missing line.

## `HELPDESK_FULL_FIRST_MESSAGE_FROM_EMAIL` out of Discontinued Settings

`src/helpdesk/email.py:1136` still reads it. Same mistake I made with `HELPDESK_ALWAYS_SAVE_INCOMING_EMAIL_MESSAGE` in #1415, where I had only grepped `src/helpdesk/settings.py` and called it the sole survivor. Scanning the whole package, the remaining fifteen names in that section really are gone.

## Six settings nothing documented

Read by the code, absent from `docs/settings.rst` and from every other page under `docs/`, so the only way to find them was to read the source: `HELPDESK_UI_ENABLED`, `HELPDESK_USE_HTTPS_IN_EMAIL_LINK`, `HELPDESK_HTML_PREVIEW_ALLOW_REMOTE_IMAGES`, `HELPDESK_USE_CDN`, `HELPDESK_ATTACHMENT_DIR_PERMS`, `HELPDESK_QUERY_CLASS`.

Each goes in the section its neighbours are in. The two resolved at import time say so, since that is the kind of thing people discover by changing a value and seeing nothing happen.

## A test so this stops happening

Three static checks over files already in the repository:

1. every `HELPDESK_`/`QUEUE_` name an example configuration assigns is read by `src/helpdesk`. This is what `HELPDESK_PUBLIC_ENABLED` failed for years.
2. every environment variable the `docs/standalone.rst` tables promise is read from `os.environ` by `standalone/config/settings.py`. This is what `HELPDESK_KANBAN_ENABLED` failed. It inspects the environment reads rather than the file text, so a hardcoded value does not satisfy it.
3. every key of a `HELPDESK_DEFAULT_SETTINGS` dict is a field on `UserSettings`, since `DEFAULT_USER_SETTINGS.update()` merges whatever it is handed.

Each was verified to fail when its own regression is reintroduced. Check 3 turned up a third dead variable: `HELPDESK_PRESET_REPLIES` fed a `preset_replies` key that is not a `UserSettings` field, in both example configurations, plus `email_on_ticket_apichange` in the demo one. Removed, along with the documented row for the environment variable. The name `preset_replies` does exist elsewhere as a template context variable holding `PreSetReply` rows, which is unrelated and probably how it got there.

The checks are tripwires rather than proofs: a name appearing only in a comment counts as used, which is the right trade for catching a whole setting falling out of use. They skip when run against an installed distribution rather than a checkout.

## Verification

* full suite: 362 tests, OK
* `sphinx-build -W` clean
* `ruff check` and `ruff format --check` clean on the touched Python files

Closes the follow-up work in #1416.

### DIFF
--- a/demodesk/config/settings.py
+++ b/demodesk/config/settings.py
@@ -105,8 +105,6 @@ HELPDESK_DEFAULT_SETTINGS = {
     "email_on_ticket_assign": True,
     "email_on_ticket_change": True,
     "login_view_ticketlist": True,
-    "email_on_ticket_apichange": True,
-    "preset_replies": True,
     "tickets_per_page": 25,
 }
 
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -86,6 +86,20 @@ Settings related to attachments
 
    If you'd like to turn of filtering of helpdesk extension types you can set this to ``False``.
 
+.. setting:: HELPDESK_ATTACHMENT_DIR_PERMS
+
+   *Default:* ``"755"``
+
+   Permissions used when creating the directories that hold ticket attachments, as an octal string. Parsed with base 8, so ``"755"`` means ``rwxr-xr-x``.
+
+.. setting:: HELPDESK_HTML_PREVIEW_ALLOW_REMOTE_IMAGES
+
+   *Default:* ``False``
+
+   Allow the HTML preview of a ticket message to load images from remote hosts. Off by default, which means a tracking pixel in a customer's e-mail does not fire when a staff member previews it. Set it to ``True`` only if full rendering fidelity matters more than that.
+
+   The preview is served with a ``Content-Security-Policy`` that sandboxes the document and disables scripting regardless of this setting. This toggle only widens ``img-src``.
+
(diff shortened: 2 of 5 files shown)


uhurusurfa (approved):
Great job Ben - looks good to me.

DavidVadnais (approved):
lgtm

PR #1429, opened 2026-09-08
Standardize templates on translate/blocktranslate

dchaudhari7177:
Closes #1427.

Django 3.1 renamed `trans`/`blocktrans`/`endblocktrans` to `translate`/`blocktranslate`/`endblocktranslate`. The templates were already **partly** migrated, so both spellings were live — sometimes in the same file:

| tag | before | after |
| --- | ---: | ---: |
| `translate` / `trans` | 216 / **424** | 640 / 0 |
| `blocktranslate` / `blocktrans` | 10 / **26** | 36 / 0 |
| `endblocktranslate` / `endblocktrans` | 10 / **26** | 36 / 0 |

**476 replacements across 49 files.** Naming only — no msgid text changes, so no `.po` file needs regenerating. The old names remain aliases in Django, so this is not a behaviour change.

### On the `sed` in the issue

I used a tag-name **word boundary** rather than the plain `\{% trans\b` substitution suggested in the issue:

```python
re.compile(r'(\{%[-+]?\s*)(endblocktrans|blocktrans|trans)\b')
```

Two reasons: it leaves the already-migrated tags alone (a naive pass ordered wrongly can produce `translatelate`), and it covers the whitespace-control forms `{%-` and `{%+`.

### Verification

Django 5.2.17, which is the floor this project supports (`pyproject.toml` declares 5.2 / 6.0, CI matrix is 5.2 / 6.0 / 6.1) — comfortably above the 3.1 that introduced these names.

- Lexed **all 94 templates** with `django.template.base.Lexer`: **0 legacy tags remain**, and the only i18n tags now present are `translate`, `blocktranslate`, `endblocktranslate`.
- Confirmed all three are registered in `django.templatetags.i18n` on 5.2.
- Compiled **and rendered all 36 `blocktranslate` blocks** (including the `with`/`trimmed`/`plural` forms) standalone: **36 OK, 0 failed**.

I did not run the full test suite here — this is a template-token rename with no Python changes.

PR #1428, opened 2026-09-08
chore(i18n): standardize templates on translate/blocktranslate tags

yunaremaia:
## Summary

Standardizes all HTML templates on the modern `translate` / `blocktranslate` / `endblocktranslate` template tags, replacing the legacy `trans` / `blocktrans` / `endblocktrans` aliases.

Closes #1427

## What changed

- Applied the rename suggested in #1427 across every `*.html` template (49 files, 453 lines changed, pure rename — no logic touched).
- `{% trans` -> `{% translate`, `{% blocktrans` -> `{% blocktranslate`, `{% endblocktrans` -> `{% endblocktranslate`.
- Zero remaining legacy tags: `grep -rn "{% trans\b\|{% blocktrans\b" --include="*.html"` returns 0 hits.
- `{% plural %}` blocks left untouched, as were `{% load i18n %}` lines.

## Why it is safe

- The repo requires `django>=5.2` (pyproject.toml), and the modern tag names exist since Django 3.1 as documented drop-in renames, so no supported version loses the alias.
- Parse-checked every template with Django 5.2's template engine before and after: identical results (62 parse OK / 28 fail only on project-specific custom tag libraries such as `saved_queries` and `bootstrap5_form`, which require the full app registry — 0 i18n-related failures in both runs).

## Test plan

- [x] `grep` confirms zero legacy tags remain
- [x] No `translatelate`-style double renames (0 hits)
- [x] Django template parse parity vs baseline (62 OK / 28 pre-existing custom-lib failures, unchanged)
- [x] Maintainer CI (not run locally — full test suite needs the project dev environment)

PR #1430, opened 2026-09-09
i18n: standardize templates on translate/blocktranslate

### LINKED ISSUE #1427: Standardize translations on modern translation tag not legacy trans tag
As mentioned here https://github.com/django-helpdesk/django-helpdesk/pull/1425#discussion_r3946469076 we should move all of our trans entries to translate.

https://docs.djangoproject.com/en/3.1/releases/3.1/#templates

> The renamed [translate](https://docs.djangoproject.com/en/3.1/topics/i18n/translation/#std:templatetag-translate) and [blocktranslate](https://docs.djangoproject.com/en/3.1/topics/i18n/translation/#std:templatetag-blocktranslate) template tags are introduced for internationalization in template code. The older [trans](https://docs.djangoproject.com/en/3.1/topics/i18n/translation/#std:templatetag-trans) and [blocktrans](https://docs.djangoproject.com/en/3.1/topics/i18n/translation/#std:templatetag-blocktrans) template tags aliases continue to work, and will be retained for the foreseeable future.

probably run and double check something like
```bash
find . -name "*.html" -type f -exec sed -i -E \
  -e 's/\{% trans\b/{% translate/g' \
  -e 's/\{% blocktrans\b/{% blocktranslate/g' \
  -e 's/\{% endblocktrans\b/{% endblocktranslate/g' \
  {} +
```

Do nothing its optional

Mihailchik:
Closes #1427.

Django 3.1 renamed `trans`/`blocktrans` to `translate`/`blocktranslate`. The codebase was mid-migration — 273 modern tags already sitting next to 373 legacy ones — so this finishes the move rather than starting it.

**425 tags across 45 templates:** 373 `trans`, 26 `blocktrans`, 26 `endblocktrans`. Nothing left behind: `grep -rE '\{%-? *(trans|blocktrans|endblocktrans)\b' --include='*.html'` now returns nothing.

## Checks I ran before pushing

**Only tag names changed.** The diff is 403 removed lines and 403 added lines. Rewriting the new tag names back to the old spelling makes every added line byte-identical to the line it replaced — so no translatable string was touched, the existing `.po` catalogues stay valid, and `makemessages` produces the same msgids.

**`{% translate %}` was not double-converted.** The word boundary after `trans` leaves the 273 already-migrated tags alone; there is no `translatelate` anywhere.

**The scope really is `.html` only.** No `.txt`, `.py`, `.rst` or `.md` file in the repo carries these tags, so the `--include='*.html'` in the issue's suggested `sed` was not hiding anything.

**No `{% verbatim %}` or documentation example was caught.** The `verbatim` blocks in `help_context.html` contain only `{{ … }}` variables.

**Every template still parses.** Compiled all of them through Django's template engine: 0 syntax errors.

**Test suite:** `python quicktest.py` → 362 tests, `OK (skipped=2)`. Identical on a clean `main`, so nothing here changed behaviour.

I used a small Python script rather than the `sed` from the issue, mostly so the whitespace after `{%` is preserved as written rather than normalised, but the result is the same substitution.


Benbb96:
Hi @Mihailchik, thanks for your contribution but you're the third one to submit this PR (see #1428 and #1429).

uhurusurfa (approved):
Seems to have it all covered - thanks for the contribution.

Benbb96 (approved):
All good to me.

PR #1431, opened 2026-09-11
Hotfix/followup card colors

gurupratap-matharu:
The last color mode merge brought in some rough patches in UI under dark mode.

This PR fixes that. 

It's just mostly ironing out contrast issues and glaring buttons.
There are a few tweaks on the ticket detail page under followups. Hope you like 'em

This is a sweet spot to ask yourself (& your users) if they find that something doesn't look right under dark mode and just put it here. We can do hot fixes and prepare the app for the next release

Thank you

### DIFF
--- a/src/helpdesk/forms.py
+++ b/src/helpdesk/forms.py
@@ -795,4 +795,4 @@ class CreateChecklistForm(ChecklistForm):
 
 
 class FormControlDeleteFormSet(forms.BaseInlineFormSet):
-    deletion_widget = forms.CheckboxInput(attrs={"class": "form-control"})
+    deletion_widget = forms.CheckboxInput(attrs={"class": "form-check-input"})
--- a/src/helpdesk/static/helpdesk/css/helpdesk-extend.css
+++ b/src/helpdesk/static/helpdesk/css/helpdesk-extend.css
@@ -6,7 +6,6 @@ a:not(.btn) {
   text-decoration: none;
 }
 
-/* Bootstrap 5 removed the breadcrumb background; restore the BS4 grey */
 .breadcrumb {
   padding: 0.75rem 1rem;
   border-radius: 0.25rem;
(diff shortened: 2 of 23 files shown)


uhurusurfa (changes requested):
The checklists functionality needs some attention.
The checkboxes either are almost not visible (see bottom right):

Image

Or they are elongated and do not visibly show when they are checked (see the "Delete" column on the right though this is not a result of dark mode and seems to have been an issue for a little while):

Image

I think the outline of a checkbox should be a bright(er) colour when in dark mode.

uhurusurfa:
Checkboxes in light mode also needed to be dark(er)

DavidVadnais:
the sort got moved into its own line.  This takes up more space then it needs to

### was 
image

### now
image

DavidVadnais:
I can not screen shot it but the dark-mode kanban board makes a dashed box when moving tickets through drag and drop.  This dashed box is missing the right side

Knowledge base voting is maybe missing the vote scores? I do not use this often so I may be missing it.

DavidVadnais (commented):
1. I can not screen shot it but the dark-mode kanban board makes a dashed box when moving tickets through drag and drop.  This dashed box is missing the right side
2. Knowledge base voting is maybe missing the vote scores? I do not use this often so I may be missing it.
3. in both light and dark mode edit checklist delete seems bugged 
image

4. the drop down chevron doesnt do anything at least on my machine
image

5. the sort got moved into its own line.  This takes up more space then it needs to

### was 
image

### now
image

gurupratap-matharu:
Thank you for your feedback!

@uhurusurfa 

1. Check lists delete now shows a nice checkbox under edit mode and has good contrast. 
checklist

On the ticket detail page you should see better contrast under dark backgrounds.

@DavidVadnais 
2. The kanban thing is a fine catch. This is now fixed Do review and confirm. Here is a screenshot

kanban-hover

3. In knowledge base I do see the vote scores. I'll let @uhurusurfa  confirm this
kbitem

4. For you chevron down problem - does your checklist have atleast one task? If yes its wise that you do a screen video and create an issue.

5. For the follow up sort button, the question is: Do you want to show that button when you click on the Respond tab? I suppose no.

uhurusurfa (changes requested):
The issues I reported now look good.

Regarding KB counts - it does show the counts provided at least 1 person has clicked either the up or down vote button. The actual way KB counting works is not what I would consider normal but that is how it seems to have been since it was implemented so is not something to consider in this PR.

The checklist chevron works fine for me on Firefox and Chrome on MacOS

Image

The only thing outstanding from Davids review is the location of the sorting button in the top right of the table that shows the Followups - it needs to align with "Follow Ups" and "Respond" tab buttons if possible.

DavidVadnais:
I'm still running into the issues I talked about (other then the kanban one) but my setup is pretty bespoke so I can believe its a my machine problem.  

> Do you want to show that button when you click on the Respond tab? 

I would say no.

Benbb96:
> I'm still running into the issues I talked about (other then the kanban one) but my setup is pretty bespoke so I can believe its a my machine problem.
> 
> > Do you want to show that button when you click on the Respond tab?
> 
> I would say no.

Maybe the sort button needs to be on the left (above the display of the the timestamp. But still inside the Followups panel, because that doesn't make sense to display it in Respond tab.

uhurusurfa:
> I'm still running into the issues I talked about (other then the kanban one) but my setup is pretty bespoke so I can believe its a my machine problem.

@DavidVadnais - I have now checked on Safari as well and checklists works as expected for me.

uhurusurfa (approved):
I have opened #1436 to dewal with the issue noted by @DavidVadnais 
This PR covers all the specifics about dark and light mode and visibility of buttons so am approving and the noted issue can be done in a separate PR.

PR #1432, opened 2026-09-13
chore: CI on maintenance branches, and bump main to 2.5.0

Benbb96:
Two pieces of housekeeping that fall out of the 2.4.1 release.

## `ci: run the test matrix on maintenance branches`

`test.yml` fires on pull requests and on pushes to `main`. A maintenance line such as `2.4.x` matches neither: it never reaches `main` through a pull request, so nothing ever runs the tests on it.

That was not theoretical last week. 2.4.1 was a security release cut from `2.4.x`, and it was ready to tag on the strength of one local run against one Python and one Django. Adding the trigger first gave the full matrix, eight jobs green, including the Python 3.10 / Django 5.2 combination that still exists on that base and no longer runs here.

The `2.4.x` branch already carries this commit, since the workflow that runs on a push is the one on the branch being pushed. This brings it to `main` so the next maintenance line inherits it. The pattern is `"*.x"` rather than the literal branch name, so nobody has to remember.

## `bump to version 2.5.0`

`pyproject.toml` still said `2.4.0`, which is now older than what is on PyPI. 2.4.1 shipped from `2.4.x`, so `main` is developing toward the next minor rather than toward another 2.4 patch: it carries the colour mode work and the per-queue behaviour change, and neither belongs in a patch release.

Full suite passes, ruff check and format are clean.


.github/workflows/test.yml
```diff
@@ -5,6 +5,11 @@ on:
   push:
     branches:
       - main
+      # Maintenance lines, 2.4.x and whatever follows. A push to one of these is
+      # a release candidate and deserves the same matrix as main, and it is the
+      # only chance it gets: these branches never reach main through a pull
+      # request, so nothing else would ever run the tests on them.
```
DavidVadnais, line 11: Do we want these comments?

.github/workflows/test.yml
```diff
@@ -5,6 +5,11 @@ on:
...
+      - "*.x"
```
DavidVadnais, line 12: do we want this .x notation?

DavidVadnais:
I think we need to wait for #1431 to merge this because we have a partial implementation of light/dark mode in the code base right now.

.github/workflows/test.yml
uhurusurfa: I would keep them so any new maintainers understand why the *.x was added

.github/workflows/test.yml
uhurusurfa: Yes- tis allows running the CI/CD on maintenance branches that end up being created to release a security patch when other branches get merged into main whilst the security patch is being formulated.

uhurusurfa (approved):
👍

uhurusurfa:
> I think we need to wait for #1431 to merge this because we have a partial implementation of light/dark mode in the code base right now.

@DavidVadnais - Ben created the 2.4.x branch and cherry picked only the security fixes into it so that the release of the security fix is only about the security issue and does not include oither stuff and hence using the` *.x ` nomenclature in the test.yml to cater for this.

DavidVadnais:
Thank you for the info.  That makes sense

PR #1435, opened 2026-09-15
Feature/password reset

gurupratap-matharu:
Introduces Password reset functionality to helpdesk!

Here is a quick demo

https://github.com/user-attachments/assets/8d4253df-840c-46f3-a249-e10846cd3ad8

- We use the standard django auth views to achieve this with our custom templates
- Do verify edge cases from your end carefully
- Simple tests are introduced to check health of these new endpoints

If a user installs helpdesk into their own django project they might already have rolled out their own custom auth. 
I'm not sure how this will play around with that?

Thank you


src/helpdesk/urls.py
```diff
@@ -259,6 +268,18 @@ def get_context_data(self, **kwargs):
...
+    path(
+        "password-reset/<uidb64>/<token>/",
+        password_reset_confirm,
+        name="password_reset_confirm",
+    ),
+    path(
+        "password-reset-complete/",
+        password_reset_complete,
+        name="password_reset_complete",
+    ),
```
uhurusurfa, line 282: I suggest we define a new setting "HELPDESK_USE_DEFAULT_PASSWORD_RESET_URLS" in helpdesk/settings.py something like:
```
HELPDESK_USE_DEFAULT_PASSWORD_RESET_URLS = getattr(
    settings, "HELPDESK_ENABLE_DEPENDENCIES_ON_TICKET", True
)
```

Then wrap these in a settings check something like:

if helpdesk_settings.HELPDESK_USE_DEFAULT_PASSWORD_RESET_URLS:
    urlpatterns += [
           ....
           ....
    ]

uhurusurfa (changes requested):
This looks good.
The app already caters for using an alternative login system using the "LOGIN_URL" setting.
There is one place in the code that needs fixing to properly support use of the LOGIN_URL:
`src/helpdesk/views/staff.py:            return redirect_to_login(self.request.path, "helpdesk:login")`
This needs to be:
`src/helpdesk/views/staff.py:            return redirect_to_login(self.request.path, settings.LOGIN_URL)`

So I think if we fix that single line above and then wrap all the change password URLs in a setting as described in a separate review comment.

src/helpdesk/urls.py
gurupratap-matharu: This is a neat suggestion. 

I'm calling this setting `HELPDESK_ENABLE_PASSWORD_RESET`

gurupratap-matharu:
Thank you for the feedback!

A new simple settings `HELPDESK_ENABLE_PASSWORD_RESET` is introduced which defaults to `True`
If set to `False` the login form won't show the `Forgot password?` blue link.

Docs are updated to reflect this.

Please double check this thoroughly on your end on a real domain with real email if possible before approving.
Do toggle these settings to verify intended behavior. 

Thank you

src/helpdesk/templates/helpdesk/registration/login.html
```diff
@@ -36,7 +36,17 @@ <h5 class="text-center py-3">{% translate "Log in to your account" %}</h5>
               </div>
               <div class="mb-3">
                 <label class="form-label" for="password">{% translate "Password" %}</label>
-                <input type="password" id="password" name="password" class="form-control" placeholder="Your Password" required="required" />
+                <input type="password"
+                       id="password"
+                       name="password"
+                       class="form-control"
+                       placeholder="Your Password"
```
Benbb96: Could you add a translate block here, please

src/helpdesk/templates/helpdesk/registration/password_reset_confirm.html
```diff
@@ -0,0 +1,66 @@
...
+                  {% endif %}
+                  {{ form.new_password1 }}
+                </div>
+                <div class="mb-3">
+                  <label class="form-label" for="id_new_password2">{% translate "Confirm password" %}</label>
+                  {{ form.new_password2 }}
+                  {% if form.new_password2.errors %}
+                    <div class="form-text text-danger">{{ form.new_password2.errors }}</div>
+                  {% endif %}
+                </div>
```
Benbb96: Isn't it possible to use the template in `include/form.html` in order to display the form? This would automatically add the `form-control` class on the fields (avoiding custom js at the end of the file).

Maybe it breaks your layout ?

src/helpdesk/templates/helpdesk/registration/password_reset_email.html
```diff
@@ -0,0 +1,13 @@
...
+{% translate "Please visit the following page to choose a new password:" %}
+
+{{ protocol }}://{{ domain }}{% url 'helpdesk:password_reset_confirm' uidb64=uid token=token %}
+
+{% translate "If you didn't request a password reset, you can safely ignore this email. Your password will not be changed." %}
+
+{% translate "Thanks," %}
+{% blocktranslate %}The {{ site_name }} team{% endblocktranslate %}
+{% endautoescape %}
+{# djlint:on #}
```
Benbb96, line 13: Why do you need to disable djlint and autoescape in this template ?

src/helpdesk/templates/helpdesk/registration/password_reset_form.html
```diff
@@ -0,0 +1,50 @@
...
+  </div>
+{% endblock body %}
+{% block helpdesk_js %}
+  <script>
+    const add_form_classes = () => {
+      const email = document.querySelector('#id_email');
+      email.classList.add('form-control');
+    }
+    document.addEventListener("DOMContentLoaded", add_form_classes);
+  </script>
```
Benbb96: Same here, possible use of the form.html template.

Benbb96 (changes requested):

uhurusurfa (dismissed):
Works fine for me. Thanks for yet another contribution.

src/helpdesk/templates/helpdesk/registration/login.html
gurupratap-matharu: Done

src/helpdesk/templates/helpdesk/registration/password_reset_confirm.html
gurupratap-matharu: This is actually a nice suggestion! 

I'm using the template tag `{% bootstrap5_form form %}` instead of the include.
JS hack has been removed.

src/helpdesk/templates/helpdesk/registration/password_reset_form.html
gurupratap-matharu: Done

src/helpdesk/templates/helpdesk/registration/password_reset_email.html
gurupratap-matharu: djlint will format the template for better readability. 

This brings in unwanted spaces in the email body though.
Autoescape is still enabled.

I can remove the disable tags and send it but might break if someones formats all html templates in one go!

Benbb96: Understood, thanks for the explanation.

Benbb96 (approved):

PR #1438, opened 2026-09-17
Add follow-up backlink capability

### LINKED ISSUE #1424: When referencing a ticket in another ticket you often want to link that reference in the referenced ticket
When referencing a ticket in another ticket you often want to link that reference in the referenced ticket.  

This is like how in github when you # reference an issue or pr in another issue or pr a small entry is made in the referenced issue.  This could be useful in this ticketing system as well

I'm not sure maybe a full followup but that seems to space restrictive on the screen.

DavidVadnais:
Fixes: #1424 

### The look 
image

### Tested:
Entering `ref #1 #1 #1` creates one followup.
Entering `ref #1 #2` create one followup in each ticket 
Multiple followups referencing the same ticket create multiple backlinks.

Security should be fine since the title cant leak due only ticket numbers being referenced in the backlink. 

I can imagine this PR may take some back and forth as we try and get it just right. I'm going to look at adding a feature toggle for this now.


src/helpdesk/update_ticket.py
```diff
@@ -17,6 +19,51 @@
...
+    for ref_ticket in referenced_tickets:
+        FollowUp.objects.create(
+            ticket=ref_ticket,
+            title=_("Referenced in ticket #%(ticket_id)s")
+            % {"ticket_id": source_ticket.id},
+            date=timezone.now(),
+            public=False,
+            user=user
+            if user and hasattr(user, "is_authenticated") and user.is_authenticated
+            else None,
```
Benbb96: Is this ternary condition useful. It seems like there's an early exit at line 49.

src/helpdesk/email.py
```diff
@@ -704,6 +704,11 @@ def create_object_from_email_message(message, ticket_id, payload, files, logger)
             send_info_email(message_id, f, ticket, context, queue, new)
         )
     f.save()
+
+    from helpdesk.update_ticket import create_ticket_backlinks
+
+    create_ticket_backlinks(ticket, f)
```
Benbb96: This call passes no user, so `user` is `None` and the function returns at the authentication guard before creating anything. Incoming emails never produce a backlink.

I confirmed it locally: calling `create_ticket_backlinks(ticket, f)` exactly as written here creates zero followups on the referenced ticket, which is also what your own `test_no_user_creates_no_backlink` asserts.

So the email path needs a decision: resolve and pass the submitter when there is one, or define a rule that allows a user-less backlink. Either way it needs a test, because as it stands this line is dead code.

src/helpdesk/email.py
```diff
@@ -704,6 +704,11 @@ def create_object_from_email_message(message, ticket_id, payload, files, logger)
             send_info_email(message_id, f, ticket, context, queue, new)
         )
     f.save()
+
+    from helpdesk.update_ticket import create_ticket_backlinks
```
Benbb96: This import can go at the top of the file. There is no cycle to break: `update_ticket` only pulls in `settings`, `decorators`, `lib`, `models`, `signals` and `user`, and none of those import `email`. `forms.py` in this same PR already imports it at module level, which confirms it.

src/helpdesk/update_ticket.py
```diff
@@ -17,6 +19,51 @@
...
 
+TICKET_REF_RE = re.compile(r"(?:[^&]|\b|^)#(\d+)\b")
+
+
+def create_ticket_backlinks(source_ticket, followup, user=None):
+    if not helpdesk_settings.HELPDESK_ENABLE_BACKLINKS:
+        # If this feature is toggled off then do not do it
+        return
+
+    from helpdesk.user import HelpdeskUser
```
Benbb96: Same here, `HelpdeskUser` can be imported at the top of the file. `helpdesk.user` only imports `settings` and `models`, so there is no cycle.

src/helpdesk/update_ticket.py
```diff
@@ -17,6 +19,51 @@
...
+            referenced_ids.add(ticket_id)
+
+    if not referenced_ids:
+        return
+
+    referenced_tickets = Ticket.objects.filter(id__in=referenced_ids).select_related(
+        "queue"
+    )
+
+    if not (user and hasattr(user, "is_authenticated") and user.is_authenticated):
```
Benbb96: Could this guard move to the very top of the function? Right now it runs after the regex scan and after the queryset is built, so that work is thrown away whenever there is no user. Hoisting it also makes the precondition obvious to the reader, and it removes the need for the ternary further down.

src/helpdesk/settings.py
```diff
@@ -71,6 +71,9 @@
 #   able to download your ticket attachments.
 HELPDESK_ENABLE_ATTACHMENTS = getattr(settings, "HELPDESK_ENABLE_ATTACHMENTS", True)
 
+# Create a backlink followup when a ticket references another ticket
+HELPDESK_ENABLE_BACKLINKS = getattr(settings, "HELPDESK_ENABLE_BACKLINKS", True)
```
Benbb96, line 75: `HELPDESK_ENABLE_BACKLINKS` defaults to `True`, so every existing install starts generating these followups as soon as it upgrades. Was that the intent? We might wait for 2.6.0 to release it.

src/helpdesk/update_ticket.py
```diff
@@ -17,6 +19,51 @@
 
 User = get_user_model()
 
+TICKET_REF_RE = re.compile(r"(?:[^&]|\b|^)#(\d+)\b")
```
Benbb96, line 24: Could you add a one line comment on what the leading `(?:[^&]|\b|^)` is for? I checked and it does correctly skip HTML entities such as `&#39;`, but that intent is invisible to the next reader and the pattern is subtle enough that someone will have to reverse engineer it.

Benbb96 (changes requested):
One blocking issue on the email path, the rest is cleanup. Details in the line comments.

src/helpdesk/email.py
Benbb96: This call passes no user, so `user` is `None` and the function returns at the authentication guard before creating anything. Incoming emails never produce a backlink.

I confirmed it locally: calling `create_ticket_backlinks(ticket, f)` exactly as written here creates zero followups on the referenced ticket, which is also what your own `test_no_user_creates_no_backlink` asserts.

So the email path needs a decision: resolve and pass the submitter when there is one, or define a rule that allows a user-less backlink. Either way it needs a test, because as it stands this line is dead code.

src/helpdesk/email.py
Benbb96: This import can go at the top of the file. There is no cycle to break: `update_ticket` only pulls in `settings`, `decorators`, `lib`, `models`, `signals` and `user`, and none of those import `email`. `forms.py` in this same PR already imports it at module level, which confirms it.

src/helpdesk/update_ticket.py
Benbb96: Same here, `HelpdeskUser` can be imported at the top of the file. `helpdesk.user` only imports `settings` and `models`, so there is no cycle.

src/helpdesk/update_ticket.py
Benbb96: Could this guard move to the very top of the function? Right now it runs after the regex scan and after the queryset is built, so that work is thrown away whenever there is no user. Hoisting it also makes the precondition obvious to the reader, and it removes the need for the ternary further down.

Benbb96 (changes requested):

src/helpdesk/settings.py
DavidVadnais: I would want it on by default but I do not have a strong preference

src/helpdesk/email.py
DavidVadnais: got it

src/helpdesk/update_ticket.py
DavidVadnais: added comment

src/helpdesk/email.py
DavidVadnais: resolved in 2nd amending push

src/helpdesk/update_ticket.py
DavidVadnais: resolved in 2nd amending push

src/helpdesk/update_ticket.py
DavidVadnais: resolved in 2nd amending push

DavidVadnais:
Oof I think the repeated back linking is a little rough. I’m pretty sure GitHub only does one back link so I’ll have to change that.

DavidVadnais:
This is cleaner

image

src/helpdesk/settings.py
uhurusurfa: I think we just add a link in the README.rst  to the release notes and a note recommending that users read them before upgrading.
The responsibility is on the implementer to ensure they disable unwanted enhancements.

uhurusurfa (approved):
Just did a quick manual test and works fine for me.

Benbb96 (approved):

PR #1439, opened 2026-09-18
Enhancement/improve email ignore templates

gurupratap-matharu:
Brings in aesthetic improvements to the email ignore templates.

Here are some visuals

ignored-emails

add-ignore-email

These are hidden in system settings in the user top right menu. 
The removal of email from the ignore list never worked so slighlty refactored the view.
We show alert notifications for create and delete actions.
Simple unit tests are added for all three endpoints.

Thank you


src/helpdesk/templates/helpdesk/include/form_field.html
```diff
@@ -16,7 +17,8 @@
   </div>
 {% elif field|is_choice_group %}
   <fieldset class="mb-3">
-    <legend class="form-label {% if field.field.required %}{{ form.required_css_class }}{% endif %}">{{ field.label }}</legend>
+    <label class="form-label
```
DavidVadnais: Is changing from legend to label here good? This could affect screen readers.

src/helpdesk/views/staff.py
```diff
@@ -1968,47 +1967,42 @@ def get_object(self, queryset=None):
...
             return HttpResponseRedirect(reverse("helpdesk:email_ignore"))
     else:
         form = EmailIgnoreForm(request.GET)
 
     return render(request, "helpdesk/email_ignore_add.html", {"form": form})
 
 
-email_ignore_add = superuser_required(email_ignore_add)
+@superuser_required
+def email_ignore_del(request: HttpRequest, id: int) -> HttpResponse:
```
DavidVadnais: I believe id is shadowed here.  I think pk is more django standard but this aligns better with the URL pattern so 🤷

DavidVadnais (dismissed):
looks good I think the delete columns X's could be a bit more obvious but thats purely my opinion. 

I really only want to make sure we like the `legend` to `label` change. I dont know of any users that need a screen reader but you never know.

src/helpdesk/templates/helpdesk/include/form_field.html
gurupratap-matharu: I think legend is better. bs5 renders it with a large font so I'm taming it down a bit.

src/helpdesk/views/staff.py
gurupratap-matharu: You are correct! `id` seems very generic. `pk` seems generic too. 
For now I'm calling it `item_id` to be more specific

DavidVadnais (approved):
Looks good to me! Thank you

Issue #1441, opened 2026-09-19
On the filter tickets page only priority fails to sort all other columns can sort


DavidVadnais:
**Describe the bug**
On the filter tickets page only priority fails to sort all other columns can sort.

**To Reproduce**
1. `make rundemo`
2. login
3. go to all tickets
4. Add column priority 
5. try and sort
6. See error

**Expected behavior**
Column behaves like other columns

**Screenshots**
<img width="953" height="451" alt="Image" src="https://github.com/user-attachments/assets/5babfff0-2594-480d-8567-3caec879b025" />

**Desktop (please complete the following information):**
 - OS: Ubuntu 24.04
 - Browser: Firefox
 - Version: 146.0.1

DavidVadnais:
Fixed by @ezra-varady

PR #1442, opened 2026-09-19
fix column sorting on undisplayed columns

ezra-varady:
column sorting on undisplayed columns is generally broken, this is apparently due to a quick of the way DataTables 3 handles listeners, this fixes that, resolving #1441. I also force the priority header to be on the left side so that the layout matches the other columns.

I also bundled some trivial css changes to conform templates to the new dark mode setup


DavidVadnais (approved):
It works.  Code looks clean.

Image

